PR #2, opened 2020-03-09
Generate basic Go module with structures

JordonPhillips:
This adds the ability to generate structure shapes as well as most primitive shapes. Primitives that require imports and specializations of shapes based on traits (e.g. errors / enums) are also not yet supported. Wholly unsupported shape types currently output `nil` symbols.

The scope of this initial PR was deliberately limited to try to keep its size down, as it already requires a decent amount of infrastructure to get things going.

Currently the following files are generated:

```
.
├── go.mod
└── api_types.go
```

The contents of `go.mod` will look like:

```
module weather
```

The base module name here is configurable.

Based on the `smithy-go-codegen-test` model, the contents of `api_types.go` will look like:

```go
package weather

type CityCoordinates struct {
	Latitude  *float32
	Longitude *float32
}

type CitySummary struct {
	Case   *string
	CityId *string
	Name   *string
	Number *string
}

	CityId *string
}

	Image []byte
}

type GetCityInput struct {
	CityId *string
}

type GetCityOutput struct {
	City        CitySummary
	Coordinates CityCoordinates
	Name        *string
}

	Time nil
}

	CityId *string
}

	ChanceOfRain *float32
}

type ListCitiesInput struct {
	NextToken *string
	PageSize  *int32
}

	Items     []CitySummary
	NextToken *string
}
```

(Note that `go fmt` is being run on the output)


jasdel:
Minor suggestion for file/folder naming. I think the following pattern will help the SDK breakup go files into smaller chunks making it easier to view source since the files can be so large. e.g. v1 SDK's service/ec2/api.go file is over 100k lines and no longer can be rendered in github. The v2 SDK's equivalent file is about 10k lines.

```
api_client.go <- generated API client

api_op_Foo.go <- generated Operation method, operation request/response types, and
     operation Input type.  Generated pagination behavior could probably
     go here as well.

api_types.go <- all other struct types

api_enums.go <- all API enums
```

smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -0,0 +1,84 @@
...
+        for (Shape shape : serviceShapes) {
+            shape.accept(this);
+        }
+
+        LOGGER.fine("Flushing go writers");
+        writers.flushWriters();
+
+        LOGGER.fine("Generating go.mod file");
+        GoModGenerator.writeGoMod(settings, fileManifest);
+    }
```
skmcgrail: I'm a bit on the fence still on whether we should be generating to the go module file as part of the smithy type generation. Especially since at the time of generating this go mod file the complete dependency enclosure is not known from a type perspective, additionally it's not clear to me whether we should be treating this file as a code-generated artifact that is rewritten every generation.

JordonPhillips: > at the time of generating this go mod file the complete dependency enclosure is not known

Sure it is, why wouldn't it be? Code generator plugins will have to register their dependencies.

> it's not clear to me whether we should be treating this file as a code-generated artifact that is rewritten every generation

It's what defines a module, why wouldn't we generate it every time? The code generator has no notion of existing state, it outputs to an ephemeral directory that could be wiped at any time that doesn't have version control. From the SDK publishing perspective you would then copy that out into its final home.

jasdel:
It would be helpfuil for all generated `.go` files to include the following as the very first line.

```
// Code generated toolName DO NOT EDIT.
```

`toolName` can be anything that describe what did the generation, e.g. `smithy-go-codegen`. Reference, for where this comes from, https://github.com/golang/go/issues/13560#issuecomment-288457920

smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
JordonPhillips: We talked about this offline. We're still going to generate it, but we're going to use the Go command line tooling to do so.

smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    @Override
+    public Symbol bigDecimalShape(BigDecimalShape shape) {
+        // TODO: add once dependency support is in
+        return createSymbolBuilder(shape, "nil").build();
+    }
+
+    @Override
+    public Symbol operationShape(OperationShape shape) {
```
kstich: Should these other "nil" entries get TODOs for easy finding?

smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoSettings.java
```diff
@@ -0,0 +1,181 @@
...
+
+        settings.setModuleName(config.expectStringMember(MODULE_NAME).getValue());
+        settings.setModuleVersion(config.expectStringMember(MODULE_VERSION).getValue());
+        settings.setModuleDescription(config.getStringMemberOrDefault(
+                MODULE_DESCRIPTION, settings.getModuleName() + " client"));
+
+        return settings;
+    }
+
+    private static ShapeId inferService(Model model) {
```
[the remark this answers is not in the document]
JordonPhillips: For sure. Maybe a new `ModelUtils` in `smithy-utils`? It could also go somewhere in `codegen-core` but it's not really codgen specific.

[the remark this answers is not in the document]
JordonPhillips: There was offline discussion about this. We opted to remove the functionality entirely, at least for now.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public String toString() {
+        String contents = super.toString();
+        String[] packageParts = fullPackageName.split("/");
+        String header = String.format(
+                "// Code generated by smithy-go-codegen DO NOT EDIT.%npackage %s%n%n",
+                packageParts[packageParts.length - 1]);
+        // TODO: add imports
+        return header + contents;
```
jasdel: Don't have to do it ATM but wanted to call out, that it would be good to consider how we'll generate [summary/overview documentation API packages](https://blog.golang.org/godoc).

The package doc string should only appear in one file within a package, (e.g. doc.go file) and provides the documentation/summary/overview of a package. We can determine if this should be a canned string, or incorporate documentation provided by the API model, (e.g [v1 sdk's ACM](https://docs.aws.amazon.com/sdk-for-go/api/service/acm/#pkg-overview))

```go

// Package weather provides the client, and types for the Weather API. <- first paragraph is summary
//
// ... could be more that only appears in API docs overview, not included in package's summary.
package weather
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoSettings.java
```diff
@@ -0,0 +1,150 @@
...
+     * Create a settings object from a configuration object node.
+     *
+     * @param config Config object to load.
+     * @return Returns the extracted settings.
+     */
+    public static GoSettings from(ObjectNode config) {
+        GoSettings settings = new GoSettings();
+        config.warnIfAdditionalProperties(Arrays.asList(SERVICE, MODULE_NAME, MODULE_DESCRIPTION, MODULE_VERSION));
+
+        settings.setService(config.expectStringMember(SERVICE).expectShapeId());
```
kstich: Can you add a TODO for using `inferService` from Smithy's codebase instead of just fully removing it?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -0,0 +1,242 @@
...
+
+    @Override
+    public Symbol serviceShape(ServiceShape shape) {
+        // TODO: implement clients
+        return createSymbolBuilder(shape, "nil").build();
+    }
+
+    @Override
+    public Symbol stringShape(StringShape shape) {
+        // TODO: support specialized strings
```
jasdel: assuming specialized strings are enums?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoSettings.java
JordonPhillips: To be clear, the intent is to never infer the service.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
JordonPhillips: correct

JordonPhillips: and mediatype

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
JordonPhillips: Eventually we'll add a `useFileWriter` to `GoDelegator`, which we'll then use to write out such files. We should include the service shape's documentation there where available.

PR #5, opened 2020-03-19
Add support for dependencies and primitive shapes that require them

JordonPhillips:
This adds dependency support and support for timestamps, big ints, and big decimals. The generated test package's types now looks like:

```go
package weather

import (
	"time"
)

type CityCoordinates struct {
	Latitude  *float32
	Longitude *float32
}

type CitySummary struct {
	Case   *string
	CityId *string
	Name   *string
	Number *string
}

	CityId *string
}

	Image []byte
}

type GetCityInput struct {
	CityId *string
}

type GetCityOutput struct {
	City        *CitySummary
	Coordinates *CityCoordinates
	Name        *string
}

	Time *time.Time
}

	CityId *string
}

	ChanceOfRain *float32
}

type ListCitiesInput struct {
	NextToken *string
	PageSize  *int32
}

	Items     []CitySummary
	NextToken *string
}
```

Additionally, `go mod edit -require` will be called for any dependency outside of the standard library.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.go.codegen;
+
+import java.util.Collections;
+import java.util.List;
+import software.amazon.smithy.codegen.core.SymbolDependency;
+import software.amazon.smithy.codegen.core.SymbolDependencyContainer;
+
+/**
+ * An enum of all the built-in dependencies used by this package.
+ */
+public enum GoDependency implements SymbolDependencyContainer {
+
+    // The version in the stdlib dependencies should reflect the minimum Go version.
+    // The values aren't currently used, but they could potentially used to dynamically
+    // set the minimum go version.
+    BIG("stdlib", "math/big", "1.14"),
+    TIME("stdlib", "time", "1.14");
+
+    public final String packageName;
+    public final String version;
+    public final SymbolDependency dependency;
+
+    GoDependency(String type, String name, String version) {
+        this.dependency = SymbolDependency.builder()
+                .dependencyType(type)
+                .packageName(name)
+                .version(version)
+                .build();
+        this.packageName = name;
+        this.version = version;
+    }
+
+    @Override
+    public List<SymbolDependency> getDependencies() {
+        return Collections.singletonList(dependency);
+    }
+}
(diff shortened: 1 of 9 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+import software.amazon.smithy.codegen.core.SymbolDependency;
+import software.amazon.smithy.codegen.core.SymbolDependencyContainer;
+
+/**
+ * An enum of all the built-in dependencies used by this package.
+ */
+public enum GoDependency implements SymbolDependencyContainer {
+
+    // The version in the stlib dependencies should reflect the minimum Go version.
```
kstich: s/stlib/stdlib

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -228,15 +235,34 @@ public Symbol memberShape(MemberShape shape) {
...
     public Symbol timestampShape(TimestampShape shape) {
-        // TODO: add once dependency support is in
-        return createSymbolBuilder(shape, "nil").build();
+        return createSymbolBuilder(shape, "*time.Time")
+                .addReference(createNamespaceReference(GoDependency.TIME))
+                .build();
     }
 
     private Symbol.Builder createSymbolBuilder(Shape shape, String typeName) {
-        return Symbol.builder().putProperty("shape", shape).name(typeName);
```
kstich: nit: Why the change? If you want to keep, moving `.putProperty` seems right too.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
```diff
@@ -37,27 +52,141 @@ public GoWriter(String fullPackageName) {
...
+    public GoWriter addImport(String packageName, String as) {
+        imports.addImport(packageName, as);
+        return this;
+    }
+
+    /**
+     * Adds one or more dependencies to the generated code.
+     *
+     * <p>The dependencies of all writers created by the {@link GoDelegator}
+     * are merged together to eventually generate a package.json file.
```
kstich: "package.json" seems wrong.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
JordonPhillips: That's an artifact of other changes I was considering, but went back on. Looks like I missed putting this back.

jasdel:
I think this update may be missing using pointer types for struct shapes. Without pointer type the SDK won't be able to distinguish if a struct is set or not, e.g. empty struct with no member set, vs struct not set at all.

e..g `GetCityOutput` member `City` should be `*CitySummary`.

skmcgrail:
LGTM, can't actually approve as I don't have permissions yet.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -228,15 +235,39 @@ public Symbol memberShape(MemberShape shape) {
...
     @Override
     public Symbol timestampShape(TimestampShape shape) {
-        // TODO: add once dependency support is in
-        return createSymbolBuilder(shape, "nil").build();
+        return createSymbolBuilder(shape, "time.Time")
+                .addReference(createNamespaceReference(GoDependency.TIME))
+                .build();
+    }
+
+    private Symbol.Builder createSymbolBuilder(Shape shape, String typeName, boolean pointable) {
```
kstich: I think this would be cleaner as `createPointableSymbolBuilder(Shape shape, String typeName)` (overloaded with + `String namespace`) that sets the pointable value to true explicitly. This removes the need to drop in a bunch of `false` on non-pointables and indicates the symbols are pointable in the specific visitor methods. This also helps having 3 methods with slightly confusing type signatures: (A, B) + (A, B, C) + (A, B, D).

PR #6, opened 2020-03-20
 Add a Smithy middleware and stack common types

jasdel:
The Smithy middleware stack provides ordered behavior to be invoked on an
underlying handler. The stack is separated into steps that are invoked in a
static order. A step is a collection of middleware that are injected into a
ordered list defined by the user. The user may add, insert, swap, and remove a
step's middleware. When the stack is invoked the step middleware become static,
and their order cannot be modified.

A stack and its step middleware are **not** safe to modify concurrently.

A stack will use the ordered list of middleware to decorate a underlying
handler. A handler could be something like an HTTP Client that round trips an
API operation over HTTP.

The following represents the predefined steps within a stack. Each step has an
expectation of what has been completed by the step earlier in the stack.

* **Initialize**: Prepares the input, and sets any default parameters as
  needed, (e.g. idempotency token, and presigned URLs).

* **Serialize**: Serializes the prepared input into a data structure that can
  be consumed by the target transport's message, (e.g. REST-JSON serialization).

* **Build**: Adds additional metadata to the serialized transport message,
  (e.g. HTTP's Content-Length header, or body checksum). Decorations and
  modifications to the message should be copied to all message attempts.

* **Finalize**: Preforms final preparations needed before sending the message.
  The message should already be complete by this stage, and is only alternated
  to meet the expectations of the recipient, (e.g. Retry and AWS SigV4 request
  signing).

* **Deserialize**: Reacts to the handler's response returned by the recipient
  of the request message. Deserializes the response into a structured type or
  error above stacks can react to.

Middleware can be added to a step front or back, or relative, by name, to an
existing middleware in that stack. If a middleware does not have a name a
unique name will be generated at the middleware is added to the step.

```go
stack := middleware.NewStack()
stack.Initialize.Add(paramValidationMiddleware, middleware.After)
stack.Serialize.Add(marshalOperationFoo, middleware.After)
stack.Deserialize.Add(unmarshalOperationFoo, middleware.After)

resp, err := stack.HandleMiddleware(ctx, req.Input, clientHandler)
```

TODO:
* [x] Refactor stack example to be more descriptive.


middleware/stack_example_test.go
```diff
@@ -0,0 +1,94 @@
+package smithy_test
```
skmcgrail: This should be `middleware_test`

transport/http/request.go
```diff
@@ -0,0 +1,25 @@
...
+
+	URL  *url.URL
+	Host string
+
+	Header  Header
+	GetBody func() (io.ReadCloser, error)
+}
+
+type Response struct {
+	Request *Request
```
JordonPhillips: Why does the response need to include a pointer to the request?

jasdel: Forgot to clean these up. They are not apart of the middleware design.

Though with that said, this was a pattern the Go HTTP request/response used, and while brainstorming with these transport types i followed that pattern. But probably not needed for a smithy transport library.

jasdel:
Update contains refactored pkg docs, Also replaces the middleware `Name` method with a `ID`, and make the value for that ID required.

Need to update stack example to be more coherent

skmcgrail:
The signer middleware PR draft was working off the following definition for a Smithy HTTP transport type

```go

type Request struct {
	HTTPRequest *http.Request
	Stream      io.Reader
}

```

middleware/step_build.go
```diff
@@ -0,0 +1,182 @@
...
+	// for the middleware chain to continue. The method must return a result or
+	// error to its caller.
+	HandleBuild(ctx context.Context, in BuildInput, next BuildHandler) (
+		out BuildOutput, err error,
+	)
+}
+
+// BuildMiddlewareFunc returns a BuildMiddleware with the unique ID provided,
+// and the func to be invoked.
+func BuildMiddlewareFunc(id string, fn func(context.Context, BuildInput, BuildHandler) (BuildOutput, error)) BuildMiddleware {
```
skmcgrail: Curious on your thoughts for not exposing the `buildMiddlewareFunc` and it's members publically? I know it's kind of hit or miss around the community on whether functions/methods in libraries should return concrete vs interface types.

middleware/ordered_group.go
```diff
@@ -0,0 +1,238 @@
...
+	return nil
+}
+
+// Swap removes the item by id, replacing it with the new item. Returns error
+// if the original item doesn't exist.
+//
+// If the value implements the ider interface, that id must be unique. If
+// the value does not implement the ider interface, a new unique id will be
+// used.
+func (g *orderedIDs) Swap(id string, m ider) error {
```
skmcgrail: Should we return the original `ider` after a swap?

skmcgrail:
Just minor comments, otherwise LGTM

middleware/step_build.go
jasdel: Main reason for not exposing the struct was to force usage of the function, to prevent any accidental miss reading of the struct's requirements and incorrect thinking ID, or Fn may be optional.

middleware/ordered_group.go
jasdel: sure can update to return the swapped out ider.

jasdel:
correct will create separate Pr that adds the transport  request/resp types.

jasdel:
Thanks will fixup GetMiddlewareIDs` returning `[]string{}`. forgot that `append` can be called on a nil slice.

PR #7, opened 2020-03-25
Introduce document type support

JordonPhillips:
*Description of changes:*

This introduces an interface for document types, and ensures that their symbols are generated. A document type was introduced into the sample model for demonstrative purposes:

```go
package weather

import (
	"time"
)

type CityCoordinates struct {
	Latitude  *float32
	Longitude *float32
}

type CitySummary struct {
	Case   *string
	CityId *string
	Name   *string
	Number *string
}

	CityId *string
}

	Image []byte
}

type GetCityInput struct {
	CityId *string
}

type GetCityOutput struct {
	City        *CitySummary
	Coordinates *CityCoordinates
	Metadata    smithy.Document
	Name        *string
}

	Time *time.Time
}

	CityId *string
}

	ChanceOfRain *float32
}

type ListCitiesInput struct {
	NextToken *string
	PageSize  *int32
}

	Items     []CitySummary
	NextToken *string
}
```

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
@@ -181,8 +181,9 @@ final class SymbolVisitor implements SymbolProvider, ShapeVisitor<Symbol> {
 
     @Override
     public Symbol documentShape(DocumentShape shape) {
-        // TODO: implement document shapes
-        return createPointableSymbolBuilder(shape, "nil").build();
+        return createPointableSymbolBuilder(shape, "smithy.Document")
+                .addReference(createNamespaceReference(GoDependency.SMITHY, "smithy"))
+                .build();
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


document.go
```diff
@@ -0,0 +1,52 @@
...
+// JSON values.
+type Document interface {
+	// Marshal will attempt to serialize the provided type into a
+	// protocol specific document value.
+	Marshal(x interface{}) error
+
+	// Unmarshall will attempt to map the protocol specific document value
+	// onto the provided type. If the type does not match the document type
+	// an error will be returned.
+	Unmarshal(x interface{}) error
```
jasdel, line 17: To prevent potentially colliding with other `Unmarshal` methods of other types it would be helpful to rename these to `MarshalDocument`, `UnmarshalDocument`. This would also follow the common pattern in the stdlib for serialization methods.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -163,8 +163,9 @@ public Symbol floatShape(FloatShape shape) {
 
     @Override
     public Symbol documentShape(DocumentShape shape) {
-        // TODO: implement document shapes
-        return createPointableSymbolBuilder(shape, "nil").build();
+        return createPointableSymbolBuilder(shape, "smithy.Document")
+                .addReference(createNamespaceReference(GoDependency.SMITHY, "smithy"))
+                .build();
```
jasdel, line 168: Since `smithy.Document` is a interface its usage should not be as a pointer.

PR #8, opened 2020-03-26
Introduce MediaType support

JordonPhillips:
*Description of changes:*

This introduces an interface for media types, and generates types for shapes that use them. A string or byte array can have a `MediaType()` function that returns its media type, which is defined in an interface in the base package. Currently there's no special handling for any particular media type.

```go
package weather

import (
	"time"
)

type CityCoordinates struct {
	Latitude  *float32
	Longitude *float32
}

type CityImageData []byte

func (m CityImageData) MediaType() string {
	return "image/jpeg"
}

type CitySummary struct {
	Case   *string
	CityId *string
	Name   *string
	Number *string
}

	CityId *string
}

	Image *CityImageData
}

type GetCityInput struct {
	CityId *string
}

type GetCityOutput struct {
	City        *CitySummary
	Coordinates *CityCoordinates
	Name        *string
}

	Time *time.Time
}

	CityId *string
}

	ChanceOfRain *float32
}

type ListCitiesInput struct {
	NextToken *string
	PageSize  *int32
}

	Items     []CitySummary
	NextToken *string
}
```

Right now each shape is getting its own new type, with no de-duplication. We *could* try to condense it down to something like one generated `MediaTypeApplicationJSON` for all `application/json` etc, but it doesn't seem necessary and it introduces complexity that we may not want. Plugins will, eventually, be able to decorate the shape writer so they can add in decoding methods.

[working sample](https://play.golang.org/p/50mzFb6PgRU)

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/MediaTypeGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/MediaTypeGenerator.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.go.codegen;
+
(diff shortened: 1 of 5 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/MediaTypeGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+        Symbol symbol = symbolProvider.toSymbol(shape);
+
+        Symbol baseType = symbol.getProperty("mediaTypeBaseSymbol", Symbol.class)
+                .orElseThrow(() -> new CodegenException(
+                        "Media type symbols must have the property mediaTypeBaseSymbol: " + shape.getId()));
+
+        writer.write("type $L $T", symbol.getName(), baseType)
+                .write("")
+                .openBlock("func (m $L) MediaType() string {", "}", symbol.getName(), () -> {
+                    writer.write("return \"$L\"", shape.expectTrait(MediaTypeTrait.class).getValue());
```
kstich: `\"$L\"` is just `$S`

PR #9, opened 2020-03-27
Generate enums

JordonPhillips:
*Description of changes:*

This introduces enum generation. Enums are generated into `api_enums.go`, like so:

```go
package weather

type SimpleYesNo string

// Enum values for SimpleYesNo
const (
	SimpleYesNoYes SimpleYesNo = "YES"
	SimpleYesNoNo  SimpleYesNo = "NO"
)

type TypedYesNo string

// Enum values for TypedYesNo
const (
	TypedYesNoYes TypedYesNo = "Yes"
	TypedYesNoNo  TypedYesNo = "No"
)
```

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.go.codegen;
+
+import java.util.Locale;
+import java.util.Map;
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.codegen.core.SymbolProvider;
+import software.amazon.smithy.model.shapes.StringShape;
+import software.amazon.smithy.model.traits.EnumConstantBody;
+import software.amazon.smithy.model.traits.EnumTrait;
+import software.amazon.smithy.utils.StringUtils;
+
+/**
+ * Renders enums and their constants.
+ */
+final class EnumGenerator implements Runnable {
+
+    private final SymbolProvider symbolProvider;
+    private final GoWriter writer;
+    private final StringShape shape;
+
+    EnumGenerator(SymbolProvider symbolProvider, GoWriter writer, StringShape shape) {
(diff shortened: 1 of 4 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+    public void run() {
+        Symbol symbol = symbolProvider.toSymbol(shape);
+        EnumTrait enumTrait = shape.expectTrait(EnumTrait.class);
+
+        writer.write("type $L string", symbol.getName()).write("");
+
+        writer.write("// Enum values for $L", symbol.getName()).openBlock("const (", ")", () -> {
+            for (Map.Entry<String, EnumConstantBody> entry : enumTrait.getValues().entrySet()) {
+                StringBuilder labelBuilder = new StringBuilder(symbol.getName());
+                String name = entry.getValue().getName().orElse(entry.getKey());
```
kstich: Do valid enum values (aside from the whitespace which is cleaned up below) work 100% as parts of Go's identifiers?

jasdel: If the enum value has any non Unicode letter, or number character those have to be stripped as well. And enum must start with an Unicode letter character. So things like `__` prefix must be stripped out or the type will not be public.

Though since Smithy allows naming of enum values e.g. map. this should resolve the issue. As long as Smithy's name keys are not too permissive.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+                for (String part : name.split("\\W")) {
+                    labelBuilder.append(StringUtils.capitalize(part.toLowerCase(Locale.US)));
+                }
+                String label = labelBuilder.toString();
+                writer.write("$L $L = $S", label, symbol.getName(), entry.getKey());
```
jasdel: Would the `Locale.US` prevent non-english Unicode letters from being used in the name of an enum?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
kstich: Not all enums will have names, even if we recommend it. We need to make sure all cases for valid values for these are handled.

Given that the name components are appended to the symbol name, the `__` prefix issue shouldn't be a problem specifically.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
JordonPhillips: No, that's just for determining what counts as upper case vs lower case and what the mappings are. E.g. if the locale isn't explicitly set then a capital `I` could become `ı` instead of `i` depending on the locale of the running system.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
JordonPhillips: The prefix issue would be the only real problem. There's standard bits like Json -> JSON but that's not critical.

PR #11, opened 2020-03-31
Emit an io.Reader for streaming blobs

JordonPhillips:
This updates blobs with the `streaming` trait to `io.Reader` instead of `[]byte`. They'll now look like:

```go
import (
	"io"
)

type MyStruct struct {
	StreamingBlob io.Reader
}
```

Or if they have the `mediatype` trait too:

```go
import (
	"io"
)

type CityImageData io.Reader

func (m CityImageData) MediaType() string {
	return "image/jpeg"
}
```

Note that this PR is dependent on [this Smithy change](https://github.com/awslabs/smithy/pull/340). This PR is still merge-able. It'll run, but the code path is effectively dead until the Smithy PR is merged and this generator updates to that version.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
@@ -29,6 +29,7 @@ public enum GoDependency implements SymbolDependencyContainer {
     // The values aren't currently used, but they could potentially used to dynamically
     // set the minimum go version.
     BIG("stdlib", "math/big", "1.14"),
+    IO("stdlib", "io", "1.14"),
     TIME("stdlib", "time", "1.14");
 
     public final String packageName;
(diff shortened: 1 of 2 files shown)


skmcgrail:
A note, since `io.Reader` is an interface we don't need to generate it as a `*io.Reader` instead it can just be `io.Reader`

jasdel:
To extend on @skmcgrail  comment, generated interface types should never be used as pointers. they should always be value as a general rule. Since interfaces are already a pointer, similar to maps and slices.

JordonPhillips:
> Does smithy allow a non-toplevel input/output shape to contain a Stream member?

Nope, wouldn't really make any sense.

> Will this account for the behavior differences between an stream used as input vs a stream used as output?

No, you can't guarantee a given shape isn't used for only output or only input. I know we discussed making custom types for operation inputs / outputs, so that might be a customization we can do then.

jasdel:
```go
type CityImageData io.Reader
```

Smithy won't be able to create a type alias of an interface. Go does not allow type alias of interfaces.

The MediaType shapes probably will need to structures that contain a underlying reader. What populatees that media type's reader will be based on how it is used. e.g. streaming vs nested.

```go
type CityImageData struct {
     io.Reader
}

type AsCityImageData(r io.Reader) *CityImageData { return &CityImageData{Reader: r} }
```

Downside of this approach is that if the underlying `io.Reader` provided by the user has any additional optimizations like `ReadFrom` etc, those will not be used by utilities like `io.Copy` because Go type assertions will not query embedded interface implementations. Potentially could generate these helpers onto the generated MediaType to compensate, as a fallback.

jasdel:
We'll also want to be aware that the the existing SDK's designs treat input and output streams differently. Input stream being be an io.Reader, and output io.ReadCloser.

The smithy generation could consider requiring an io.ReadCoser for input as well, but this is uncommon outside of the http.Client's Request since most ownership of Closing is expected to be the caller, not the callee.

Utilities like [ioutil.NopCloser](https://pkg.go.dev/io/ioutil?tab=doc#NopCloser) are available if a user doesn't want Close to be called or dont' want to use a type that implements io.Closer.

PR #12, opened 2020-04-10
Add Smithy APIError base and wrapping error types

jasdel:
Adds the base error and metadata wrapping error types that Smithy SDKs will use to compose APIErrors received from API responses with additional metadata.

Smithy generated SDK API errors will implement a common smithy APIError interface. This interface provides generic accessor methods that all generated client API errors, will implement. Metadata about the operation, response, retry, etc will all be provided by error types
wrapping the returned APIError value. This allows the SDK to use errors across multiple protocols like HTTP and event stream without additional noise. This approach also protects the SDK against additional metadata fields being added to the chain forcing the SDK to use either optional interfaces, or breaking change to expose the new metadata.

Layering errors this way acts similar to the mix-in design pattern. Each layer provides additional metadata describing not just the error but the path it took to come back to the user.

Example of using the API errors with this design.

```go
result, err := client.GetObject(ctx, params)
if err != nil {
    var apiErr smithy.APIError
    if errors.As(err, &apiErr) {
        fmt.Printf("request failed: %s, %s\n", apiErr.ErrorCode(), apiErr.ErrorMessage())
    }

    var respErr *smithyhttp.ResponseError
    if errors.As(err, &respErr) {
        fmt.Printf("response header: %v\n", respErr.HTTPResponse().Header.Get("Extra-Header"))
    }
}
```

### DIFF
--- a/fault.go
+++ b/fault.go
@@ -1,0 +1,1 @@
+package smithy
--- a/fault.go
+++ b/fault.go
@@ -1,1 +1,0 @@
-package smithy
--- a/errors.go
+++ b/errors.go
@@ -1,0 +1,82 @@
+package smithy
+
+import "fmt"
+
+// APIError provides the generic API and protocol agnostic error type all SDK
+// generated exception types will implement.
+type APIError interface {
+	error
+
+	// ErrorCode returns the error code for the API exception.
+	ErrorCode() string
+	// ErrorMessage returns the error message for the API exception.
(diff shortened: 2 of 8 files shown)


fault.go
```diff
@@ -0,0 +1 @@
+package smithy
```
JordonPhillips, line 1: Why is this here?

PR #17, opened 2020-04-17
Generate error interfaces and structures

JordonPhillips:
*Description of changes:*

This adds error generation. The errors are produced as interfaces in order to support inheritance.

```go
package weather

import (
	"fmt"
)

// Error encountered when no resource could be found.
type NoSuchResourceInterface interface {
	smithy.APIError
	isNoSuchResource()

	ErrorMessage() string
	ErrorFault() string
	ErrorCode() string

	// The type of resource that was not found.
	GetResourceType() string
	// HasResourceType returns whether ResourceType exists.
	HasResourceType() bool
}

// The concrete implementation for NoSuchResourceInterface. This should not be used directly.
type NoSuchResource struct {
	Message *string

	ResourceType *string
}

func (e *NoSuchResource) Error() string {
	return fmt.Sprintf("%s: %s", e.ErrorCode(), e.ErrorMessage())
}
func (e *NoSuchResource) isNoSuchResource() {}
func (e *NoSuchResource) ErrorMessage() string {
	if e.Message == nil {
		return ""
	}
	return *e.Message
}
func (e *NoSuchResource) ErrorCode() string             { return "NoSuchResource" }
func (e *NoSuchResource) ErrorFault() smithy.ErrorFault { return smithy.FaultClient }
func (e *NoSuchResource) GetMessage() string {
	return *e.Message
}
func (e *NoSuchResource) HasMessage() bool {
	return e.Message != nil
}
func (e *NoSuchResource) GetResourceType() string {
	return *e.ResourceType
}
func (e *NoSuchResource) HasResourceType() bool {
	return e.ResourceType != nil
}
```

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
@@ -96,7 +96,6 @@ final class SymbolVisitor implements SymbolProvider, ShapeVisitor<Symbol> {
         // Reserved words that only apply to error members.
         ReservedWords reservedErrorMembers = new ReservedWordsBuilder()
                 .put("ErrorCode", "ErrorCode_")
-                .put("ErrorMessage", "ErrorMessage_")
                 .put("ErrorFault", "ErrorMessage_")
                 .put("Unwrap", "Unwrap_")
                 .put("Error", "Error_")
@@ -342,7 +341,8 @@ final class SymbolVisitor implements SymbolProvider, ShapeVisitor<Symbol> {
         Symbol.Builder builder = createPointableSymbolBuilder(shape, name, rootModuleName);
         if (shape.hasTrait(ErrorTrait.ID)) {
             builder.definitionFile("./api_errors.go")
-                    .addReference(createNamespaceReference(GoDependency.SMITHY, "smithy"));
+                    .addReference(createNamespaceReference(GoDependency.SMITHY, "smithy"))
+                    .addReference(createNamespaceReference(GoDependency.FMT));
         } else {
             builder.definitionFile("./api_types.go");
         }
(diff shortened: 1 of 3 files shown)


jasdel:
Is there anyway the smithy generator would be able to de-duplicate the `message` and Message` members? I'm guessing they are they same. Also it would be good to avoid private members since users may want to create SDK operation stubs which returns specific errors to exercise their application's error handling behavior. Private member makes it more difficult for them to do that.

It would be great to consider at least `Message`, and maybe `Code` built in that should be de-duplicated.

JordonPhillips:
What exactly makes ErrorMessage not stubbable? the fact that it's backed by some hidden thing is an implementation detail no?

jasdel:
If the customer has code that reacts to a specific generated error type, not code string, or interface, (e.g. `NoSuchResource` type), its impossible for the customer to provide a value for the `message` (unexported) member of that type.  The customer could provide an alternative implementation for `NoSuchResourceInterface`, but that that also requires them to use that type in their code, and not `NoSuchResource` type.  Since both are exported there are no implementation details, other than `ErrorCode` is return value is hard coded.


It would be possible to not export `NoSuchResource`, but that requires users to implement the `NoSuchResourceInterface` type them selves, which is probably not a good experience needing to re-implement generated error types them selves.

jasdel:
Didn't notice Error method just returned `e.ErrorMessage` that should include the code as well. e.g. `return fmt.Sprintf("%s: %s", e.ErrorCode(), e.ErrorMessage)`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -80,12 +86,79 @@
...
                 .escapePredicate((shape, symbol) -> !StringUtils.isEmpty(symbol.getDefinitionFile()))
                 .buildEscaper();
+
+        // Reserved words that only apply to error members.
+        ReservedWords reservedErrorMembers = new ReservedWordsBuilder()
+                .put("ErrorCode", "ErrorCode_")
+                .put("ErrorFault", "ErrorMessage_")
+                .put("Unwrap", "Unwrap_")
+                .put("Error", "Error_")
+                .build();
```
jasdel, line 102: I think `ErrorFault` and `ErrorMessage` might be an incorrect here.

JordonPhillips: whoops

PR #20, opened 2020-04-23
Add Smithy Middleware Generator Utility

skmcgrail:
Adds a generator utility class that can be used for constructing specific types of middleware such as serialize and deserialize middleware. Will allow for easier extension for services requiring customization via specific middleware that are code generated.

### Sample Output
```go
package middlewaregentest

import (
	"context"
	"github.com/awslabs/smithy-go/middleware"
)

type someMiddlewareId struct {
}

func (*someMiddlewareId) ID() string {
	return "someMiddlewareId"
}

func (m *someMiddlewareId) HandleSerialize(ctx context.Context, in middleware.SerializeInput, next middleware.SerializeHandler) (
	out middleware.SerializeOutput, metadata middleware.Metadata, err error,
) {
	return next.HandleSerialize(ctx, in)
}

```


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+
+/*
+ * Helper for generating Smithy Go middleware
+ */
+public class GoMiddlewareGenerator {
+    private static final SymbolReference CONTEXT_TYPE;
+    private static final SymbolReference METADATA_TYPE;
+
+    private final String identifier;
+    private final String handleMethodName;
```
skmcgrail, line 31: Curious if anyone has thoughts on this also being defined as a symbol with a relevant getter method. Reason being is the expected pattern is that the implementation of the handler body calls the `next.%handleMethodName(ctx, in)`. Maybe this makes more sense as a symbol?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+
+import java.util.function.BiConsumer;
+
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.codegen.core.SymbolReference;
+
+/*
+ * Helper for generating Smithy Go middleware
+ */
+public class GoMiddlewareGenerator {
```
JordonPhillips, line 26: This should be finalized

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+
+    public GoMiddlewareGenerator(Builder builder) {
+        this.identifier = builder.identifier;
+        this.handleMethodName = builder.handleMethodName;
+        this.inputType = builder.inputType;
+        this.outputType = builder.outputType;
+        this.handlerType = builder.handlerType;
+    }
+
+    public static GoMiddlewareGenerator newSerializeMiddleware(String identifier) {
```
JordonPhillips, line 57: All public functions should have full docs

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+    private static final SymbolReference METADATA_TYPE;
+
+    private final String identifier;
+    private final String handleMethodName;
+    private final SymbolReference inputType;
+    private final SymbolReference outputType;
+    private final SymbolReference handlerType;
+
+    static {
+        CONTEXT_TYPE = SymbolReference.builder()
```
JordonPhillips, line 37: These can can just be put on the declarations

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+                        .symbol(SymbolUtils.createValueSymbolBuilder("middleware.DeserializeHandler")
+                                .addReference(SymbolUtils.createNamespaceReference(GoDependency.SMITHY_MIDDLEWARE))
+                                .build())
+                        .build())
+                .build();
+    }
+
+    public void writeMiddleware(GoWriter writer, BiConsumer<GoMiddlewareGenerator, GoWriter> handlerBodyConsumer) {
+        writeMiddleware(writer, (m, w) -> {
+        }, handlerBodyConsumer);
```
JordonPhillips, line 103: This should probably be on one line

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    public void writeMiddleware(GoWriter writer, BiConsumer<GoMiddlewareGenerator, GoWriter> fieldConsumer,
```
JordonPhillips, line 106: Sadly we don't have something as nice as go fmt, but when args make a line longer than 120 characters you should just put one arg per line, e.g.:

```java
public void foo(
    String bar,
    String baz
) {
    // stuff
}
```

I would also recommend swapping the order of `fieldConsumer` and `handlerBodyConsumer` as it's a bit more ergonomic for optional parameters to be at the end.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+                                BiConsumer<GoMiddlewareGenerator, GoWriter> handlerBodyConsumer) {
+        writer.addUseImports(CONTEXT_TYPE);
+        writer.addUseImports(METADATA_TYPE);
+        writer.addUseImports(inputType);
+        writer.addUseImports(outputType);
+        writer.addUseImports(handlerType);
+
+        Symbol middlwareSym = SymbolUtils.createPointableSymbolBuilder(identifier).build();
+
+        writer.openBlock("type $L struct {", "}", middlwareSym, () -> {
```
JordonPhillips, line 116: Each of these blocks should have a comment explaining what the block being created does.

codegen/smithy-go-codegen/src/test/java/software/amazon/smithy/go/codegen/GoMiddlewareGeneratorTest.java
```diff
@@ -0,0 +1,27 @@
...
+        GoWriter writer = new GoWriter("middlewaregentest");
+
+        GoMiddlewareGenerator.newSerializeMiddleware("someMiddlewareId")
+                .writeMiddleware(writer, (m, w) -> {
+                    w.openBlock("return next.$L(ctx, in)", m.getHandleMethodName());
+                });
+
+        String generated = writer.toString();
+
+        assertThat(generated, containsString("type someMiddlewareId struct {"));
```
JordonPhillips, line 20: You don't need to get rid of this, but these sorts of tests aren't super useful. Better to test the generated code than to make assertions about what the output looks like. Granted that's more or less impossible until we have a complete vertical slice.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+        this.outputType = builder.outputType;
+        this.handlerType = builder.handlerType;
+    }
+
+    public static GoMiddlewareGenerator newSerializeMiddleware(String identifier) {
+        return builder()
+                .identifier(identifier)
+                .handleMethodName("HandleSerialize")
+                .inputType(SymbolReference.builder()
+                        .symbol(SymbolUtils.createValueSymbolBuilder("middleware.SerializeInput")
```
kstich, line 62: These should probably be DRY'd up.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+        Symbol middlwareSym = SymbolUtils.createPointableSymbolBuilder(identifier).build();
+
+        writer.openBlock("type $L struct {", "}", middlwareSym, () -> {
+            fieldConsumer.accept(this, writer);
+        });
+        writer.openBlock("func ($P) ID() string {", "}", middlwareSym, () -> {
+            writer.openBlock("return $S", identifier);
+        });
+        writer.openBlock("func (m $P) $L(ctx $T, in $T, next $T) (\n"
+                        + "\tout $T, metadata $T, err error,\n"
```
kstich, line 123: Minor: it's nice down the road to line up the starting quotes of multi-line components like this.

```
writer.openBlock("foobar\n"
               + "\tbaz"...
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoMiddlewareGenerator.java
```diff
@@ -0,0 +1,201 @@
...
+                        + "){", "}",
```
kstich, line 124: Missing a " " between "){"?

codegen/smithy-go-codegen/src/test/java/software/amazon/smithy/go/codegen/GoMiddlewareGeneratorTest.java
skmcgrail: Agreed not a worthwhile test once we have other functioning components.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStackStepMiddlewareGenerator.java
```diff
@@ -0,0 +1,329 @@
...
+                    .addReference(SymbolUtils.createNamespaceReference(GoDependency.CONTEXT))
+                    .build())
+            .build();
+    private static final SymbolReference METADATA_TYPE = SymbolReference.builder()
+            .symbol(SymbolUtils.createValueSymbolBuilder("middleware.Metadata")
+                    .addReference(SymbolUtils.createNamespaceReference(GoDependency.SMITHY_MIDDLEWARE))
+                    .build())
+            .build();
+
+    private final Symbol middlewareSym;
```
JordonPhillips, line 38: Abbreviations like `Sym` are generally not done in Java

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStackStepMiddlewareGenerator.java
```diff
@@ -0,0 +1,329 @@
...
+     * Generates a new step middleware generator
+     *
+     * @param identifier        the name of the middleware type
+     * @param handlerMethodName identifier method name to be implemented
+     * @param inputType         the middleware input type
+     * @param outputType        the middleware output type
+     * @param handlerType       the next handler type
+     * @return the middleware generator
+     */
+    public static GoStackStepMiddlewareGenerator createMiddleware(String identifier,
```
JordonPhillips, line 95: Hate to be pedantic, but the first should be on its own line too, indented only one over from the function declaration itself.

kstich: As an example, here's how we'd likely write this function signature:

```java
public static GoStackStepMiddlewareGenerator createMiddleware(
        String identifier,
        String handlerMethodName,
        SymbolReference inputType,
        SymbolReference outputType,
        SymbolReference handlerType
) {
```

This helps from having hard to grok really long lines of WS if you have a long main signature that takes a long argument.

PR #30, opened 2020-05-06
Generate clients

JordonPhillips:
*Description of changes:*

This is a draft of client generation, based on https://github.com/aws/aws-sdk-go-v2/pull/530

Outside of that the only real gap is the lack application protocol handling, which will be necessary to set some config defaults and to initialize the middleware stack.

A generated service currently looks like:

```go
package weather

import (
	"github.com/awslabs/smithy-go/middleware"
	smithyhttp "github.com/awslabs/smithy-go/transport/http"
)

// Provides weather forecasts.
type Client struct {
	options Options
}

// New returns an initialized Client based on the functional options.
// Provide additional functional options to further configure the behavior
// of the client, such as changing the client's endpoint or adding custom
// middleware behavior.
func New(options Options) (*Client, error) {
	options = options.Copy()

	client := &Client{
		options: options,
	}

	return client, nil
}

// ServiceID returns the name of the identifier for the service API.
func (c *Client) ServiceID() string { return "weather" }

// ServiceName returns the full service title.
func (c *Client) ServiceName() string { return "weather" }

type Options struct {
	// Set of options to modify how an operation is invoked. These apply to all operations invoked for this client. Use functional options on operation call to modify this list for per operation behavior.
	APIOptions []APIOptionFunc

	// The HTTP client to invoke API calls with. Defaults to client's default HTTP implementation if nil.
	HTTPClient HTTPClient
}

type HTTPClient interface {
	Do(*smithyhttp.Request) (*smithyhttp.Response, error)
}

// Copy creates a clone where the APIOptions list is deep copied.
func (o Options) Copy() Options {
	to := o
	to.APIOptions = make([]APIOptionFunc, len(o.APIOptions))
	copy(to.APIOptions, o.APIOptions)
	return to
}

type APIOptionFunc func(*middleware.Stack) error
```

A generated operation current looks like:

```go
package weather

import (
	"context"
	"github.com/awslabs/smithy-go/middleware"
	smithyhttp "github.com/awslabs/smithy-go/transport/http"
	"weather/types"
)

func (c *Client) GetCity(ctx context.Context, params *GetCityInput, opts ...func(*Options)) (*GetCityOutput, error) {
	stack := middleware.NewStack("GetCity", smithyhttp.NewStackRequest)
	options := c.options.Copy()
	for _, fn := range optFns {
		fn(&options)
	}
	for _, fn := range options.APIOptions {
		if err := fn(stack); err != nil {
			return nil, err
		}
	}
	handler := middleware.DecorateHandler(smithyhttp.ClientHandler{Client: options.HTTPClient}, stack)
	result, metadata, err := handler.Handle(ctx, params)
	if err != nil {
		return nil, err
	}
	out := result.(*GetCityOutput)
	out.ResultMetadata = metadata
	return out, nil
}

// The input used to get a city.
type GetCityInput struct {
	CityId *string
}

type GetCityOutput struct {
	City        *types.CitySummary
	Coordinates *types.CityCoordinates
	Metadata    smithy.Document
	Name        *string

	// Metadata pertaining to the operation's result.
	ResultMetadata middleware.Metadata
}
```

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.go.codegen;
+
+import software.amazon.smithy.codegen.core.CodegenException;
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.codegen.core.SymbolProvider;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.OperationIndex;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.StructureShape;
+
+/**
+ * Generates a client operation and associated custom shapes.
+ */
+final class OperationGenerator implements Runnable {
+
+    private final GoSettings settings;
+    private final Model model;
+    private final SymbolProvider symbolProvider;
+    private final GoWriter writer;
+    private final ServiceShape service;
+    private final OperationShape operation;
+    private final Symbol operationSymbol;
+
+    OperationGenerator(
+            GoSettings settings,
+            Model model,
+            SymbolProvider symbolProvider,
+            GoWriter writer,
+            ServiceShape service,
+            OperationShape operation,
+            Symbol operationSymbol
+    ) {
+        this.settings = settings;
+        this.model = model;
+        this.symbolProvider = symbolProvider;
+        this.writer = writer;
+        this.service = service;
+        this.operation = operation;
+        this.operationSymbol = operationSymbol;
+    }
+
+    @Override
+    public void run() {
+        OperationIndex operationIndex = model.getKnowledge(OperationIndex.class);
+        Symbol serviceSymbol = symbolProvider.toSymbol(service);
+
+        if (!operationIndex.getInput(operation).isPresent()) {
+            // Theoretically this shouldn't ever get hit since we automatically insert synthetic inputs / outputs.
+            throw new CodegenException(
+                    "Operations are required to have input shapes in order to allow for future evolution.");
+        }
+        StructureShape inputShape = operationIndex.getInput(operation).get();
+        Symbol inputSymbol = symbolProvider.toSymbol(inputShape);
+
+        if (!operationIndex.getOutput(operation).isPresent()) {
+            throw new CodegenException(
+                    "Operations are required to have output shapes in order to allow for future evolution.");
+        }
+        StructureShape outputShape = operationIndex.getOutput(operation).get();
+        Symbol outputSymbol = symbolProvider.toSymbol(outputShape);
+
+        writer.writeShapeDocs(operation);
+        Symbol contextSymbol = SymbolUtils.createValueSymbolBuilder("Context", GoDependency.CONTEXT).build();
+        writer.openBlock("func (c $P) $T(ctx $T, params $P, opts ...func(*Options)) ($P, error) {", "}",
+                serviceSymbol, operationSymbol, contextSymbol, inputSymbol, outputSymbol, () -> {
+                    // TODO: create middleware stack
+                    writer.write("options := c.options.Copy()");
+                    writer.openBlock("for _, fn := range optFns {", "}", () -> {
+                        writer.write("fn(&options)");
+                    });
+                    writer.openBlock("for _, fn := range options.APIOptions {", "}", () -> {
+                        writer.write("if err := fn(stack); err != nil { return nil, err }");
+                    });
+                    // TODO: resolve middleware stack
+                    writer.write("return nil, nil");
+                }).write("");
+
+        // Write out the input and output structures. These are written out here to prevent naming conflicts with other
+        // shapes in the model.
+        new StructureGenerator(model, symbolProvider, writer, inputShape, inputSymbol).run();
+
+        // The output structure gets a metadata member added.
+        Symbol metadataSymbol = SymbolUtils.createValueSymbolBuilder(
+                "Metadata", GoDependency.SMITHY_MIDDLEWARE).build();
+        new StructureGenerator(model, symbolProvider, writer, outputShape, outputSymbol).renderStructure(() -> {
+            writer.write("");
+            writer.writeDocs("Metadata pertaining to the operation's result.");
+            writer.write("ResultMetadata $T", metadataSymbol);
+        });
+    }
+}
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
@@ -36,6 +36,7 @@ final class OperationGenerator implements Runnable {
     private final ServiceShape service;
     private final OperationShape operation;
     private final Symbol operationSymbol;
+    private final ApplicationProtocol applicationProtocol;
 
     OperationGenerator(
             GoSettings settings,
@@ -44,7 +45,8 @@ final class OperationGenerator implements Runnable {
             GoWriter writer,
             ServiceShape service,
             OperationShape operation,
-            Symbol operationSymbol
+            Symbol operationSymbol,
+            ApplicationProtocol applicationProtocol
     ) {
         this.settings = settings;
         this.model = model;
@@ -53,6 +55,7 @@ final class OperationGenerator implements Runnable {
         this.service = service;
         this.operation = operation;
         this.operationSymbol = operationSymbol;
+        this.applicationProtocol = applicationProtocol;
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -331,13 +338,32 @@ public Symbol stringShape(StringShape shape) {
...
+            if (boundOperationName.isPresent()) {
+                return SymbolUtils.createPointableSymbolBuilder(shape, name, rootModuleName)
+                        .definitionFile("./api_Op_" + boundOperationName.get())
+                        .build();
+            }
+        }
+
         Symbol.Builder builder = SymbolUtils.createPointableSymbolBuilder(shape, name, typesPackageName);
         if (shape.hasTrait(ErrorTrait.ID)) {
-            builder.definitionFile("./types/errors.go");
```
kstich, line 336: Is there a reason for swapping this to include repeated code?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -311,7 +318,7 @@ public Symbol resourceShape(ResourceShape shape) {
     @Override
     public Symbol serviceShape(ServiceShape shape) {
         // TODO: implement client generation
```
kstich, line 320: Seems like this comment should be removed :)

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
JordonPhillips: That's an aspect of it having been part of the great namespace reconstruction of 2020. It doesn't really matter to me either way though, since it's just a build and return that's duped

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -298,8 +303,10 @@ private Symbol createBigSymbol(Shape shape, String symbolName) {
 
     @Override
     public Symbol operationShape(OperationShape shape) {
-        // TODO: implement operations
-        return SymbolUtils.createPointableSymbolBuilder(shape, "nil").build();
+        String name = getDefaultShapeName(shape);
+        return SymbolUtils.createPointableSymbolBuilder(shape, name, rootModuleName)
+                .definitionFile(String.format("./api_Op_%s.go", name))
```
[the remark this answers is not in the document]
JordonPhillips: The [design](https://github.com/aws/aws-sdk-go-v2/pull/530) has the operation name with preserved casing

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
```diff
@@ -0,0 +1,107 @@
...
+                    writer.write("return nil, nil");
+                }).write("");
+
+        // Write out the input and output structures. These are written out here to prevent naming conflicts with other
+        // shapes in the model.
+        new StructureGenerator(model, symbolProvider, writer, inputShape, inputSymbol).run();
+
+        // The output structure gets a metadata member added.
+        Symbol metadataSymbol = SymbolUtils.createValueSymbolBuilder(
+                "Metadata", GoDependency.SMITHY_MIDDLEWARE).build();
```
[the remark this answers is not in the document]
JordonPhillips: The metadata interface is currently in the middleware package. We could change that, but it should probably be in a different PR

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+        generateConstructor(serviceSymbol);
+
+        String clientId = CodegenUtils.getDefaultPackageImportName(settings.getModuleName());
+        String clientTitle = service.getTrait(TitleTrait.class).map(StringTrait::getValue).orElse(clientId);
+
+        writer.writeDocs("ServiceID returns the name of the identifier for the service API.");
+        writer.write("func (c $P) ServiceName() string { return $S }", serviceSymbol, clientId);
+
+        writer.writeDocs("ServiceName returns the full service title.");
+        writer.write("func (c $P) ServiceID() string { return $S }", serviceSymbol, clientTitle);
```
jasdel, line 79: Nit, the doc strings are on the wrong methods.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+        writer.writeDocs(() -> writer.write(
+                    "New returns an initialized $T based on the functional options.\n"
+                    + "Provide additional functional options to further configure the behavior\n"
+                    + "of the client, such as changing the client's endpoint or adding custom\n"
+                    + "middleware behavior.", serviceSymbol
+        ));
+        writer.openBlock("func New(options $L) ($P, error) {", "}",
+                CONFIG_NAME, serviceSymbol, () -> {
+
+            writer.write("resolvedOptions = options.Copy()").write("");
```
jasdel, line 97: Its find to replace the original `options` value`. since we want to ensure nothing later ever refers to it by mistake.

```go
options = options.Copy()
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+
+            // Run any config initialization functions registered by runtime plugins.
+            for (RuntimeClientPlugin runtimeClientPlugin : runtimePlugins) {
+                writer.write("resolvedOptions, err = $T(resolvedOptions)", runtimeClientPlugin.getResolveFunction());
+                writer.write("if err != nil { return nil, err }").write("");
+            }
```
jasdel, line 103: Minor nit, instead of creating a copy of the options each time, passing in the address would be a little bit better. Since the options value is copied when received from the constructor, additional copies aren't needed. Minor but can also collapse if and function call into same line for brevity.  I think the current generation will actually have compile failure since `err` is not already defined. the following resolves these issues.

```go
if err := $T(&options); err != nil {
     return nil, err
}
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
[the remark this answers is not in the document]
jasdel: Not significant but, the V2 SDK's generation followed the `api_op_GetObject.go` pattern where the operation name was its exported case, but the literal rest was lowercase.

Maybe having a helper exposed by the GoIntegrations that provide these filenames formats?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
JordonPhillips: amazing

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -0,0 +1,168 @@
...
+                    + "middleware behavior.", serviceSymbol
+        ));
+        writer.openBlock("func New(options $L) ($P, error) {", "}",
+                CONFIG_NAME, serviceSymbol, () -> {
+
+            writer.write("options = options.Copy()").write("");
+
+            // Run any config initialization functions registered by runtime plugins.
+            for (RuntimeClientPlugin runtimeClientPlugin : runtimePlugins) {
+                if (!runtimeClientPlugin.matchesService(model, service)
```
kstich, line 104: You'll want to filter this set once in the constructor to pre-compute it when you need other aspects of the `RuntimeClientPlugin` class in here. Fine to fix up when you add that next integration.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
```diff
@@ -0,0 +1,123 @@
...
+                    writer.write("options := c.options.Copy()");
+                    writer.openBlock("for _, fn := range optFns {", "}", () -> {
+                        writer.write("fn(&options)");
+                    });
+                    writer.openBlock("for _, fn := range options.APIOptions {", "}", () -> {
+                        writer.write("if err := fn(stack); err != nil { return nil, err }");
+                    });
+
+                    writer.write("result, err := handler.Handle(ctx, params)");
+                    writer.write("if err != nil { return nil, err }");
```
jasdel, line 96: Is decorating the transport client with middleware stack waiting on protocol specific delegation to be implemented in the generation?

```go
    h := middleware.DecorateHandler(awshttp.ClientHandler{
        Client: options.HTTPClient,
    }, stack)

    res, metadata, err := h.Handle(ctx, input)
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
```diff
@@ -0,0 +1,123 @@
...
+
+                    writer.write("options := c.options.Copy()");
+                    writer.openBlock("for _, fn := range optFns {", "}", () -> {
+                        writer.write("fn(&options)");
+                    });
+                    writer.openBlock("for _, fn := range options.APIOptions {", "}", () -> {
+                        writer.write("if err := fn(stack); err != nil { return nil, err }");
+                    });
+
+                    writer.write("result, err := handler.Handle(ctx, params)");
```
jasdel, line 95: Nit handler will return , `result, metadata, err`.  for the return can do the following to capture the metadata in the result output shape.

```
    out := res.($P)
    out.ResultMetadata = metadata
    
    return out, nil
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
JordonPhillips: Nope, just missed it

PR #32, opened 2020-05-08
Update namespace handling

JordonPhillips:
This updates and fixes the way namespacing is handled, particularly
with regard to aliasing. Previously a SymbolReference was being
created to represent the namespace, but that isn't right since a
namespace isn't an individually referenceable entity. Now the symbol
itself controls how its namespace should be referenced. This ends up
being much more concise to define as an added bonus.


skmcgrail:
This will help a ton, I ran into some weird situations while implementing our document serializers.

PR #35, opened 2020-05-14
Add support for generating protocol tests

jasdel:
Adds support for generating protocol tests defined in the Smithy model.

Generates a `protocol_test.go` file into the API client's package. Ideally this should only be generated if there are actually protocol tests defined.

The generated protocol tests functions current have `nil` in their name until the operation symbol support is available.

* [x] Provide stubs for protocol `request`, `response`, `response error` test generation 
* [x] Query asserts
* [x] required query key asserts
* [x] forbidden query key asserts
* [x] Header asserts
* [x] required header key asserts
* [x] forbidden header key asserts
* [x] compare serialized JSON document body with expectation
* [x] generate struct type's member values from model (e.g input/output types)
* [x] Integrate with API Client generation when merged
    * [x] Client options type value setting could be improved. It is pretty hard coded and makes no account for protocol or flavor specific behavior like AWS sigv4, or mock credentials.
* [x] unit test function names have `nil` until operation symbol support is available.
* [x] Generate assertions of deserialized output shape
* [x] Generate assertions of deserialized error shape

Example `protocol_test.go` file snip of a single operation's modeled `httpRequestTest`.

* [Serialization Test](https://gist.github.com/jasdel/89adb3919e4eecf820d74bac5605f02b#file-serializationtest-go)
* [Deserialization Test](https://gist.github.com/jasdel/89adb3919e4eecf820d74bac5605f02b#file-deserializetest-go)
* [API Error Deserialization Test](https://gist.github.com/jasdel/89adb3919e4eecf820d74bac5605f02b#file-deserializererror-go)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -0,0 +1,358 @@
...
+                break;
+
+            default:
+                scalarWrapShapeValue(writer, shape, () -> {
+                    params.accept(new ShapeValueNodeVisitor(writer, this, shape));
+                });
+        }
+    }
+
+    public void structDeclShapeValue(GoWriter writer, StructureShape shape, Runnable inner) {
```
JordonPhillips, line 85: Do these need to be public?

jasdel: No, can move these to default scope, or protected.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolTestGenerator.java
```diff
@@ -0,0 +1,572 @@
...
+
+    // Only generate test cases when its protocol matches the target protocol.
+    private <T extends HttpMessageTestCase> void onlyIfProtocolMatches(T testCase, Runnable runnable) {
+        if (testCase.getProtocol().equals(settings.getProtocol())) {
+            LOGGER.fine(() -> format("Generating protocol test case for %s.%s", service.getId(), testCase.getId()));
+            runnable.run();
+        }
+    }
+
+//    public void generateRequestTests(List<HttpRequestTestCase> testCases, OperationShape operation) {
```
kstich, line 125: Looks like a ton of code got left behind?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -0,0 +1,367 @@
...
+                writer.writeInline("nil");
+            }
+
+            return null;
+        }
+
+        @Override
+        public Void numberNode(NumberNode node) {
+            switch (currentShape.getType()) {
+                case BOOLEAN:
```
kstich, line 269: The non-[number | timestamp] shape types won't have number nodes representing them. Smithy validates this for the params property of each of the test traits.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -0,0 +1,367 @@
...
+                case INTEGER:
+                case LONG:
+                case FLOAT:
+                case DOUBLE:
+                    writer.writeInline("$L", node.getValue());
+                    break;
+
+                case BIG_INTEGER:
+                    writer.addUseImports(GoDependency.BIG);
+                    writer.writeInline("func() *big.Int {\n"
```
kstich, line 296: Looks like the logic here and for `BIG_DECIMAL` is duplicated with the `stringNode` method, consider abstracting.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestGenerator.java
```diff
@@ -0,0 +1,285 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ *
```
kstich, line 14: Nit: A few extra blank lines here.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolTestGenerator.java
```diff
@@ -0,0 +1,572 @@
...
+    private final Set<String> additionalStubs = new TreeSet<>();
+
+    private final HttpProtocolUnitTestRequestGenerator.Builder requestTestBuilder;
+    private final HttpProtocolUnitTestResponseGenerator.Builder responseTestBuilder;
+    private final HttpProtocolUnitTestResponseErrorGenerator.Builder responseErrorTestBuilder;
+
+    //private ProtocolGenerator.GenerationContext context;
+
+    public HttpProtocolTestGenerator(
+            ProtocolGenerator.GenerationContext context,
```
kstich, line 57: You can import inner classes like this directly:

```
import software.amazon.smithy.go.codegen.integration.ProtocolGenerator.GenerationContext;
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolTestGenerator.java
```diff
@@ -0,0 +1,572 @@
...
+    private final ServiceShape service;
+    private final String protocolName;
+
+    private final Set<String> additionalStubs = new TreeSet<>();
+
+    private final HttpProtocolUnitTestRequestGenerator.Builder requestTestBuilder;
+    private final HttpProtocolUnitTestResponseGenerator.Builder responseTestBuilder;
+    private final HttpProtocolUnitTestResponseErrorGenerator.Builder responseErrorTestBuilder;
+
+    //private ProtocolGenerator.GenerationContext context;
```
kstich, line 54: rm

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
```diff
@@ -0,0 +1,238 @@
...
+        writeHeaderStructField(writer, "ExpectHeader", testCase.getHeaders());
+        writeStringSliceStructField(writer, "RequireHeader", testCase.getRequireHeaders());
+        writeStringSliceStructField(writer, "ForbidHeader", testCase.getForbidHeaders());
+
+        String bodyMediaType = "";
+        if (testCase.getBodyMediaType().isPresent()) {
+            bodyMediaType = testCase.getBodyMediaType().get();
+            writeStructField(writer, "BodyMediaType", "$S", bodyMediaType);
+        }
+        if (testCase.getBody().isPresent()) {
```
kstich, line 108: There's a decent amount of code in here that doesn't vary based on anything being generated for and can live in isolation. I'd recommend moving these out to static files and loading them.

```
IoUtils.readUtf8Resource(getClass(), "protocol-test-empty-comparator.go")
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
```diff
@@ -0,0 +1,238 @@
...
+import software.amazon.smithy.go.codegen.GoDependency;
+import software.amazon.smithy.go.codegen.GoWriter;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.protocoltests.traits.HttpRequestTestCase;
+
+public class HttpProtocolUnitTestRequestGenerator extends HttpProtocolUnitTestGenerator<HttpRequestTestCase> {
+    private static final Logger LOGGER = Logger.getLogger(HttpProtocolUnitTestRequestGenerator.class.getName());
+
+    public static class Builder {
```
kstich, line 32: It doesn't look like you're using this as a builder, but rather as a provider? Can you explain a little about the design here?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -0,0 +1,367 @@
...
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.traits.EnumTrait;
+
+public class ShapeValueGenerator {
+    private static final Logger LOGGER = Logger.getLogger(ShapeValueGenerator.class.getName());
+
+    protected final Model model;
+    protected final SymbolProvider symbolProvider;
+
+    public ShapeValueGenerator(Model model, SymbolProvider symbolProvider) {
```
kstich, line 47: The public interface (abstract functions, public/protected functions, and public classes) should have javadoc as well. This applies throughout the PR.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolTestGenerator.java
jasdel: thanks! yeah this was forgotten code that should of been deleted.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
jasdel: Mainly the idea here was to provide a way to for a builder to be provided to the caller so that the caller can create context specific HttpProtocolUnitTestRequestGenerator  value. This design was used to allow the concrete implementations of `HttpProtocolUnitTestRequestGenerator` to be extended and custom implementations to be used if needed. This also allows as dependency injects in the ProtocolGenerator in aws-sdk-go-v2's codegen, https://github.com/aws/aws-sdk-go-v2/pull/577

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
jasdel: Ended up moving these to static Go code in the smithy-go module.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
```diff
@@ -92,8 +96,12 @@ public void run() {
                         writer.write("if err := fn(stack); err != nil { return nil, err }");
                     });
 
-                    constructHandler();
+                    if (protocolGenerator != null) {
```
JordonPhillips, line 99: This check will need to go into the function itself, as the function is going to be used to add other middleware. Probably not a big deal unless that other PR gets merged first.

jasdel: ^^ Sorry this is based on a PR by @skotambkar that is old and will be removed once his middleware PRs are merged

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
jasdel: Will update PR without that stuff,

kstich: Given the context, I'd recommend moving this to a more standard Java `Builder` pattern. This would have individual methods for setting the components, [like on this class](https://github.com/awslabs/smithy-typescript/blob/0b83c74fd13321ef1ad008185322f9dc5fcd9b75/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/RuntimeClientPlugin.java#L289).

jasdel: Wouldn't that implies that the fields were optional? Whereas a single method specifies all the required members.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -0,0 +1,434 @@
...
+
+            case MAP:
+                mapDeclShapeValue(writer, shape.asMapShape().get(), () -> {
+                    params.accept(new ShapeValueNodeVisitor(writer, this, shape));
+                });
+                break;
+
+            case UNION:
+            case DOCUMENT:
+                // TODO not supported yet
```
skmcgrail, line 91: Should this throw an exception for now?

jasdel: yeah can add throwing exception to make sure we see these later

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -0,0 +1,434 @@
...
+                    writer.writeInline("$L", node.getValue());
+                    break;
+
+                case BIG_INTEGER:
+                    writeInlineBigIntegerInit(writer, node.getValue());
+                    break;
+
+                case BIG_DECIMAL:
+                    writeInlineBigDecimalInit(writer, node.toString());
+                    break;
```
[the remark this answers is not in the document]
jasdel: I think this was a big, this should be a getValue

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestGenerator.java
```diff
@@ -0,0 +1,497 @@
...
+        writer.openBlock("func " + unitTestFuncNameFormat() + "(t *testing.T) {", "}", unitTestFuncNameArgs(),
+                () -> {
+                    generateTestSetup(writer);
+
+                    writer.write("cases := map[string]struct {");
+                    generateTestCaseParams(writer);
+                    writer.openBlock("}{", "}", () -> {
+                        for (T testCase : testCases) {
+                            testCase.getDocumentation().ifPresent(writer::writeDocs);
+                            writer.openBlock("$S: {", "},", testCase.getId(), () -> {
```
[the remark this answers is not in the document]
jasdel: Yeah I'd expect the test cases to be stable, and should be able to support test cases running, or generated in any order. Even transversal of the members of the params document can be arbitrary order since the fields are always referred to by name.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestResponseGenerator.java
```diff
@@ -0,0 +1,180 @@
...
+         * @param operation      shape for the Operation the unit test is generated for.
+         * @param testCases      Set of test cases for the operation.
+         * @return initialize protocol test generator.
+         */
+        public HttpProtocolUnitTestResponseGenerator build(
+                Model model, SymbolProvider symbolProvider, String protocolName, OperationShape operation,
+                List<HttpResponseTestCase> testCases
+        ) {
+            return new HttpProtocolUnitTestResponseGenerator(model, symbolProvider, protocolName, operation, testCases);
+        }
```
skmcgrail, line 54: Usually the builder provide setter functions to set the relevant fields and then the build function takes no arguments and passed the builder to a `HttpProtocolUnitTestResponseGenerator` constructor. If you implement the SmithyBuilder interface you can also assert that certain values were set.

testing/reader.go
```diff
@@ -0,0 +1,39 @@
...
+		p[i] = l.value
+	}
+	return len(p), nil
+}
+
+// Close closes the ByteLoop, and prevents any further reading.
+func (l *ByteLoop) Close() error {
+	l.mu.Lock()
+	l.closed = true
+	l.mu.Unlock()
```
skmcgrail, line 36: nit: defer

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestResponseGenerator.java
jasdel: Was talking with Kevin about this too. Will update these builders to follow that pattern.

testing/rest.go
```diff
@@ -0,0 +1,231 @@
...
+		parts := strings.SplitN(item, `=`, 2)
+		var value string
+		if len(parts) > 1 {
+			value = parts[1]
+		}
+		items = append(items, QueryItem{
+			// Go Query encoder escapes space as `+` whereas smithy protocol
+			// tests expect `%20`.
+			Key:   strings.ReplaceAll(parts[0], `+`, `%20`),
+			Value: strings.ReplaceAll(value, `+`, `%20`),
```
skmcgrail, line 129: Do we need to address this in our encoder?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
skmcgrail: If you have fields that are not optional can't you use the SmithyBuilder interface to assert what is required?

skmcgrail:
Other then some general formatting changes, and the builder pattern note this looks like we can get it merged in.

testing/rest.go
jasdel: Most likely i think so. I'm assuming it is using the default `Encode` method on uri.Value, which does `+` space escaping. Probably need to make a second pass after encode and replace those values with `%20`. I think the sigv4 spec has some expectations with this as well.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestGenerator.java
```diff
@@ -0,0 +1,534 @@
...
+        }
+        return false;
+    }
+
+    public abstract static class Builder<T extends HttpMessageTestCase> {
+        protected Model model;
+        protected SymbolProvider symbolProvider;
+        protected String protocolName;
+        protected OperationShape operation;
+        protected List<T> testCases;
```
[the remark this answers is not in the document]
jasdel: makes sense will update these to default to empty.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
```diff
@@ -0,0 +1,249 @@
...
+        });
+        writer.openBlock("if c.BodyAssert != nil {", "}", () -> {
+            writer.openBlock("if err := c.BodyAssert(actualReq.Body); err != nil {", "}", () -> {
+                writer.write("t.Errorf(\"expect body equal, got %v\", err)");
+            });
+        });
+    }
+
+    public static class Builder extends HttpProtocolUnitTestGenerator.Builder<HttpRequestTestCase> {
+        @Override
```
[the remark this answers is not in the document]
jasdel: Good point this is actually only needed for the ResponseError generator, since it adds an additional field.

PR #36, opened 2020-05-19
Add wrapping generated doc string to a fixed length

jasdel:
Adds a default wrap length to generated doc strings to improve generated
code documentation's readability.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
@@ -47,10 +47,13 @@ import software.amazon.smithy.utils.StringUtils;
 public final class GoWriter extends CodeWriter {
 
     private static final Logger LOGGER = Logger.getLogger(GoWriter.class.getName());
+    private static final int DEFAULT_DOC_WRAP_LENGTH = 80;
 
     private final String fullPackageName;
     private final ImportDeclarations imports = new ImportDeclarations();
     private final List<SymbolDependency> dependencies = new ArrayList<>();
+
+    private int docWrapLength;
 
     public GoWriter(String fullPackageName) {
         this.fullPackageName = fullPackageName;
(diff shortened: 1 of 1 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
```diff
@@ -183,7 +209,7 @@ public GoWriter addDependency(SymbolDependencyContainer dependencies) {
      * @param runnable Runnable that handles actually writing docs with the writer.
      * @return Returns the writer.
      */
-    GoWriter writeDocs(Runnable runnable) {
+    protected GoWriter writeDocs(Runnable runnable) {
```
JordonPhillips, line 212: Why is this protected now? The class is final, so it's not like it will ever have inheritors

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
```diff
@@ -47,18 +47,44 @@
...
+
     public GoWriter(String fullPackageName) {
         this.fullPackageName = fullPackageName;
         trimBlankLines();
         trimTrailingSpaces();
         setIndentText("\t");
         putFormatter('T', new GoSymbolFormatter());
         putFormatter('P', new PointableGoSymbolFormatter());
+
+        this.docWrapLength = DEFAULT_DOC_WRAP_LENGTH;
```
JordonPhillips, line 66: better to just set this at declaration

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
jasdel: Would private be better? Idea was to prevent this version of the method from being called outside of the `GoWriter`.  In Java the default visibility is package right?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
jasdel: will update, thanks.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
JordonPhillips: Yes, default is package private

PR #44, opened 2020-05-26
Fix nested slice and map shape generation

skmcgrail:
Fixes the generation of maps and slices when nested in other map or slice types.

```go
type KitchenSink struct {
	Blob *[]byte
	Boolean *bool
	Double *float64
	EmptyStruct *EmptyStruct
	Float *float32
	HttpdateTimestamp *time.Time
	Integer *int32
	Iso8601Timestamp *time.Time
	JsonValue *string
	ListOfLists [][]*string
	ListOfMapsOfStrings []map[string]*string
	ListOfStrings []*string
	ListOfStructs []*SimpleStruct
	Long *int64
	MapOfListsOfStrings map[string][]*string
	MapOfMaps map[string]map[string]*string
	MapOfStrings map[string]*string
	MapOfStructs map[string]*SimpleStruct
	RecursiveList []*KitchenSink
	RecursiveMap map[string]*KitchenSink
	RecursiveStruct *KitchenSink
	SimpleStruct *SimpleStruct
	String_ *string
	StructWithLocationName *StructWithLocationName
	Timestamp *time.Time
	UnixTimestamp *time.Time
}
```

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
@@ -233,7 +233,9 @@ final class SymbolVisitor implements SymbolProvider, ShapeVisitor<Symbol> {
 
     private Symbol createCollectionSymbol(CollectionShape shape) {
         Symbol reference = toSymbol(shape.getMember());
-        return SymbolUtils.createValueSymbolBuilder(shape, shape.getId().getName())
+        // Shape name will be unused for symbols that represent a slice, but in the event it does we set the collection
+        // shape's name to make debugging simpler.
+        return SymbolUtils.createValueSymbolBuilder(shape, getDefaultShapeName(shape))
                 .putProperty(SymbolUtils.GO_SLICE, true)
                 .putProperty(SymbolUtils.GO_UNIVERSE_TYPE,
                         reference.getProperty(SymbolUtils.GO_UNIVERSE_TYPE, Boolean.class).orElse(false))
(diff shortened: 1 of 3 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -233,22 +233,22 @@ public Symbol setShape(SetShape shape) {
 
     private Symbol createCollectionSymbol(CollectionShape shape) {
         Symbol reference = toSymbol(shape.getMember());
-        return SymbolUtils.createValueSymbolBuilder(shape, reference.getName(), reference.getNamespace())
+        return SymbolUtils.createValueSymbolBuilder(shape, shape.getId().getName())
```
JordonPhillips, line 236: Why this particular change? Also, you'll want to use `getDefaultShapeName`, which handles casing.

skmcgrail: Well the naming is actually meaningless as the `GO_ELEMENT_TYPE` storing the symbol will convey the actual name to be used when it is further unraveled since this is a slice. Honestly the name here could be empty string or null. I chose to use the name of the actual collection shape in case it were to get incorrectly rendered into code.

JordonPhillips: This is probably fine then, if you use the previously mentioned function and add the comment. That name theoretically be real if we were to generate a type alias. The code won't compile if you mess it up, but you'll know where to look since you've got the shape name.

skmcgrail:
Addressed feedback items.

PR #59, opened 2020-06-03
Fixup for smithy generated clients imports

jasdel:
Updates the import paths used by generated smithy API clients. Also minor tweaks to generated code to compile.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
@@ -145,15 +145,53 @@ public final class CodegenUtils {
      * This method can be used by deserializers to get pointer to
      * operand.
      *
+     * @param writer  The writer dependencies will be added to, if needed.
      * @param shape   The shape whose value needs to be assigned.
      * @param operand The Operand is the value to be assigned to the symbol shape.
      * @return The Operand, along with pointer reference if applicable
      */
-    public static String generatePointerReferenceIfPointable(Shape shape, String operand) {
-        if (isShapePassByReference(shape)) {
-            return '&' + operand;
+    public static String generatePointerValueIfPointable(GoWriter writer, Shape shape, String operand) {
+        switch (shape.getType()) {
+            case STRING:
+                if (shape.hasTrait(EnumTrait.class)) {
+                    return operand;
+                }
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.String(" + operand + ")";
+            case BOOLEAN:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Bool(" + operand + ")";
+
+            case BYTE:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Int8(" + operand + ")";
+            case SHORT:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Int16(" + operand + ")";
+            case INTEGER:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Int32(" + operand + ")";
+            case LONG:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Int64(" + operand + ")";
+
+            case FLOAT:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Float32(" + operand + ")";
+            case DOUBLE:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Float64(" + operand + ")";
+
+            case TIMESTAMP:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Time(" + operand + ")";
+
+            default:
+                if (isShapePassByReference(shape)) {
+                    return '&' + operand;
+                }
+                return operand;
         }
-        return operand;
     }
 
     /**
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
@@ -151,40 +151,45 @@ public final class CodegenUtils {
      * @return The Operand, along with pointer reference if applicable
      */
     public static String generatePointerValueIfPointable(GoWriter writer, Shape shape, String operand) {
+        String prefix = "";
+        String suffix = ")";
+
         switch (shape.getType()) {
             case STRING:
                 if (shape.hasTrait(EnumTrait.class)) {
                     return operand;
                 }
-                writer.addUseImports(GoDependency.SMITHY_PTR);
-                return "ptr.String(" + operand + ")";
+
+                prefix = "ptr.String(";
+                break;
+
             case BOOLEAN:
-                writer.addUseImports(GoDependency.SMITHY_PTR);
-                return "ptr.Bool(" + operand + ")";
+                prefix = "ptr.Bool(";
+                break;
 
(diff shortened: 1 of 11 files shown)


io/ringbuffer.go
```diff
@@ -0,0 +1,72 @@
+package sdkio
```
skotambkar, line 1: should the package name still be sdkio?

transport/http/client.go
```diff
@@ -0,0 +1,55 @@
...
+	return ClientHandler{
+		client: client,
+	}
+}
+
+// Handle implements the middleware Handler interface, that will invoke the
+// underlying HTTP client. Requires the input to be an Smithy *Request. Returns
+// a smithy *Response, or error if the request failed.
+func (c ClientHandler) Handle(ctx context.Context, input interface{}) (
+	out interface{}, metadata middleware.Metadata, err error,
```
skotambkar, line 42: out is not used anywhere; out will have rawresponse, so may be we should expose that somehow?

errors.go
```diff
@@ -60,6 +60,21 @@ func (e *OperationError) Error() string {
...
+// deserialization.
+type DeserializationError struct {
+	Err      error //  original error
+	Snapshot []byte
+}
+
+// Error returns a formatted error for DeserializationError
+func (e *DeserializationError) Error() string {
+	return fmt.Sprintf("deserialization failed, %v", e.Err)
+}
```
skotambkar, line 73: Nit: I have the change in my improvement PR, basically this should check if e.Err is nil. Similar to how SerializationError is doing. 

```
// Error returns a formatted error for DeserializationError
func (e *DeserializationError) Error() string {
	const msg = "deserialization failed"
	if e.Err == nil {
		return msg
	}
	return fmt.Sprintf("%s: %v", msg, e.Err)
}

```

io/byte.go
```diff
@@ -0,0 +1,12 @@
+package sdkio
```
skmcgrail, line 1: package name does not match path

io/ringbuffer.go
skmcgrail, line 1: package name should be `io`

io/ringbuffer_test.go
```diff
@@ -0,0 +1,320 @@
+package sdkio
```
skmcgrail, line 1: package name `io`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
```diff
@@ -54,6 +54,8 @@
             "github.com/awslabs/smithy-go/httpbinding", null, Versions.SMITHY_GO),
     SMITHY_JSON("dependency", "github.com/awslabs/smithy-go",
             "github.com/awslabs/smithy-go/json", "smithyjson", Versions.SMITHY_GO),
+    SMITHY_IO("dependency", "github.com/awslabs/smithy-go",
```
skotambkar, line 57: I don't see this being used, but usage might be in another PR. Just want to call out

io/byte.go
skotambkar: Ah had the same comment

transport/http/client.go
jasdel: Still needed for the middleware.Handler interface though. The deserialization middleware is where this `out` value will be used as the RawResponse

errors.go
jasdel: cool will update.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
jasdel: This is used by the aws-sdk-go-v2 pr that i've not posted yet.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -975,34 +980,35 @@ private void writeRestDeserializerMember(
...
-                            CodegenUtils.generatePointerReferenceIfPointable(targetShape, value));
+                            CodegenUtils.generatePointerValueIfPointable(writer, targetShape, value));
                 });
     }
 
     private void writePrefixHeaderDeserializerFunction(
-            GoWriter writer, Model model, String memberName,
-            Shape targetShape, HttpBinding binding
+            GoWriter writer, Model model, SymbolProvider symbolProvider,
+            String memberName, Shape targetShape, HttpBinding binding
```
skmcgrail, line 1011: nit: Style says that for function parameter definitions that span lines should be formatted such that each parameter is on its own line.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -975,34 +980,35 @@ private void writeRestDeserializerMember(
...
             default:
                 throw new CodegenException("unexpected http binding found");
         }
     }
 
     private void writeHeaderDeserializerFunction(
-            GoWriter writer, Model model, String memberName,
-            Shape targetShape, HttpBinding binding
+            GoWriter writer, Model model, SymbolProvider symbolProvider,
+            String memberName, Shape targetShape, HttpBinding binding
```
skmcgrail, line 998: nit: Style says that for function parameter definitions that span lines should be formatted such that each parameter is on its own line.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -936,29 +940,30 @@ private String generateHttpBindingsValue(
...
+                return getCollectionDeserializer(writer, model, symbolProvider, targetValueListShape, binding, operand);
             default:
                 throw new CodegenException("unexpected shape type " + targetShape.getType());
         }
     }
 
     private String getCollectionDeserializer(
-            GoWriter writer, Model model,
+            GoWriter writer, Model model, SymbolProvider symbolProvider,
             Shape targetShape, HttpBinding binding, String operand
```
skmcgrail, line 955: nit: Style says that for function parameter definitions that span lines should be formatted such that each parameter is on its own line.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
jasdel: cool will update to use that style

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
```diff
@@ -145,15 +145,53 @@ public static String getSyntheticTypeNamespace() {
...
+            case FLOAT:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Float32(" + operand + ")";
+            case DOUBLE:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Float64(" + operand + ")";
+
+            case TIMESTAMP:
+                writer.addUseImports(GoDependency.SMITHY_PTR);
+                return "ptr.Time(" + operand + ")";
```
skotambkar, line 187: This feels repetitive, as default seems to handle the behavior.

jasdel: Yeah its repetitive :/ will see if this can be compressed. There are several cases where we don't want to include the extra dep. and the default case has significantly different behavior.

PR #63, opened 2020-06-05
Validation Middleware, Functions and Error Types

skmcgrail:
Code generation for validating member shapes marked as being required. Supports nested shape validation through the use of helpers.

[Sample Generation](https://gist.github.com/skmcgrail/81b9d3b69136791d29a9fed32c3731fc)


skmcgrail:
> Why not implement this as a GoIntegration? It would give you more flexibility in the future if, say, you wanted to add some configuration.

After giving this some though I agree. I've migrated this to a draft as I rearrange some items around. Specifically I want to refactor some of the GoDepdency enum some as to have a better way to allow the AWS integrations to track their own dependencies in the same manner.

PR #64, opened 2020-06-08
Use SDK Smithy branch for Smithy codegen

jasdel:
Sets the Smithy codegen's SDK and smithy-go versions to explicit go.mod hash. This works around the issue of smithy-go's dependencies referring to a version of the SDK that doesn't include the smithy codegen behavior.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
@@ -89,8 +89,8 @@ public enum GoDependency implements SymbolDependencyContainer {
 
     private static final class Versions {
         private static final String GO_STDLIB = "1.14";
-        private static final String SMITHY_GO = "latest";
-        private static final String AWS_SDK = "v0.22.0";
-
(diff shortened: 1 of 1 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
```diff
@@ -89,8 +89,8 @@
 
     private static final class Versions {
         private static final String GO_STDLIB = "1.14";
-        private static final String SMITHY_GO = "latest";
-        private static final String AWS_SDK = "v0.22.0";
-
+        private static final String SMITHY_GO = "v0.0.0-20200604194311-25e885347bc8";
```
skotambkar, line 92: Curious why smithy-go version need to be explicit.

jasdel: smithy-go doesn't have any tagged versions atm, and because the way the go.mod is being updated, an explicit version must be used, either tag or commit. Cannot use branch names/latest with this method. (e.g. `go mod edit --require`)

PR #65, opened 2020-06-08
Fixup REST-JSON payload serializer separating it from document vs payload

jasdel:
Updates the REST-JSON payload serializer generator to use separate generation branch for payload serialization vs document http bindings.

The previous logic was generation payloads as if they were documents and `[]byte` where they would be better suited to be `io.Reader`.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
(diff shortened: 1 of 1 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -389,6 +407,25 @@ protected abstract void writeMiddlewareDocumentSerializerDelegator(
...
+     * Generate the payload serializer logic for the serializer middleware body.
+     *
+     * @param model          the model
+     * @param symbolProvider the symbol provider
+     * @param operation      the operation
+     * @param memberShape    the payload target member
+     * @param generator      middleware generator definition
+     * @param writer         the writer within the middlware context
+     */
+    protected abstract void writeMiddlewarePayloadSerializerDelegator(
```
JordonPhillips, line 420: You could simplify this by just passing around the generationcontext

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
```diff
@@ -92,8 +96,12 @@ public void run() {
                         writer.write("if err := fn(stack); err != nil { return nil, err }");
                     });
 
-                    constructHandler();
+                    if (protocolGenerator != null) {
+                        // add middleware to operation stack
```
skmcgrail, line 100: Shouldn't this just be a generic GoIntegration extension point? This would allow other handlers to be registered in a generic way from plugins.

jasdel: Looks like this PR mistakenly included changes that it shouldn't of. this is a part of https://github.com/awslabs/smithy-go/pull/66 will update the pr to not include those extra changes.

PR #70, opened 2020-06-11
Refactor GoDependency and Dependency Constants

skmcgrail:
This PR refactors the GoDependency type and constants in order to decouple AWS specific dependencies from the smithy repository.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
@@ -15,82 +15,233 @@
 
 package software.amazon.smithy.go.codegen;
 
-import java.util.Collections;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.TreeSet;
 import software.amazon.smithy.codegen.core.SymbolDependency;
 import software.amazon.smithy.codegen.core.SymbolDependencyContainer;
+import software.amazon.smithy.utils.SetUtils;
+import software.amazon.smithy.utils.SmithyBuilder;
 
-/**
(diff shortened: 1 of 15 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDependency.java
```diff
@@ -15,82 +15,233 @@
...
+                    return "stdlib";
+                case DEPENDENCY:
+                    return "dependency";
+                default:
+                    return "unknown";
+            }
+        }
+    }
+
+    public static GoDependency moduleDependency(String sourcePath, String importPath, String version, String alias) {
```
jasdel, line 170: Minor nit, it would be handy for this function to be overridden with a form that doesn't take the `alias` parameter, or defaults `alias` to null.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SmithyGoDependency.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+public enum SmithyGoDependency {
+
+    // The version in the stdlib dependencies should reflect the minimum Go version.
+    // The values aren't currently used, but they could potentially used to dynamically
+    // set the minimum go version.
+    BIG(stdlib("math/big")),
+    TIME(stdlib("time")),
+    FMT(stdlib("fmt")),
+    CONTEXT(stdlib("context")),
```
jasdel, line 29: Would it make sense for these to be public static final members of `SmithyGoDependency` as type `GoDependency` and not enums? If they were that style they could still be treated as if they were enums in our code, and the extra `getDependency` method call wouldn't be needed.

```java
public class SmithyGoDependency {
      public static final GoDependency BIG = stdlib("math/big") // or what ever the equivalent is.
...
}
```

PR #74, opened 2020-06-16
Add config field interface to RuntimeClientPlugin

JordonPhillips:
*Description of changes:*

This changes the way that client config options are added. Now a `RuntimeClientPlugin` can register a number of fields to add to a config. Each option will be added to that config, and a getter will be generated to enable creating interfaces.

For example:

```java
@Override
public List<RuntimeClientPlugin> getClientPlugins() {
    return ListUtils.of(RuntimeClientPlugin.builder()
            .addConfigField(ConfigField.builder()
                    .name("MyConfig")
                    .type(SymbolUtils.createPointableSymbolBuilder("int64")
                            .putProperty(SymbolUtils.GO_UNIVERSE_TYPE, true)
                            .build())
                    .documentation("My config")
                    .build())
            .build());
}
```

```go
type Options struct {
	// Set of options to modify how an operation is invoked. These apply to all
	// operations invoked for this client. Use functional options on operation call to
	// modify this list for per operation behavior.
	APIOptions []APIOptionFunc

	// My config
	MyConfig *int64
	
	// The HTTP client to invoke API calls with. Defaults to client's default HTTP
	// implementation if nil.
	HTTPClient HTTPClient
}

func (o Options) GetMyConfig() *int64 {
	return o.MyConfig
}
```

~~EDIT: this was supposed to be a draft - there are still a few touch ups before this could be merged, e.g. removing the writer from GoIntegration and adding docs to ConfigField~~

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
@@ -16,8 +16,11 @@
 package software.amazon.smithy.go.codegen;
 
 import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
+import software.amazon.smithy.go.codegen.integration.ConfigField;
 import software.amazon.smithy.go.codegen.integration.GoIntegration;
 import software.amazon.smithy.go.codegen.integration.RuntimeClientPlugin;
 import software.amazon.smithy.model.Model;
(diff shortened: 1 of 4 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -125,13 +128,26 @@ private void generateConfig() {
             );
             writer.write("APIOptions []$L", API_OPTIONS_FUNC_NAME).write("");
 
-            for (GoIntegration integration : integrations) {
-                integration.addConfigFields(settings, model, symbolProvider, writer);
+            // Add config fields to the options struct.
+            for (ConfigField configField : getAllConfigFields()) {
+                configField.getDocumentation().ifPresent(writer::writeDocs);
+                writer.write("$L $P", configField.getName(), configField.getType());
```
skmcgrail, line 134: Should we generate option types as values rather then pointers? Additionally could we maybe make that configurable via the config field? `GetForatter` or something similar to determine $T vs $P

JordonPhillips: Discussed offline, but this can be controlled by the pointable property of the symbol you're given.

PR #78, opened 2020-06-18
Add Go build to finalize codegen step, and fixup error generation

jasdel:
Adds go build to the set of commands run after the code generation completes. This ensures that the generated code compiles. Both test and regular code.

Fixes up error shape generation so that exception getters methods return the expected type and use the smithy-go/ptr pkg for converting scalar pointers to values.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -199,6 +199,10 @@ void execute() {
 
         LOGGER.fine("Running go fmt");
         CodegenUtils.runCommand("go fmt", fileManifest.getBaseDir());
+
+        // TODO this should be moved to a smithy gradle task
+        LOGGER.fine("Running go build");
+        CodegenUtils.runCommand("go test -run NONE ./...", fileManifest.getBaseDir());
```
skotambkar, line 205: What's the use of this?

PR #80, opened 2020-06-18
Support HTTP RPC protocols

JordonPhillips:
*Description of changes:*

This add support for http rpc protocols. There is some refactoring work done here, mostly in ProtocolUtils, to reduce copied code, but existing code is as of yet still intact.

related: https://github.com/aws/aws-sdk-go-v2/pull/599

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
@@ -33,6 +33,7 @@ public abstract class HttpRpcProtocolGenerator implements ProtocolGenerator {
 
     private final boolean isErrorCodeInBody;
     private final Set<Shape> serializingDocumentShapes = new TreeSet<>();
+    private final Set<Shape> deserializingDocumentShapes = new TreeSet<>();
 
     /**
      * Creates a Http RPC protocol generator.
@@ -198,6 +199,23 @@ public abstract class HttpRpcProtocolGenerator implements ProtocolGenerator {
     );
 
     @Override
+    public void generateSharedDeserializerComponents(GenerationContext context) {
+        deserializingDocumentShapes.addAll(ProtocolUtils.resolveRequiredDocumentShapeSerde(
+                context.getModel(), deserializingDocumentShapes));
+        generateDocumentBodyShapeDeserializers(context, deserializingDocumentShapes);
+    }
+
+    /**
+     * Generated deserialization functions for shapes in the passed set. These functions
+     * should return a value that can then be serialized by the implementation of
+     * {@code deserializeOutputDocument}.
+     *
+     * @param context The generation context.
+     * @param shapes The shapes to generate deserialization for.
+     */
+    protected abstract void generateDocumentBodyShapeDeserializers(GenerationContext context, Set<Shape> shapes);
+
+    @Override
     public void generateResponseDeserializers(GenerationContext context) {
         TopDownIndex topDownIndex = context.getModel().getKnowledge(TopDownIndex.class);
         Set<OperationShape> containedOperations = new TreeSet<>(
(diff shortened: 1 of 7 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
```diff
@@ -0,0 +1,327 @@
...
+import software.amazon.smithy.go.codegen.GoStackStepMiddlewareGenerator;
+import software.amazon.smithy.go.codegen.GoWriter;
+import software.amazon.smithy.go.codegen.SmithyGoDependency;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.shapes.StructureShape;
+
+public abstract class HttpRpcProtocolGenerator implements ProtocolGenerator {
```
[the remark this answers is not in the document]
JordonPhillips: Query and EC2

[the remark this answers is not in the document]
jasdel: Didn't know that Query and EC2 were considered RPC good to know :)  Seems like there is a lot of behavior between Http and HttpRpc. Can we collapse these two more, or should they stay separate?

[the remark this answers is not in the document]
JordonPhillips: > Didn't know that Query and EC2 were considered RPC

Well they don't support http binding traits 🤷‍♂️

In the refactoring stage I'm gonna see what else I can pull out. The initial handling of the request type and response type for instance. There were a few things I noted in the other PR around the json decoder that can also probably be pulled out.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
```diff
@@ -0,0 +1,329 @@
...
+        ApplicationProtocol applicationProtocol = getApplicationProtocol();
+        Symbol requestType = applicationProtocol.getRequestType();
+        GoWriter writer = context.getWriter();
+
+        middleware.writeMiddleware(context.getWriter(), (generator, w) -> {
+            writer.addUseImports(SmithyGoDependency.SMITHY);
+            writer.addUseImports(SmithyGoDependency.FMT);
+            writer.addUseImports(SmithyGoDependency.SMITHY_HTTP_BINDING);
+
+            // Cast the input request to the transport request type and check for errors.
```
kstich, line 107: It might be useful to [generate a single function](https://github.com/awslabs/smithy-typescript/blob/master/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpRpcProtocolGenerator.java#L98:L129) that handles preparing the request and [just invoking that with the altered contents](https://github.com/awslabs/smithy-typescript/blob/master/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpRpcProtocolGenerator.java#L182:L186). A lot of this content looks static enough to warrant it, and there'll be a lot of duplicated code other wise.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
```diff
@@ -0,0 +1,329 @@
...
+            writer.write("");
+
+            writer.write("response, ok := out.RawResponse.($P)", responseType);
+            writer.openBlock("if !ok {", "}", () -> {
+                writer.write(String.format("return out, metadata, &smithy.DeserializationError{Err: %s}",
+                        "fmt.Errorf(\"unknown transport type %T\", out.RawResponse)"));
+            });
+            writer.write("");
+
+            writer.openBlock("if response.StatusCode < 200 || response.StatusCode >= 300 {", "}", () -> {
```
kstich, line 253: This looks like it's routing redirections through the error routing. Make sure this is handled properly downstream, as you may start retrying responses that contain valid redirections.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
```diff
@@ -0,0 +1,329 @@
...
+
+            deserializeOutputDocument(context, operation);
+            deserializingDocumentShapes.add(ProtocolUtils.expectOutput(model, operation));
+            writer.write("");
+
+            writer.write("return out, metadata, err");
+        });
+        writer.write("");
+
+        Set<StructureShape> errorShapes = HttpProtocolGeneratorUtils.generateErrorDispatcher(
```
kstich, line 269: It doesn't look like the error dispatcher is handling the private `isErrorCodeInBody` property that's used for some implementing protocols. That property just looks set on the constructor and not utilized.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
JordonPhillips: Yeah I think the only real catch here is needing to check for status code 0 which is something go will do in some cases.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
JordonPhillips: I'm not sure if it matters here. Unlike TS, the go dispatcher is never attempting to parse the error body, only copying it into a seekable type. If it happens that it's empty, nothing changes. Perhaps we should just get rid of the property

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
JordonPhillips: Looks like the go http client automatically follows redirects

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
JordonPhillips: Minor context: Go v1 apparently would preparse (json) into an empty interface map, but that was fairly memory intensive.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
JordonPhillips: Hmmm maybe. I can probably make it work for RPC but I'm not sure about rest since the encoder needs to be passed around. I will say that duplicated code doesn't matter *that* much in the generated go, where the binary size is less important than performance.

JordonPhillips: Looked into this a bit more. Even for RPC this would require a pretty large overhaul of the way the http binding encoder works. Probably worth doing, so I'll add a comment about it, but I think it would be out of scope.

PR #82, opened 2020-06-19
Fixes Serializer Middleware to correctly handle Smithy URI containing query parameters

skmcgrail:
Fixes serializer middleware to correctly handle Smithy URI containing query parameters

### DIFF
--- a/httpbinding/uri.go
+++ b/httpbinding/uri.go
@@ -80,8 +80,8 @@ func (u URIValue) BigDecimal(v *big.Float) error {
 	return u.modifyURI(v.Text('e', -1))
 }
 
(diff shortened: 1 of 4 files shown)


httpbinding/uri.go
```diff
@@ -78,3 +79,20 @@ func (u URIValue) BigDecimal(v *big.Float) error {
 	}
 	return u.modifyURI(v.Text('e', -1))
 }
+
+// ParseURI parses a Smithy HTTP binding trait URI
+func ParseURI(uri string) (path, query string) {
```
jasdel, line 84: Nit should this be `split` instead of parse since not really parsing the query string?

PR #91, opened 2020-06-24
Add sort order to implementations of GoIntegrations

jasdel:
Adds sorting order to GoIntegrations to help improve consistent code generation order.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/IdempotencyTokenMiddlewareGenerator.java
```diff
@@ -67,6 +67,17 @@ private void execute(
         });
     }
 
+    /**
+	 * Gets the sort order of the customization from -128 to 127, with lowest
+	 * executed first.
```
JordonPhillips, line 72: One space back.

jasdel: Thanks missed that when updated comments,

PR #93, opened 2020-06-24
Add bypassing document in protocol tests until supported

jasdel:
Adds bypassing document members in protocol tests until supported, and Updated smithy-go's `Document` type to be simple interface until implemented.


document.go
```diff
@@ -0,0 +1,8 @@
+package smithy
```
skotambkar, line 1: Is moving this to smithy package a temporary change? Or why should this be in smithy package rather than document package?

skotambkar: I guess you answered this in PR description.

jasdel: Yeah after the design iterations, the basic `Document` type should stay in the smithy pkg. Concrete implementations probably will be in their own package, e.g. JSON and ION

PR #97, opened 2020-06-26
Fixup header value, list, and prefix (de)serialization

jasdel:
Updates how HTTP header bindings are generated for (de)serialization to correctly serialize and deserialize header prefix, header values, and lists. 

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
@@ -995,6 +995,7 @@ public abstract class HttpBindingProtocolGenerator implements ProtocolGenerator 
             String operand
     ) {
         if (targetShape.getType() != ShapeType.LIST && targetShape.getType() != ShapeType.SET) {
+            writer.addUseImports(SmithyGoDependency.STRINGS);
             writer.write("$L = strings.TrimSpace($L)", operand, operand);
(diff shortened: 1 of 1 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -1176,18 +1195,35 @@ private void writePrefixHeaderDeserializerFunction(
...
-            writer.openBlock("if val := response.Header.Get($S); val != $S {",
-                    "}", locationName, "", () -> {
-                        String value = generateHttpBindingsValue(writer, model, symbolProvider,
-                                targetValueShape, binding, "val");
-                        writer.write("v.$L[$S] = $L", memberName, name,
-                                CodegenUtils.generatePointerValueIfPointable(writer, targetValueShape, value));
+
+        MemberShape valueMemberShape = targetShape.asMapShape()
+                .orElseThrow(() -> new CodegenException("prefix headers must target map shape"))
+                .getValue();
```
skotambkar, line 1201: We can also use         

`MapShape mapshape = model.expectShape(targetShape.getId(), MapShape.class);`

then you won't have to explicitly throw an exception. 

 I think even `targetShape.asMapShape().get()` would throw an exception directly.

jasdel: cool will switch to the expectShape call

PR #101, opened 2020-06-30
Update SDKs middleware generation utils to separate name from id

jasdel:
Updates the helpers for generating SDK middleware in a client package to separate the middleware name from the ID. The SDK generated middleware generic names were added.

This change does not create constants or other methods of organizing the IDs.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStackStepMiddlewareGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStackStepMiddlewareGenerator.java
@@ -85,6 +85,7 @@ public final class GoStackStepMiddlewareGenerator {
      * Create a new SerializeStep middleware generator with the provided type name.
      *
      * @param name is the type name to identify the middleware.
+     * @param id   the unique ID for the middleware.
      * @return the middleware generator.
      */
     public static GoStackStepMiddlewareGenerator createSerializeStepMiddleware(String name, String id) {
@@ -100,6 +101,7 @@ public final class GoStackStepMiddlewareGenerator {
      * Create a new DeserializeStep middleware generator with the provided type name.
      *
      * @param name is the type name to identify the middleware.
+     * @param id   the unique ID for the middleware.
      * @return the middleware generator.
      */
     public static GoStackStepMiddlewareGenerator createDeserializeStepMiddleware(String name, String id) {
@@ -116,6 +118,7 @@ public final class GoStackStepMiddlewareGenerator {
      * Generates a new step middleware generator.
      *
      * @param name              the name of the middleware type.
+     * @param id   the unique ID for the middleware.
      * @param handlerMethodName method name to be implemented.
      * @param inputType         the middleware input type.
      * @param outputType        the middleware output type.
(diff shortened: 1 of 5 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStackStepMiddlewareGenerator.java
```diff
@@ -78,11 +84,12 @@ public static GoStackStepMiddlewareGenerator createBuildStepMiddleware(String id
     /**
      * Create a new SerializeStep middleware generator with the provided type name.
      *
-     * @param identifier is the type name to identify the middleware.
+     * @param name is the type name to identify the middleware.
```
JordonPhillips, line 87: Gotta add the docs for id

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -243,8 +243,9 @@ private void addOperationDocumentShapeBindersForSerializer(GenerationContext con
 
     private void generateOperationSerializerMiddleware(GenerationContext context, OperationShape operation) {
         GoStackStepMiddlewareGenerator middleware = GoStackStepMiddlewareGenerator.createSerializeStepMiddleware(
-                ProtocolGenerator.getSerializeMiddlewareName(operation.getId(), getProtocolName()
-                ));
+                ProtocolGenerator.getSerializeMiddlewareName(operation.getId(), getProtocolName()),
+                "OperationSerializer"
```
JordonPhillips, line 247: Why not just have the functions set these so we don't have to populate this everywhere? If nothing else, these should probably be public static strings somewhere.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStackStepMiddlewareGenerator.java
```diff
@@ -64,11 +68,13 @@ public static GoStackStepMiddlewareGenerator createInitalizeStepMiddleware(Strin
...
+     * @param id   the unique ID for the middleware.
      * @return the middleware generator.
      */
-    public static GoStackStepMiddlewareGenerator createBuildStepMiddleware(String identifier) {
-        return createMiddleware(identifier,
+    public static GoStackStepMiddlewareGenerator createBuildStepMiddleware(String name, String id) {
+        return createMiddleware(name,
+                id,
```
skotambkar, line 77: How will this behavior be replicated for non generated middlewares? i.e. making sure they always have an id too. 

Should we have an interface contract to make sure middlewares have an id defined.

jasdel: Non-generated middleware will have to be set when they are defined. That is outside of the scope of this since this making the generated middleware bit more generic.

jasdel: Also sorry forgot to answer the first part of your question. The `ID` is an expected member of the smithy Middleware interface types, https://github.com/awslabs/smithy-go/blob/master/middleware/step_build.go#L28-L39

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
jasdel: yeah can move this to a more generic method for getting the generator for the operation serializer middleware

jasdel: Ended up moving these to constants. We're holding off publishing these as constants within the Go code until we have a better understanding of user story for helpers that will be created for adding and removing specific middleware. Middleware IDs like these might end up being specified as consts in the SDK's core library.

PR #102, opened 2020-07-01
Fixup/idempotency token tests

jasdel:
Adds a default implementation of the IdempotencyTokenProvider to the smithy-go rand pkg.

### DIFF
--- a/rand/uuid.go
+++ b/rand/uuid.go
@@ -5,6 +5,23 @@ import (
 	"io"
 )
 
+// UUIDIdempotencyToken provides a utility to get idempotency tokens in the
+// UUID format.
+type UUIDIdempotencyToken struct {
+	uuid *UUID
+}
+
+// NewUUIDIdempotencyToken returns a idempotency token provider returning
+// tokens in the UUID random format using the reader provided.
+func NewUUIDIdempotencyToken(r io.Reader) *UUIDIdempotencyToken {
+	return &UUIDIdempotencyToken{uuid: NewUUID(r)}
+}
+
+// GetIdempotencyToken returns a random UUID value for Idempotency token.
+func (u UUIDIdempotencyToken) GetIdempotencyToken() (string, error) {
+	return u.uuid.GetUUID()
+}
+
 // UUID provides computing random UUID version 4 values from a random source
 // reader.
 type UUID struct {


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/IdempotencyTokenMiddlewareGenerator.java
```diff
@@ -202,14 +216,25 @@ private String getIdempotencyTokenMiddlewareHelperName(OperationShape operationS
...
     }
 
+    /**
+     * Returns if there are any operations within the service that use idempotency token auto fill trait.
+     *
+     * @param model   Model used for generation.
+     * @param service Service for which idempotency token map is retrieved.
+     * @return if operations use idempotency token auto fill trait.
+     */
+    public static boolean haOperationsWithIdempotencyTokens(Model model, ServiceShape service) {
```
skotambkar, line 226: Nit: ha`\s\`OperationsWithIdempotencyTokens

rand/uuid.go
```diff
@@ -5,6 +5,23 @@ import (
...
+// tokens in the UUID random format using the reader provided.
+func NewUUIDIdempotencyToken(r io.Reader) *UUIDIdempotencyToken {
+	return &UUIDIdempotencyToken{uuid: NewUUID(r)}
+}
+
+// GetIdempotencyToken returns a random UUID value for Idempotency token.
+func (u UUIDIdempotencyToken) GetIdempotencyToken() (string, error) {
+	return u.uuid.GetUUID()
+}
+
```
skotambkar, line 24: Shouldn't this be used when in `func(o *Options) GetIdempotencyToken() (string, error)`, don't see it being used. Can you verify it's generated?

jasdel: Ah so the generated helper doesn't need to use a  `GetIdempotencyToken` interface because the helper is defined in the same package as the Client. This allows the helper function to make use of the `Options` type directly instead of being forced to use an interface.  https://github.com/aws/aws-sdk-go-v2/pull/628/files?file-filters%5B%5D=.go#diff-9ced4025b14d09f3f49aafe35890b89eR91-R94

This is the reason there is no `GetIdempotencyToken()` call.

skotambkar: Gotcha. I think NewUUIDIdempotencyToken will be used as default value when we have default client config options.

jasdel:
I think i might have to split this into two different PRs, first for Go module update then second for Java PR. CI are failing because the smithy-go module versions are invalid.

PR #108, opened 2020-07-09
Add middleware to set HTTP request content length

jasdel:
Adds middleware that will set the HTTP request's Content-Length header if the length of the serialized stream can be determined. If the length cannot be determined or the Header value is already set the middleware will do nothing.

Fails the request if the underlying request stream reader is seekable but fails to seek.

### DIFF
--- a/transport/http/middleware_content_length.go
+++ b/transport/http/middleware_content_length.go
@@ -42,7 +42,8 @@ func (m *ContentLengthMiddleware) HandleBuild(
 	if n, ok, err := req.StreamLength(); err != nil {
 		return out, metadata, fmt.Errorf(
 			"failed getting length of request stream, %w", err)
-	} else if ok {
+	} else if ok && n > 0 {
+		// Only set content-length header when it is a positive value.
 		req.Header.Set("Content-Length", strconv.FormatInt(n, 10))
 	}
 
--- a/transport/http/middleware_content_length.go
+++ b/transport/http/middleware_content_length.go
@@ -1,0 +1,50 @@
+package http
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/awslabs/smithy-go/middleware"
+)
+
+// ContentLengthMiddleware provides a middleware to set the content-length
+// header for the length of a serialize request body.
+type ContentLengthMiddleware struct {
+}
+
+// AddContentLengthMiddleware adds ContentLengthMiddleware to the middleware
+// stack's Build step.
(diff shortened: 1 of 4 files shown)


transport/http/middleware_content_length.go
```diff
@@ -0,0 +1,50 @@
...
+) (
+	out middleware.BuildOutput, metadata middleware.Metadata, err error,
+) {
+	req, ok := in.Request.(*Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown request type %T", req)
+	}
+
+	// Don't set content length if header is already set.
+	if vs := req.Header.Values("Content-Length"); len(vs) != 0 {
```
skotambkar, line 38: We have an open ticket that says Go SDK will include a ```content-type``` even when ```Content-Length=0``` in header. 

If it's the case, should we remove that header here?

transport/http/request.go
```diff
@@ -38,6 +38,32 @@ func (r *Request) Clone() *Request {
...
+	s := r.stream.(io.Seeker)
+	endOffset, err := s.Seek(0, io.SeekEnd)
+	if err != nil {
+		return 0, false, err
+	}
+
+	_, err = s.Seek(r.streamStartPos, io.SeekStart)
+	if err != nil {
+		return 0, false, err
+	}
```
skotambkar, line 62: Can you comment an example for when streamStartPos < io.seekStart? 

I thought io.SeekStart will always be less than streamStartPos.

jasdel: `SeekStart` will start from the beginning of the stream and seek `streamStartPos` bytes forward.  The reason to seek to `streamStartPos` instead of `0` is to ensure that the SDK only sends the stream from the starting position the user's application provided it to the SDK at. For example application opens a file, and wants to skip the first N bytes uploading the rest. The application would move the file's offset N bytes, then hand it off to the SDK to send the remaining.

transport/http/middleware_content_length.go
jasdel: Will take a look at this. Removing `Content-Type` probably should be its own middleware. Its kind of contextual though, because an for example its value for a S3 PutObject operation call to send a Content-Type=text/plain for a zero length object.  Maybe we should only remove 


Though this does remind me that the SDK's signer must not include `Content-Length` as a signed header if the value is `0` because intermittent network hops can remove the header if its value is zero.

jasdel: Created https://github.com/awslabs/smithy-go/issues/110 to track removing Content-Type when length is 0.

PR #121, opened 2020-07-22
Strip insignificant whitespace from form bodies and export NewQueryValue

JordonPhillips:
This updates the url form equal function to strip out insignificant whitespace, which is whitespace that is not percent encoded.

I've also updated `NewQueryValue` to be exported so it can be used in the query serializer.

Needed for: https://github.com/aws/aws-sdk-go-v2/pull/652

### DIFF
--- a/httpbinding/query.go
+++ b/httpbinding/query.go
@@ -14,7 +14,7 @@ type QueryValue struct {
 	append bool
 }
 
-func newQueryValue(query url.Values, key string, append bool) QueryValue {
+func NewQueryValue(query url.Values, key string, append bool) QueryValue {
 	return QueryValue{
 		query:  query,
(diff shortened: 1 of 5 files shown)


httpbinding/query.go
```diff
@@ -14,7 +14,7 @@ type QueryValue struct {
 	append bool
 }
 
-func newQueryValue(query url.Values, key string, append bool) QueryValue {
+func NewQueryValue(query url.Values, key string, append bool) QueryValue {
```
jasdel, line 17: Did you intend to export this function? It doesn't look like it is used by any other package.

JordonPhillips: Yes I intended to - I'll be using it in the query serializer

PR #123, opened 2020-07-23
Sort consistently in http binding generator

JordonPhillips:
This updates the http binding generator to iterate on sorted collections whenever code is being generated so that the generated output is stable. This results in way less noise when generating code that gets committed.

This also removes a few instances of duplicated code.


JordonPhillips:
related: https://github.com/aws/aws-sdk-go-v2/pull/658

jasdel:
can run `make test-protocols test-services` on `aws-sdk-go-v2` now to validate everything generated compiles.

JordonPhillips:
tests passed!

PR #134, opened 2020-08-17
Adds Get method for stack steps. Adds User-Agent Builder.

skmcgrail:
* Adds a `Get` method for retrieving middleware by a given ID from a stack step.
* Adds an HTTP `User-Agent` builder.

### DIFF
--- a/httpbinding/user_agent.go
+++ b/httpbinding/user_agent.go
@@ -7,12 +7,12 @@ import (
 
 // UserAgentBuilder is a builder for a HTTP User-Agent string.
 type UserAgentBuilder struct {
-	sb *strings.Builder
+	sb strings.Builder
 }
 
 // NewUserAgentBuilder returns a new UserAgentBuilder.
 func NewUserAgentBuilder() *UserAgentBuilder {
-	return &UserAgentBuilder{sb: &strings.Builder{}}
+	return &UserAgentBuilder{sb: strings.Builder{}}
 }
 
 // AddKey adds the named component/product to the agent string
--- a/httpbinding/user_agent.go
+++ b/httpbinding/user_agent.go
@@ -1,0 +1,35 @@
+package httpbinding
+
+import (
+	"fmt"
+	"strings"
+)
+
+// UserAgentBuilder is a builder for a HTTP User-Agent string.
+type UserAgentBuilder struct {
+	sb *strings.Builder
+}
+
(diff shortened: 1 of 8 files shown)


httpbinding/user_agent.go
```diff
@@ -0,0 +1,38 @@
...
+
+// UserAgentBuilder is a builder for a HTTP User-Agent string.
+type UserAgentBuilder struct {
+	sb *strings.Builder
+}
+
+// NewUserAgentBuilder returns a new UserAgentBuilder.
+func NewUserAgentBuilder() *UserAgentBuilder {
+	return &UserAgentBuilder{sb: &strings.Builder{}}
+}
```
jasdel, line 16: not too important but don't really need to take a pointer for `strings.Builder` since its a member of a struct that is a pointer.

httpbinding/user_agent_test.go
```diff
@@ -0,0 +1,12 @@
...
+
+import "testing"
+
+func TestUserAgentBuilder(t *testing.T) {
+	b := NewUserAgentBuilder()
+	b.AddKeyValue("foo", "1.2.3")
+	b.AddKey("baz")
+	if e, a := "foo/1.2.3 baz", b.Build(); e != a {
+		t.Errorf("expect %v, got %v", e, a)
+	}
```
skotambkar, line 11: Nit: Are multiple entries possible? If so, should we add a simple test for them?

httpbinding/user_agent.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// UserAgentBuilder is a builder for a HTTP User-Agent string.
+type UserAgentBuilder struct {
+	sb strings.Builder
+}
```
skotambkar, line 11: Curious if user agent is only added for http based protocol. That seems incorrect, so probably not?

skmcgrail: HTTP is the only protocol we support, since we don't have a standard for other protocols and their user-agents it would be premature to make the utility generic.

httpbinding/user_agent_test.go
skmcgrail: AddKey and AddKeyValue are the only two methods, and this is a sufficient test to cover them. There is no logic that limits the number that can be added.

PR #139, opened 2020-08-19
 Add operation middleware to close response body

jasdel:
Adds additional RuntimePlugin for HTTP protocols that will ensure the HTTP response body is closed when it is no longer needed by the operation. e.g. after deserialize.

Also keeps the body open in the case of streaming operation responses, if there was not an error.

The generated API client looks like:

```diff
diff --git a/internal/protocoltest/restxml/api_op_ConstantAndVariableQueryString.go b/internal/protocoltest/restxml/api_op_ConstantAndVariableQueryString.go
index f64117ab..360c933f 100644
--- a/internal/protocoltest/restxml/api_op_ConstantAndVariableQueryString.go
+++ b/internal/protocoltest/restxml/api_op_ConstantAndVariableQueryString.go
@@ -20,13 +20,15 @@ func (c *Client) ConstantAndVariableQueryString(ctx context.Context, params *Con
        for _, fn := range optFns {
                fn(&options)
        }
+       addawsRestxml_serdeOpConstantAndVariableQueryStringMiddlewares(stack)
        awsmiddleware.AddRequestInvocationIDMiddleware(stack)
        smithyhttp.AddContentLengthMiddleware(stack)
        AddResolveEndpointMiddleware(stack, options)
        retry.AddRetryMiddlewares(stack, options)
        awsmiddleware.AddAttemptClockSkewMiddleware(stack)
+       smithyhttp.AddErrorCloseResponseBodyMiddleware(stack)
+       smithyhttp.AddCloseResponseBodyMiddleware(stack)
        stack.Initialize.Add(newServiceMetadataMiddleware_opConstantAndVariableQueryString(options.Region), middleware.Before)
-       addawsRestxml_serdeOpConstantAndVariableQueryStringMiddlewares(stack)
```

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/CloseResponseMiddlewareGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/CloseResponseMiddlewareGenerator.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ *
+ *
+ */
+
+package software.amazon.smithy.go.codegen.integration;
+
+import java.util.List;
+import java.util.Optional;
+import software.amazon.smithy.go.codegen.SmithyGoDependency;
+import software.amazon.smithy.go.codegen.SymbolUtils;
+import software.amazon.smithy.model.knowledge.HttpBinding;
+import software.amazon.smithy.model.knowledge.HttpBindingIndex;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.traits.StreamingTrait;
+import software.amazon.smithy.utils.ListUtils;
+
+public class CloseResponseMiddlewareGenerator implements GoIntegration {
+    /**
+     * Gets the sort order of the customization from -128 to 127, with lowest
+     * executed first.
+     *
+     * @return Returns the sort order, defaults to -40.
+     */
+    @Override
+    public byte getOrder() {
+        return 127;
+    }
+
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins() {
+        return ListUtils.of(
+                // Add deserialization middleware to close the response in case of errors.
+                RuntimeClientPlugin.builder()
+                        .servicePredicate((model, service) -> {
+                            // TODO is HTTP based protocol
+                            return true;
+                        })
+                        .registerMiddleware(MiddlewareRegistrar.builder()
+                                .resolvedFunction(SymbolUtils.createValueSymbolBuilder(
+                                        "AddErrorCloseResponseBodyMiddleware", SmithyGoDependency.SMITHY_HTTP_TRANSPORT)
+                                        .build())
+                                .build()
+                        )
+                        .build(),
+
+                // Add deserialization middleware to close the response for non-output-streaming operations.
+                RuntimeClientPlugin.builder()
+                        .servicePredicate((model, service) -> {
+                            // TODO is HTTP based protocol
+                            return true;
+                        })
+                        .operationPredicate((model, service, operation) -> {
+                            // TODO operation output NOT event stream
+
+                            // Don't auto close response body when response is streaming.
+                            HttpBindingIndex httpBindingIndex = model.getKnowledge(HttpBindingIndex.class);
+                            Optional<HttpBinding> payloadBinding = httpBindingIndex.getResponseBindings(operation,
+                                    HttpBinding.Location.PAYLOAD).stream().findFirst();
+                            if (payloadBinding.isPresent()) {
+                                MemberShape memberShape = payloadBinding.get().getMember();
+                                Shape payloadShape = model.expectShape(memberShape.getTarget());
+
+                                return !payloadShape.hasTrait(StreamingTrait.class);
+                            }
+
+                            return true;
+                        })
+                        .registerMiddleware(MiddlewareRegistrar.builder()
+                                .resolvedFunction(SymbolUtils.createValueSymbolBuilder(
+                                        "AddCloseResponseBodyMiddleware",
+                                        SmithyGoDependency.SMITHY_HTTP_TRANSPORT)
+                                        .build())
+                                .build()
+                        )
+                        .build()
+        );
+    }
+}
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpCloseResponseMiddleware.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpCloseResponseMiddleware.java
@@ -19,36 +19,40 @@ package software.amazon.smithy.go.codegen.integration;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.function.BiPredicate;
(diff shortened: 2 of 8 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/CloseResponseMiddlewareGenerator.java
```diff
@@ -0,0 +1,92 @@
...
+public class CloseResponseMiddlewareGenerator implements GoIntegration {
+    /**
+     * Gets the sort order of the customization from -128 to 127, with lowest
+     * executed first.
+     *
+     * @return Returns the sort order, defaults to -40.
+     */
+    @Override
+    public byte getOrder() {
+        return 127;
```
skotambkar, line 40: Are we sure we won't need any customization after response body is closed?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/CloseResponseMiddlewareGenerator.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins() {
+        return ListUtils.of(
+                // Add deserialization middleware to close the response in case of errors.
+                RuntimeClientPlugin.builder()
+                        .servicePredicate((model, service) -> {
+                            // TODO is HTTP based protocol
```
skotambkar, line 49: Why is this a TODO?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/CloseResponseMiddlewareGenerator.java
jasdel: Yeah we absolutely could, but this can be updated to reflect relevant ordering.

I picked 127 (last) so that it can put it self as the very first middleware in the deserializer step. Which will make it the last one to be invoked during deserialization of the response.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolGeneratorUtils.java
```diff
@@ -60,8 +60,6 @@ private HttpProtocolGeneratorUtils() {}
         writer.openBlock("func $L(response $P) error {", "}", errorFunctionName, responseType, () -> {
             writer.addUseImports(SmithyGoDependency.BYTES);
             writer.addUseImports(SmithyGoDependency.IO);
-            writer.write("defer response.Body.Close()");
```
skotambkar, line 63: This was just for errors. Do we have to do it for each response deser?

jasdel: The AddErrorCloseResponseBodyMiddleware  helper adds the middleware for all operation responses, and will close response body in case of any response error.

The AddCloseResponseBodyMiddleware is only used for successful responses (which is also why it is only applied to API operations without streaming responses.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/CloseResponseMiddlewareGenerator.java
jasdel: Wasn't sure how to detect that the protocol for the service is an HTTP one. I'd prefer not to do a check on explicit protocol names. I need to reach out to smithy to get some guidence on this.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolGeneratorUtils.java
skotambkar: Still won't we have to remove the explicit response body close statements used by the deserializers?

jasdel: I thought i removed all of them as well, but i'll got back and check to see if i missed any.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/CloseResponseMiddlewareGenerator.java
jasdel: changes this to a utility that is added by the SDK's protocols.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/CloseResponseMiddlewareGenerator.java
jasdel: Updated to be added explicitly by the SDK's `AddProtocols` so don't need to detect is http anymore.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpCloseResponseMiddleware.java
```diff
@@ -0,0 +1,93 @@
...
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.model.traits.StreamingTrait;
+import software.amazon.smithy.utils.ListUtils;
+
+/**
+ * Provides a set of RuntimePlugins to ensure that a HTTP response is closed when it needs to be needed. This should be
+ * used by all HTTP based protocols to ensure that the response body is closed, and any errors are checked. This
+ * ensures that connections are not leaked by the underlying HTTP client.
+ */
+public final class HttpCloseResponseMiddleware {
```
JordonPhillips, line 39: You could put this in HttpProtocolUtils (+0)

transport/http/middleware_close_response_body.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+
+	"github.com/awslabs/smithy-go/middleware"
+)
+
+// AddErrorCloseResponseBodyMiddleware adds the middleware to automatically
+// close the response body of an operation request if the request response
+// failed.
+func AddErrorCloseResponseBodyMiddleware(stack *middleware.Stack) error {
+	return stack.Deserialize.Insert(&errorCloseResponseBodyMiddleware{}, "OperationDeserializer", middleware.Before)
```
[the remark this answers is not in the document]
jasdel: yeah ... we need to close out on that, to decide where that should go, the idea could exist in smithy somewhere but would need to make sure to avoid circular dependencies what ever pkg it went into.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpCloseResponseMiddleware.java
jasdel: Good point, renamed this class to HttpProtocolUtils, and renamed the `getClientPlugins` to `getCloseResponseClientPlugins`.

I wasn't sure if it was better to keep the `getClientPlugins` method generic since the class was renamed to be generic.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUtils.java
```diff
@@ -0,0 +1,93 @@
...
+
+
+    /**
+     * Returns a set of RuntimePlugs to close the HTTP operation response. Uses the servicePredicate parameter to
+     * filter the RuntimePlugins to protocols that are relevant.
+     *
+     * @param servicePredicate service filter
+     * @return RuntimePlugins
+     */
+    public static List<RuntimeClientPlugin> getCloseResponseClientPlugins(BiPredicate<Model, ServiceShape> servicePredicate) {
```
skotambkar, line 51: Line length more than 120, breaks smithy build. Also how about naming this, 
`getClientPluginsToCloseResponse`. would that be non-idiomatic?

jasdel: thanks, yeah realized i was pushing to the wrong branch and didn't see why it was failing.

PR #146, opened 2020-08-20
Update unions

JordonPhillips:
This updates generated unions. Now there will only be a single parent interface, and each member will only be a struct that implements that interface.

This also adds generation of unions to the protocol test generator.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
@@ -142,7 +142,7 @@ public final class ShapeValueGenerator {
                         symbol.getNamespace()
                 ).build();
 
-                writer.writeInline("$T{Value: ", memberSymbol);
+                writer.writeInline("&$T{Value: ", memberSymbol);
                 if (target instanceof SimpleShape) {
                     writeScalarValueInline(writer, target, entry.getValue());
                 } else {
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
@@ -138,7 +138,7 @@ public final class ShapeValueGenerator {
             if (member.isPresent()) {
                 Shape target = model.expectShape(member.get().getTarget());
                 Symbol memberSymbol = SymbolUtils.createValueSymbolBuilder(
-                        String.format("%sMember%s", symbol.getName(), symbolProvider.toMemberName(member.get())),
+                        symbolProvider.toMemberName(member.get()),
                         symbol.getNamespace()
                 ).build();
 
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
@@ -17,6 +17,8 @@
 
 package software.amazon.smithy.go.codegen;
 
+import java.util.Map;
+import java.util.Optional;
 import java.util.logging.Logger;
 import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.Symbol;
@@ -36,6 +38,7 @@ import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.ShapeType;
 import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.shapes.UnionShape;
 import software.amazon.smithy.model.traits.EnumTrait;
 import software.amazon.smithy.model.traits.StreamingTrait;
 
@@ -98,6 +101,9 @@ public final class ShapeValueGenerator {
                 break;
 
             case UNION:
+                unionDeclShapeValue(writer, shape.asUnionShape().get(), params.expectObjectNode());
+                break;
+
             case DOCUMENT:
                 LOGGER.warning("Skipping " + shape.getType() + " shape type not supported, " + shape.getId());
                 writer.writeInline("nil");
@@ -124,6 +130,33 @@ public final class ShapeValueGenerator {
         writer.write("&$T{", symbol);
         inner.run();
         writer.writeInline("}");
+    }
+
+    /**
+     * Writes the declaration for a Go union.
+     *
+     * @param writer writer to write generated code with.
+     * @param shape  the union shape.
+     * @param params the params.
+     */
+    protected void unionDeclShapeValue(GoWriter writer, UnionShape shape, ObjectNode params) {
+        Symbol symbol = symbolProvider.toSymbol(shape);
+        for (Map.Entry<StringNode, Node> entry : params.getMembers().entrySet()) {
+            Optional<MemberShape> member = shape.getMember(entry.getKey().toString());
+            if (member.isPresent()) {
+                Shape target = model.expectShape(member.get().getTarget());
+                String exportedMemberName = symbol.getName() + symbolProvider.toMemberName(member.get());
+                Symbol castingFunction = SymbolUtils.createValueSymbolBuilder(
+                        "As" + exportedMemberName,
+                        symbol.getNamespace()
+                ).build();
+
+                writer.writeInline("$T(", castingFunction);
+                writeShapeValueInline(writer, target, entry.getValue());
+                writer.writeInline(")");
+            }
+            return;
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/UnionGenerator.java
```diff
@@ -51,48 +52,41 @@ public void run() {
...
+            writer.openBlock("type $L struct {", "}", exportedMemberName, () -> {
+                // Union members can't have null values, so for simple shapes we don't
+                // use pointers. We have to use pointers for complex shapes since,
+                // for example, we could still have a map that's empty or which has
+                // null values.
+                if (target instanceof SimpleShape) {
+                    writer.write("Value $T", memberSymbol);
+                } else {
+                    writer.write("Value $P", memberSymbol);
+                }
```
jasdel, line 72: Not specific to this PR, but at some point we should take a look at how slice/map types are generated from `$T`, and `$P`. If a member is or is not a pointer of a slice/map should not be driven by `$T`, or `$P`, the member being a pointer is driven by the modeled list/map shape, and its member.

This mean that `$T` should generate the correct Go type for a map of strings, i.e. `map[string]*string`. In addition the same for lists (assuming lists cannot be sparse, i.e. `[]string`. Using `$P` should generate the pointer type of the outer shape, `*map[string]*string` and `*[]string`.  `$P` should have no impact on the generate of nested members within collections.

jasdel: created, https://github.com/awslabs/smithy-go/issues/149 for this.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -79,66 +83,90 @@ public void writeShapeValueInline(GoWriter writer, Shape shape, Node params) {
...
+        for (Map.Entry<StringNode, Node> entry : params.getMembers().entrySet()) {
+            Optional<MemberShape> member = shape.getMember(entry.getKey().toString());
+            if (member.isPresent()) {
+                Shape target = model.expectShape(member.get().getTarget());
+                Symbol memberSymbol = SymbolUtils.createValueSymbolBuilder(
+                        symbol.getName() + symbolProvider.toMemberName(member.get()),
+                        symbol.getNamespace()
+                ).build();
+
+                writer.writeInline("$T{Value: ", memberSymbol);
```
jasdel, line 145: Should this `$T` be a `&$T` since union's interface method? e.g. `is<UnionShapeName` should be a pointer receiver to prevent any confusion code bugs when users try to extract a union member from the interface. Without the pointer receiver for `is<UnionShapeName` user doesn't know if the SDK did pointer or not for the union member wrapper types, and they'll need to test both.

```go
switch tv := range somUnion.(type) {
    case *FooMemberBar:
         fmt.Printf("%T is pointer, %v", tv, tv.Value)

    case FooMemberBar:
         fmt.Printf("%T is value, %v", tv, tv.Value)
}
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/UnionGenerator.java
```diff
@@ -51,48 +52,41 @@ public void run() {
...
-            writer.openBlock("type $L struct {", "}", unExportedMemberName, () -> {
-                writer.write("value $T", memberSymbol);
-            });
-
-            writer.write("func ($L) is$L() {}", unExportedMemberName, symbol.getName());
-            writer.write("func ($L) is$L() {}", unExportedMemberName, exportedMemberName);
-            writer.openBlock("func (v $L) Value() $T {", "}", unExportedMemberName, memberSymbol, () -> {
-                writer.write("return v.value");
-            });
+            writer.write("func ($L) is$L() {}", exportedMemberName, symbol.getName());
```
jasdel, line 75: We probably want to do a pointer receiver here for this method instead of value. Pointer receiver ensures that when user does type assert for this union member wrapper, they don't need to compare against both value and pointer types. This would prevents the guess work, guessing if the interface value was a pointer or value.

I think typed errors used this pattern of pointer receivers as well.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/UnionGenerator.java
```diff
@@ -51,48 +52,41 @@ public void run() {
...
+            writer.write("Tag string");
             // The value received.
-            writer.write("value []byte");
+            writer.write("Value []byte");
         });
-        writer.write("func (v $LUnknown) is$L() {}", symbol.getName(), symbol.getName())
-                .write("func (v $LUnknown) Value() []byte { return v.value }", symbol.getName())
-                .write("func (v $LUnknown) Tag() string { return v.tag }", symbol.getName());
+
+        writer.write("func (v $L) is$L() {}", unknownStructName, symbol.getName());
```
jasdel, line 90: Is it possible to only generate a single one of these unknown member structs, instead of one for ever union defined?  The definition between them will not change. Each API client's type package wopuld need its own `UnknownUnionMember` struct generated, with a method for each union type in the model added.

```go
// UnknownUnionMember
type UnknownUnionMember struct {
    Tag string
    Value []byte
}
```

For each union in an  API a `is` method would be generated, e.g. for unions `Foo`, `Baz`, `Bar`.

```go
func (UnknownUnionMember) isFoo() {}
func (UnknownUnionMember) isBaz() {}
func (UnknownUnionMember) isBar() {}
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/UnionGenerator.java
```diff
@@ -51,48 +52,41 @@ public void run() {
             writer.write("is$L()", symbol.getName());
         }).write("");
 
-        // Create type aliases for each member that satisfy the interface.
+        // Create structs for each member that satisfy the interface.
         for (MemberShape member : shape.getAllMembers().values()) {
             Symbol memberSymbol = symbolProvider.toSymbol(member);
             String exportedMemberName = symbol.getName() + symbolProvider.toMemberName(member);
```
jasdel, line 58: What do you think about adding the quantifier `Member` between the Union shape name and the member name?

e.g. `FooMemberBar`

JordonPhillips: Sure

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/UnionGenerator.java
```diff
@@ -51,48 +54,56 @@ public void run() {
...
         }).write("");
 
-        // Create type aliases for each member that satisfy the interface.
+        // Create structs for each member that satisfy the interface.
         for (MemberShape member : shape.getAllMembers().values()) {
             Symbol memberSymbol = symbolProvider.toSymbol(member);
-            String exportedMemberName = symbol.getName() + symbolProvider.toMemberName(member);
-            String unExportedMemberName = StringUtils.uncapitalize(exportedMemberName);
+            String exportedMemberName = String.format(
+                    "%sMember%s", symbol.getName(), symbolProvider.toMemberName(member));
```
jasdel, line 61: would be nice to be a helper function for this pattern, so easier to ensure its consistent.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ProtocolUtils.java
```diff
@@ -160,4 +162,15 @@ public static void writeSafeMemberAccessor(
...
+    /**
+     * Determines whether a given shape will use a pointer when the shape is used as a union value.
+     *
+     * @param shape the shape to check
+     * @return true if the shape should use pointers
+     */
+    public static boolean usesPointerWhenUnionValue(Shape shape) {
+        return !(shape instanceof SimpleShape) || shape.isBlobShape() || shape.hasTrait(EnumTrait.class)
+                || shape.hasTrait(StreamingTrait.class);
+    }
```
jasdel, line 175: I'm potentially reading this wrong, but would this conditional make `struct`, `blob` and `enum`, and `streaming` shape value member types pointers? Or is this an artifact how the code writer generates list/maps?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -111,11 +118,29 @@
...
+     * <p>These have the format {UnionName}Member{MemberName}.
+     *
+     * @param model The model whose unions should be reserved.
+     * @param builder A reserved words builder to add on to.
+     */
+    private void reserveUnionMemberNames(Model model, ReservedWordsBuilder builder) {
+        model.shapes(UnionShape.class).forEach(union -> {
+            for (MemberShape member : union.getAllMembers().values()) {
+                String memberName = String.format("%sMember%s",
+                        getDefaultShapeName(union), getDefaultMemberName(member));
```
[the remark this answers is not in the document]
JordonPhillips: I moved that directly into the symbol visitor - where it probably should have been anyway.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ProtocolUtils.java
JordonPhillips: whoops you're right i think i have this backwards

PR #152, opened 2020-08-25
Fix Content-Length Behavior; Define GoIntegration for ServiceId

skmcgrail:
Fixes the Content-Length behavior by correctly utilizing the net/http.Request member for setting the value. The header value if set directly is ignored by the standard library. This was a regression from the original AWS SDK for Go V1/V2 behavior.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
@@ -113,7 +113,7 @@ final class OperationGenerator implements Runnable {
                     writer.openBlock("if err != nil {", "}", () -> {
                         writer.addUseImports(SmithyGoDependency.SMITHY);
                         writer.openBlock("return nil, &smithy.OperationError{", "}", () -> {
-                            writer.write("ClientID: c.ClientID(),");
+                            writer.write("ServiceID: ServiceID,");
                             writer.write("OperationName: \"$T\",", operationSymbol);
                             writer.write("Err: err,");
                         });
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
@@ -113,7 +113,7 @@ final class OperationGenerator implements Runnable {
                     writer.openBlock("if err != nil {", "}", () -> {
                         writer.addUseImports(SmithyGoDependency.SMITHY);
                         writer.openBlock("return nil, &smithy.OperationError{", "}", () -> {
-                            writer.write("ServiceID: c.ServiceID(),");
+                            writer.write("ClientID: c.ClientID(),");
                             writer.write("OperationName: \"$T\",", operationSymbol);
                             writer.write("Err: err,");
                         });
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
@@ -26,8 +26,6 @@ import software.amazon.smithy.go.codegen.integration.GoIntegration;
 import software.amazon.smithy.go.codegen.integration.RuntimeClientPlugin;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.shapes.ServiceShape;
-import software.amazon.smithy.model.traits.StringTrait;
-import software.amazon.smithy.model.traits.TitleTrait;
 
 /**
  * Generates a service client and configuration.
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
@@ -66,7 +66,7 @@ final class ServiceGenerator implements Runnable {
 
     @Override
     public void run() {
-        String serviceId = CodegenUtils.getDefaultPackageImportName(settings.getModuleName());
+        String serviceId = settings.getService().toString();
         for (GoIntegration integration : integrations) {
             serviceId = integration.processServiceId(settings, model, serviceId);
         }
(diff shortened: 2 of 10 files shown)


httpbinding/encode.go
```diff
@@ -4,9 +4,12 @@ import (
...
 	"net/url"
+	"strconv"
 	"strings"
 )
 
+const contentLengthHeader = "Content-Length"
+
 // An Encoder provides encoding of REST URI path, query, and header components
 // of an HTTP request. Can also encode a stream as the payload.
 //
```
skotambkar, line 15: May be call out the behavior for content-length header here. Just for visibility.

skmcgrail: Documented it on `Encode` call specifically.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
```diff
@@ -113,7 +113,7 @@ public void run() {
                     writer.openBlock("if err != nil {", "}", () -> {
                         writer.addUseImports(SmithyGoDependency.SMITHY);
                         writer.openBlock("return nil, &smithy.OperationError{", "}", () -> {
-                            writer.write("ServiceID: c.ServiceID(),");
+                            writer.write("ClientID: c.ClientID(),");
```
jasdel, line 116: Since the type is already named `Client` is `ClientID` redundant?  could this method be changed to just `ID`, or removed all together, and just use the constant from the package?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -76,14 +82,8 @@ public void run() {
...
-        String clientId = CodegenUtils.getDefaultPackageImportName(settings.getModuleName());
-        String clientTitle = service.getTrait(TitleTrait.class).map(StringTrait::getValue).orElse(clientId);
-
-        writer.writeDocs("ServiceID returns the name of the identifier for the service API.");
-        writer.write("func (c $P) ServiceID() string { return $S }", serviceSymbol, clientId);
-
-        writer.writeDocs("ServiceName returns the full service title.");
-        writer.write("func (c $P) ServiceName() string { return $S }", serviceSymbol, clientTitle);
+        writer.writeDocs("ClientID returns the name of the identifier for the service API.");
+        writer.write("func (c $P) ClientID() string { return ClientID }", serviceSymbol);
```
jasdel, line 86: What do you think about removing this method in favor of just the exported constant?  Currently we don't have any interfaces that would make use of the `ClientID` method.  If we did in the future, maybe that is something we could add back if there is need.

errors.go
```diff
@@ -42,13 +42,13 @@ var _ APIError = (*GenericAPIError)(nil)
...
 type OperationError struct {
-	ServiceID     string
+	ClientID      string
 	OperationName string
 	Err           error
 }
 
 // Service returns the name of the API service the error occurred with.
-func (e *OperationError) Service() string { return e.ServiceID }
+func (e *OperationError) Service() string { return e.ClientID }
```
jasdel, line 51: Does it make sense to refactor the `OperationError.Service` method to be `Client` instead of service? or is it better to stick with `Service` in this context?

transport/http/middleware_content_length.go
```diff
@@ -34,17 +33,11 @@ func (m *ContentLengthMiddleware) HandleBuild(
...
-		return next.HandleBuild(ctx, in)
-	}
-
 	if n, ok, err := req.StreamLength(); err != nil {
 		return out, metadata, fmt.Errorf(
 			"failed getting length of request stream, %w", err)
 	} else if ok && n > 0 {
-		// Only set content-length header when it is a positive value.
-		req.Header.Set("Content-Length", strconv.FormatInt(n, 10))
+		req.ContentLength = n
```
jasdel, line 40: Should this check if the `ContentLength` value  no `0` as well? and if it is not `0` don't set it? Kind of an edge case.

There is a potential conflict with the user providing a value for a `ContentLength` modeled input  member shape that conflicts with the derived length of the input stream. In similar cases like content-sha256, the middleware skips this step if the field already has a value.

errors.go
skmcgrail: Ah this was a miss when I auto-renamed, I missed that this method was named differently. I don't have a strong preference between `Service` vs `Client`, I just think we want to be consistent, and likely want to choose a name that is both generic and not Amazon specific. Maybe we should stick with `Service`?. As this matches with the smithy naming scheme of a `service` shape.

jasdel: ok, yeah lets stick with `service` naming.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/OperationGenerator.java
jasdel: If sticking with `service` naming, this could be `ServiceID` or be removed all together and just have the const.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/GoIntegration.java
```diff
@@ -190,4 +190,16 @@ default void processFinalizedModel(GoSettings settings, Model model) {
...
+     * Processes the given serviceId and may return a unmodified, modified, or replacement value.
+     *
+     * @param settings Settings used to generate
+     * @param model model to generate from
+     * @param serviceId the serviceId
+     * @return the new serviceId
+     */
+    default String processServiceId(GoSettings settings, Model model, String serviceId) {
+        return serviceId;
+    }
```
skotambkar, line 204: Should this method not take in a serviceID, and instead retrieve it from settings? That would be more like a `getProcessedServiceID` rather than `processServiceID`. Currently it doesn't seem to do anything than return the provided value.

```
return CodegenUtils.getDefaultPackageImportName(settings.getModuleName());
```

@JordonPhillips might be able to provide his insight on this.

skotambkar: This seems similar to `getSerializedMemberName` func used by serde where we fetch member name and process it based on available traits.

skmcgrail: The issue is that multiple integrations will be executed and will have this method called to give them a chance to process the serviceId and adjust it. There is an integration in the aws-sdk-v2 repo that makes use of this.

skmcgrail: Primarily my issue with this is that if an integration relies on the default implementation, it can override an integration that previously processed and redefined the serviceId.

PR #160, opened 2020-08-27
Add Header Helper Middleware

skmcgrail:
Adds `AddHeaderValue` and `SetHeaderValue` stack mutators that allow easy manipulation of HTTP request headers.

### DIFF
--- a/transport/http/middleware_headers.go
+++ b/transport/http/middleware_headers.go
@@ -1,0 +1,88 @@
+package http
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/awslabs/smithy-go/middleware"
+)
+
+type headerValue struct {
+	header string
+	value  string
+	append bool
+}
+
+type headerValueHelper struct {
+	headerValues []headerValue
+}
+
+func (h *headerValueHelper) addHeaderValue(value headerValue) {
+	h.headerValues = append(h.headerValues, value)
+}
+
+func (h *headerValueHelper) ID() string {
+	return "headerValueHelper"
+}
+
+func (h *headerValueHelper) HandleBuild(ctx context.Context, in middleware.BuildInput, next middleware.BuildHandler) (out middleware.BuildOutput, metadata middleware.Metadata, err error) {
+	req, ok := in.Request.(*Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown transport type %T", in.Request)
+	}
+
+	for _, value := range h.headerValues {
+		if value.append {
+			req.Header.Add(value.header, value.value)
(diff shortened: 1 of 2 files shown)


transport/http/middleware_headers.go
```diff
@@ -0,0 +1,91 @@
...
+		} else {
+			req.Header.Set(value.header, value.value)
+		}
+	}
+
+	return next.HandleBuild(ctx, in)
+}
+
+func getOrAddHeaderValueHelper(stack *middleware.Stack) (*headerValueHelper, error) {
+	id := (*headerValueHelper)(nil).ID()
```
skotambkar, line 49: Is this casting or type asserting? We have used similar statements to assert the type, but not sure how this works here. 

Wouldn't this panic as headerValueHelper is nil?

transport/http/middleware_headers.go
```diff
@@ -0,0 +1,91 @@
...
+func SetHeaderValue(header string, value string) func(stack *middleware.Stack) error {
+	return func(stack *middleware.Stack) error {
+		helper, err := getOrAddHeaderValueHelper(stack)
+		if err != nil {
+			return err
+		}
+		helper.addHeaderValue(headerValue{header: header, value: value, append: false})
+		return nil
+	}
+}
```
skotambkar, line 91: Would it be helpful if this function also returns the existing values if present?

transport/http/middleware_headers.go
jasdel: This would be initialization of a `nil` value of `headerValueHelper`. This is a way to call a method on a type, without initialing that type first. Since `ID` uses a pointer receiver, and does not use any member of `headerValueHelper` the `nil` type value will never be de-referenced. So no panic.

jasdel: Alternatives would be to create a variable of `headerValueHelper` and call ID separatly.

```go
var ider headerValueHelper

id := ider.ID()
```

transport/http/middleware_headers.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"context"
+	"crypto/rand"
+	"fmt"
+
+	"github.com/awslabs/smithy-go/middleware"
+)
+
+var randReader = rand.Reader
```
jasdel: is this needed?

transport/http/middleware_headers.go
skmcgrail: I would argue no, the intention is that if you call set you are explicitly opting to clear any existing value. Additionally the helpers are meant to be plug-able into APIFuncs easily.

PR #178, opened 2020-09-21
Adds support for returning metadata from error deser function

skotambkar:
*Issue #, if available:*

- Current design does not return any metadata that is deserialized from a response error. 

*Description of changes:*

- We update the top level error deserializer function signature to return metadata. 


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
```diff
@@ -249,7 +249,11 @@ private void generateOperationDeserializer(GenerationContext context, OperationS
             writer.write("");
 
             writer.openBlock("if response.StatusCode < 200 || response.StatusCode >= 300 {", "}", () -> {
-                writer.write("return out, metadata, $L(response)", errorFunctionName);
+                writer.write("deserMetadata, err := $L(response)", errorFunctionName);
```
jasdel, line 252: These deserializer functions probably should take `metadata` as a pointer input, `&metadata`, instead of returning a new one that will need to be merged into the stack's metadata.

jasdel (changes requested):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
skotambkar: I didn't do that as that change would in the smithy-go client and would force all smithy-go clients to take metadata as input for top level error deser func

I think returning a empty interface{} allows us to keep it more generic, especially because we have a `processErrorResponseMetadata` method follow by error deserialization.

With the functions returning an empty interface, we can keep aws specific implementation separate and not dictate every other Smithy-go client.

jasdel (approved):

PR #179, opened 2020-09-21
Support endpoint and hostLabel traits

JordonPhillips:
This adds support for the endpoint and hostLabel traits. This is
accomplished by adding a `HostPrefix` member to the Smithy Request
struct, which is prepended to the host when the underlying http
Request is finalized.

If there are no labels, the host prefix is simply generated as a
a string that is set on the smithy request like so:

```go
request.HostPrefix = "data-"
```

If there are labels, a string builder is used:

```go
var prefix strings.Builder
prefix.WriteString("data-")
if input.ServiceName == nil {
	return out, metadata, &smithy.SerializationError{Err: fmt.Errorf("ServiceName forms part of the endpoint host and so may not be nil")}
} else if !smithyhttp.ValidateHostLabel(*input.ServiceName) {
	return out, metadata, &smithy.SerializationError{Err: fmt.Errorf("ServiceName forms part of the endpoint host and so must match \"[a-zA-Z0-9-]{1,63}\", but was \"%s\"", *input.ServiceName)}
} else {
	prefix.WriteString(*input.ServiceName)
}
request.HostPrefix = prefix.String()
```

While the need of filling out a template is similar to what is done
with the URL path, the restriction of `hostLabel` to string shapes
that are required means that we don't need such a heavy-weight
solution. Since this solution is minimal, the strategy can easily
be changed if needed.

Additionally, this was inlined into the operation serializer for
the sake of simplicity. You necessarily need access to the operation
input struct, so if this were a middleware then it would need to be
generated.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolGeneratorUtils.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolGeneratorUtils.java
@@ -169,16 +169,26 @@ public final class HttpProtocolGeneratorUtils {
 
         // If the pattern has labels, we need to build up the host prefix using a string builder.
         writer.addUseImports(SmithyGoDependency.STRINGS);
+        writer.addUseImports(SmithyGoDependency.SMITHY_HTTP_TRANSPORT);
+        writer.addUseImports(SmithyGoDependency.FMT);
         writer.write("var prefix strings.Builder");
         for (Segment segment : pattern.getSegments()) {
             if (!segment.isLabel()) {
                 writer.write("prefix.WriteString($S)", segment.toString());
             } else {
                 MemberShape member = input.getMember(segment.getContent()).get();
-                String memberReference = "input." + symbolProvider.toMemberName(member);
+                String memberName = symbolProvider.toMemberName(member);
+                String memberReference = "input." + memberName;
 
                 // Theoretically this should never be nil by this point unless validation has been disabled.
-                writer.openBlock("if $L != nil {", "}", memberReference, () -> {
+                writer.write("if $L == nil {", memberReference).indent();
+                writer.write("return out, metadata, &smithy.SerializationError{Err: "
+                        + "fmt.Errorf(\"$L forms part of the endpoint host and so may not be nil\")}", memberName);
+                writer.dedent().write("} else if !smithyhttp.ValidateHostLabel(*$L) {", memberReference).indent();
+                writer.write("return out, metadata, &smithy.SerializationError{Err: "
+                        + "fmt.Errorf(\"$L forms part of the endpoint host and so must match \\\"[a-zA-Z0-9-]{1,63}\\\""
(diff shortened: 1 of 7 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolGeneratorUtils.java
```diff
@@ -131,4 +136,36 @@ public static boolean isShapeWithResponseBindings(Model model, Shape shape, Http
...
+        StringBuilder builder = new StringBuilder();
+        for (Segment segment : pattern.getSegments()) {
+            if (!segment.isLabel()) {
+                builder.append(segment.toString());
+            } else {
+                builder.append("%s");
+            }
+        }
+
+        writer.writeInline("request.HostPrefix = fmt.Sprintf($S", builder.toString());
```
jasdel, line 161: An alternative to building this prefix string at runtime it would be better to use codegen to boild this down to the bare minimum for runtime. An better alternative to Sprintf is `strings.Builder`.  So could code generate the statements in Go that do a `strings.Builder.WriteString` for each of the prefix's `Segments`, and label values from input parameters.

e.g. code generated for an operation with a label.  modeled host prefix: `{FooBar}.`

```go
var prefix strings.Builder
prefix.WriteString(*input.FooBar)
prefix.WriteString(".")

request.HostPrefix = prefix.String()
```

If there are no labels the java code gen could just branch and do a single static string assignment instead of using the strings.Builder at all. modeled host prefix: `blah.`

```go
request.HostPrefix = "blah."
```

jasdel (changes requested):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolGeneratorUtils.java
```diff
@@ -131,4 +136,54 @@ public static boolean isShapeWithResponseBindings(Model model, Shape shape, Http
...
+            if (!segment.isLabel()) {
+                writer.write("prefix.WriteString($S)", segment.toString());
+            } else {
+                MemberShape member = input.getMember(segment.getContent()).get();
+                String memberReference = "input." + symbolProvider.toMemberName(member);
+
+                // Theoretically this should never be nil by this point unless validation has been disabled.
+                writer.openBlock("if $L != nil {", "}", memberReference, () -> {
+                    writer.write("prefix.WriteString(*$L)", memberReference);
+                });
```
jasdel, line 183: This probably should return an error if the member is nil, its possible for this to fail silently. This is one of those checks that probably should be required even if validation is disabled, like URL path label replaces. e.g. GetObject with validation disabled and no-bucket, will result in an operation that does a ListObjects instead.

Also this should validate that the member being used is a valid host label, (V1's implementation for reference)

https://github.com/aws/aws-sdk-go/blob/master/private/protocol/host.go#L51-L68

skmcgrail (approved):

PR #180, opened 2020-09-21
Add RequestSendError And CanceledError to HTTP ClientHandler

jasdel:
Adds wrapping the HTTP client handler's response with an `RequestSendError` for all HTTP request send errors, and `CanceledError` if the context passed in to the handler was canceled and the HTTP client encountered an error.

### DIFF
--- a/transport/http/client.go
+++ b/transport/http/client.go
@@ -63,13 +63,17 @@ func (c ClientHandler) Handle(ctx context.Context, input interface{}) (
 	return &Response{Response: resp}, metadata, err
 }
 
-// RequestSendError provides a generic request transport error.
+// RequestSendError provides a generic request transport error. This error
+// should wrap errors making HTTP client requests.
+//
+// The ClientHandler will wrap the HTTP client's error if the client request
+// fails, and did not fail because of context canceled.
 type RequestSendError struct {
 	Err error
 }
 
 // ConnectionError return that the error is related to not being able to send
-// the request.
+// the request, or receive a response from the service.
 func (e *RequestSendError) ConnectionError() bool {
 	return true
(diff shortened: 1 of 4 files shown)


transport/http/client.go
```diff
@@ -49,10 +50,37 @@ func (c ClientHandler) Handle(ctx context.Context, input interface{}) (
...
+// RequestSendError provides a generic request transport error.
+type RequestSendError struct {
+	Err error
+}
+
+// ConnectionError return that the error is related to not being able to send
+// the request.
+func (e *RequestSendError) ConnectionError() bool {
+	return true
+}
```
skmcgrail, line 75: Hmmm does this need to be smarter? This would cause all errors to be retried. As the retryer will retry all errors that implement ConnectionError and return true.

jasdel: I think that was the original intention. But will double check to make sure the intention here.

transport/http/client.go
```diff
@@ -49,10 +50,37 @@ func (c ClientHandler) Handle(ctx context.Context, input interface{}) (
...
+		default:
+		}
 	}
 
-	return &Response{Response: resp}, metadata, nil
+	return &Response{Response: resp}, metadata, err
+}
+
+// RequestSendError provides a generic request transport error.
+type RequestSendError struct {
```
skotambkar, line 67: Can you add more documentation as to what type of errors should/will be wrapped by the `RequestSendError`? By the name, i would think they are connection errors

skmcgrail (approved):

skotambkar (approved):

JordonPhillips (approved):

PR #190, opened 2020-09-30
codegen: Add skipping protocol tests cases that are not support

jasdel:
Adds logic that allows skipping protocol test cases that are not supported. Tests are still generated but skip statements are generated for either the operation test, or the individual test case run.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestGenerator.java
```diff
@@ -58,19 +62,22 @@
...
      *
      * @param builder the builder initializing the generator.
      */
     protected HttpProtocolUnitTestGenerator(Builder<T> builder) {
-        this.model = builder.model;
-        this.symbolProvider = builder.symbolProvider;
-        this.protocolName = builder.protocolName;
-        this.operation = builder.operation;
-        this.testCases = builder.testCases;
+        this.model = SmithyBuilder.requiredState("model", builder.model);
```
skmcgrail, line 73: I believe the first argument string is meant to match the property setter name. So instead of “symbol provider” it should be “symbolProvider”

jasdel: I seen, wasn't sure about that can fix that. I assumed it was for a message, but makes sense to match the field name directly.

skmcgrail (approved):

PR #191, opened 2020-09-30
fix to_ptr address

shogo82148:
*Issue #, if available:*

related to https://github.com/aws/aws-sdk-go-v2/issues/777
back port of https://github.com/aws/aws-sdk-go-v2/pull/778

all members of the slice point same address.

https://play.golang.org/p/FUBdYewJvLH

```go
package main

import (
	"fmt"
)

func StringSlice(vs []string) []*string {
	ps := make([]*string, len(vs))
	for i, v := range vs {
		ps[i] = &v
	}

	return ps
}

func main() {
	r := StringSlice([]string{"abc", "def"})

	fmt.Printf("%p\n", r[0])
	fmt.Printf("%p\n", r[1])
	//Output:
	//0xc0000961f0
	//0xc0000961f0
}
```

*Description of changes:*

```go
	for i, v := range vs {
		v := v // NEEDS THIS LINE
		ps[i] = &v
	}
```

### DIFF
--- a/ptr/generate.go
+++ b/ptr/generate.go
@@ -152,6 +152,7 @@ var ptrTmpl = template.Must(template.New("ptrTmpl").Parse(`
 	func {{ $.Name }}Slice(vs []{{ $.Symbol }}) []*{{ $.Symbol }} {
 		ps := make([]*{{ $.Symbol }}, len(vs))
 		for i, v := range vs {
+			v := v
 			ps[i] = &v
 		}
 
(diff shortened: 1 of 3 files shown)


jasdel (approved):
Thanks for reaching out and taking the time to fixing this bug @shogo82148. The change looks good, and we'll get this merged, and included in our next tagged release.

I see you also created https://github.com/aws/aws-sdk-go-v2/pull/778 for the v2 SDK, we'll get that merged in too.

skmcgrail (approved):

jasdel:
Made a minor change the the PR to not shadow the `v` variable, and copied the tests from https://github.com/aws/aws-sdk-go-v2/pull/778/ to here.  I'd like to see if we can change the v2 SDK's reference of these to use Go aliases instead of coping the files.

PR #194, opened 2020-09-30
Add documenation of requirement members.

jasdel:
Adds documentation to state the member is required in the generated structures.

```go
 type CreateGroupInput struct {
        // The case-sensitive name of the new group. Default is a reserved name and names
        // must be unique.
+       //
+       // This member is required.
        GroupName *string
```


skotambkar (approved):

JordonPhillips (approved):
There's a handful of other traits we'll want to give this attention too

PR #203, opened 2020-10-07
codegen: Add support for codegen adding unexported members to api structure shape

jasdel:
Adds support for generating unexported members into an API model's structure shapes. This allows unexported members to be injected into the structure shapes.

Adds utility to expect a member by name or predicate in a structure shape.


jasdel:
need to create this PR in smithy-go repo for gohash changes.

PR #204, opened 2020-10-07
add support for checksum required trait

skotambkar:
* Adds an integration to use `httpChecksumRequiredTrait` for operations that require Content-MD5 header. 

### DIFF
--- a/transport/http/checksum_middleware.go
+++ b/transport/http/checksum_middleware.go
@@ -9,24 +9,24 @@ import (
 
 const contentMD5Header = "Content-Md5"
 
-// ChecksumMiddleware provides a middleware to compute and set
-// required checksum for a http request
-type checksumMiddleware struct {
+// contentMD5ChecksumMiddleware provides a middleware to compute and set
+// content-md5 checksum for a http request
+type contentMD5ChecksumMiddleware struct {
 }
 
 // AddChecksumMiddleware adds checksum middleware to middleware's
 // build step.
 func AddChecksumMiddleware(stack *middleware.Stack) {
 	// This middleware must be executed before request body is set.
-	stack.Build.Add(&checksumMiddleware{}, middleware.Before)
+	stack.Build.Add(&contentMD5ChecksumMiddleware{}, middleware.Before)
 }
 
 // ID the identifier for the checksum middleware
-func (m *checksumMiddleware) ID() string { return "ChecksumRequiredMiddleware" }
+func (m *contentMD5ChecksumMiddleware) ID() string { return "ChecksumRequiredMiddleware" }
 
 // HandleBuild adds behavior to compute md5 checksum and add content-md5 header
 // on http request
-func (m *checksumMiddleware) HandleBuild(
+func (m *contentMD5ChecksumMiddleware) HandleBuild(
 	ctx context.Context, in middleware.BuildInput, next middleware.BuildHandler,
 ) (
 	out middleware.BuildOutput, metadata middleware.Metadata, err error,
@@ -48,11 +48,11 @@ func (m *checksumMiddleware) HandleBuild(
 			return out, metadata, fmt.Errorf("error computing md5 checksum, %w", err)
 		}
 
-		// reset the request body
+		// reset the request stream
 		req.RewindStream()
 
 		// set the 'Content-MD5' header
-		req.Header.Set("Content-MD5", string(v))
+		req.Header.Set(contentMD5Header, string(v))
 	}
 
 	// set md5 header value
--- a/transport/http/checksum_middleware.go
+++ b/transport/http/checksum_middleware.go
@@ -1,0 +1,62 @@
+package http
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	"github.com/awslabs/smithy-go/middleware"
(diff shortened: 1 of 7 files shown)


transport/http/checksum_middleware.go
```diff
@@ -0,0 +1,62 @@
...
+	body := io.TeeReader(req.Body, readBody)
+
+	// compute md5 checksum
+	v, err := computeMD5Checksum(body)
+	if err != nil {
+		return out, metadata, fmt.Errorf("error computing md5 checksum, %w", err)
+	}
+
+	// reset the request body
+	req.Body = ioutil.NopCloser(readBody)
```
jasdel, line 57: This should use the `smithyhttp.Request.SetStream` method instead of setting `Body` directly. Setting Body directly prevents correct behavior with regard to retrys (e.g. rewind).

This interface also prevents the SDK's `IsStreamRewindable` from functioning because `ioutil.NopCloser` will "hide" the fact that the underlying buffer is or is not seekable.

Since in this case we know the content will always be seekable. Probably should create a `SeekableNopCloser`, if we don't already have one, to wrap a `bytes.Reader` that is constructed instead of a byte buffer.

```go
req, err = req.SetStream(SeekableNopCloser(bytes.NewReader(readBuff.Bytes()))
in.Request = req // need to update req reference.
```

transport/http/checksum_middleware.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// set md5 header value
+	req.Header.Set("Content-MD5", string(v))
```
jasdel, line 60: Should use the const header defined above.

transport/http/checksum_middleware_test.go
```diff
@@ -0,0 +1,57 @@
+package http
```
jasdel, line 1: Would be great to have a test that ensures the `Stream` is `Rewindable`, not empty, and has original length after after the checksum is computed.

jasdel (changes requested):

transport/http/checksum_middleware.go
skotambkar: updated to use `req.Stream` directly, which returns an `io.reader`. Don't think `SeekableNopCloser` is required here, as the request's Build func handles setting request body from stream.

jasdel: `Build` will only apply the `r.stream` value to the `req.Body` if the value isn't seekable  then the request cannot be retried.

transport/http/checksum_middleware.go
```diff
@@ -0,0 +1,60 @@
...
+		if err != nil {
+			return out, metadata, fmt.Errorf("error computing md5 checksum, %w", err)
+		}
+
+		// reset the request stream
+		req.RewindStream()
+
+		// set the 'Content-MD5' header
+		req.Header.Set(contentMD5Header, string(v))
+	}
```
jasdel, line 56: This shouldn't be reaching direclty into the `req.stream`. Probably should consider that an internal detail of `smithyhttp.Request`.  Instead should use `req.GetStream()` accessor method.

Also the `req.RewindStream`'s error needs to be checked, in case the stream is not seekable.

transport/http/checksum_middleware_test.go
```diff
@@ -0,0 +1,80 @@
...
+			expectedMD5Checksum:   "1B2M2Y8AsgTpgAmY7PhCfg==",
+		},
+		"standard req body": {
+			payload: smithyio.ReadSeekNopCloser{
+				ReadSeeker: bytes.NewReader([]byte(`abc`)),
+			},
+			expectedPayloadLength: 3,
+			expectedMD5Checksum:   "kAFQmDzST7DWlj99KOF/cg==",
+		},
+		"nil body": {},
```
jasdel, line 35: Would be helpful to have a payload that is a `bytes.Buffer` without `ReadSeekNopCloser` to exercise the `RewindStream` method failing because the underlying buffer isn't seekable.

jasdel (changes requested):

transport/http/checksum_middleware_test.go
```diff
@@ -0,0 +1,80 @@
...
+			size, ok, err := req.StreamLength()
+			if err != nil {
+				t.Fatalf("error fetching request stream length")
+			}
+			if !ok {
+				t.Fatalf("request stream is not seekable")
+			}
+			if e, a := c.expectedPayloadLength, size; e != a {
+				t.Fatalf("expected request stream content length to be %v, got length %v", e, a)
+			}
```
jasdel, line 77: in addition might be helpful to also read the contents of the stream to ensure that none of the body contents where lost.

transport/http/checksum_middleware.go
```diff
@@ -0,0 +1,65 @@
...
+		v, err := computeMD5Checksum(stream)
+		if err != nil {
+			return out, metadata, fmt.Errorf("error computing md5 checksum, %w", err)
+		}
+
+		// reset the request stream
+		if err := req.RewindStream(); err != nil {
+			return out, metadata, fmt.Errorf(
+				"error rewinding request stream after computing md5 checksum")
+		}
```
jasdel, line 57: should include `%w` with `err` :)

jasdel (approved):

PR #205, opened 2020-10-08
codegen: Add support for codegen adding unexported members to api structure shape

jasdel:
Adds support for generating unexported members into an API model's structure shapes. This allows unexported members to be injected into the structure shapes.

Adds utility to expect a member by name or predicate in a structure shape.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
@@ -23,6 +23,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.function.Predicate;
 import java.util.logging.Logger;
 import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.Symbol;
@@ -368,5 +369,32 @@ public final class CodegenUtils {
             return symbolProvider.toMemberName(a)
                     .compareTo(symbolProvider.toMemberName(b));
         }
+    }
+
+    /**
+     * Attempts to find the first member by exact name in the containing structure. If the member is not found an
+     * exception will be thrown.
+     *
+     * @param shape structure containing member
+     * @param name  member name
+     * @return MemberShape if found
+     */
+    public static MemberShape expectMember(StructureShape shape, String name) {
+        return expectMember(shape, name::equals);
+    }
+
+    /**
+     * Attempts to find the first member by name using a member name predicate in the containing structure. If the
+     * member is not found an exception will be thrown.
+     *
+     * @param shape               structure containing member
+     * @param memberNamePredicate member name to search for
+     * @return MemberShape if found
+     */
+    public static MemberShape expectMember(StructureShape shape, Predicate<String> memberNamePredicate) {
+        return shape.getAllMembers().values().stream()
+                .filter((p) -> memberNamePredicate.test(p.getMemberName()))
+                .findFirst()
+                .orElseThrow(() -> new CodegenException("did not find member in structure shape, " + shape.getId()));
     }
 }
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
@@ -25,6 +25,7 @@ import software.amazon.smithy.codegen.core.ReservedWords;
 import software.amazon.smithy.codegen.core.ReservedWordsBuilder;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
+import software.amazon.smithy.go.codegen.trait.UnexportedMemberTrait;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
 import software.amazon.smithy.model.neighbor.NeighborProvider;
@@ -222,7 +223,11 @@ final class SymbolVisitor implements SymbolProvider, ShapeVisitor<Symbol> {
     }
 
     private String getDefaultMemberName(MemberShape shape) {
-        return StringUtils.capitalize(removeLeadingUnderscores(shape.getMemberName()));
+        String memberName =  StringUtils.capitalize(removeLeadingUnderscores(shape.getMemberName()));
+        if (shape.hasTrait(UnexportedMemberTrait.class)) {
+            memberName = Character.toLowerCase(memberName.charAt(0)) + memberName.substring(1);
+        }
+        return memberName;
     }
 
     private String removeLeadingUnderscores(String value) {
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
@@ -195,11 +195,13 @@ final class SymbolVisitor implements SymbolProvider, ShapeVisitor<Symbol> {
 
     @Override
     public String toMemberName(MemberShape shape) {
-        String memberName = getDefaultMemberName(shape);
         Shape container = model.expectShape(shape.getContainer());
         if (container.isUnionShape()) {
-            memberName = formatUnionMemberName(container.asUnionShape().get(), shape);
+            // Union member names are not escaped as they are used to build the escape set.
+            return formatUnionMemberName(container.asUnionShape().get(), shape);
         }
+
+        String memberName = getDefaultMemberName(shape);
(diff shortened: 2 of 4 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
```diff
@@ -369,4 +370,46 @@ public int compare(MemberShape a, MemberShape b) {
...
+     */
+    public static MemberShape expectMemberByPredicate(
+            StructureShape shape,
+            Predicate<MemberShape> memberShapePredicate
+    ) {
+        return shape.getAllMembers().values().stream()
+                .filter(memberShapePredicate)
+                .findFirst()
+                .orElseThrow(() -> new CodegenException("did not find member in structure shape, " + shape.getId()));
+    }
```
skotambkar, line 414: These can be merged.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -222,7 +223,11 @@ private String getDefaultShapeName(Shape shape) {
     }
 
     private String getDefaultMemberName(MemberShape shape) {
-        return StringUtils.capitalize(removeLeadingUnderscores(shape.getMemberName()));
+        String memberName =  StringUtils.capitalize(removeLeadingUnderscores(shape.getMemberName()));
+        if (shape.hasTrait(UnexportedMemberTrait.class)) {
+            memberName = Character.toLowerCase(memberName.charAt(0)) + memberName.substring(1);
+        }
```
skotambkar, line 229: Just to verify, do we have validations for member shape name to not have numeric's in name? 

example case : `__123Alpha`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/trait/UnexportedMemberTrait.java
```diff
@@ -0,0 +1,46 @@
...
+package software.amazon.smithy.go.codegen.trait;
+
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.AnnotationTrait;
+
+/**
+ * Provides a trait to decorate an member as unexported, so it will not be exported within the containing structure.
+ * This should only be used on top level input/output structure shapes, because internal members of non-input/output
```
skotambkar, line 27: Would this trait be applicable on exception shapes?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/trait/UnexportedMemberTrait.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public UnexportedMemberTrait() {
+        this(Node.objectNode());
+    }
+
+    public UnexportedMemberTrait(ObjectNode node) {
+        super(ID, node);
+    }
+
+    public static final class Provider extends AnnotationTrait.Provider<UnexportedMemberTrait> {
```
skotambkar, line 41: This trait provider might need to be registered in `META-INF` somewhere, for it to be usable. 

Smithy does it: 
https://github.com/awslabs/smithy/blob/34d51e3f7f9c351ed0867a816e082ba3ac56cc4c/smithy-aws-traits/src/main/resources/META-INF/services/software.amazon.smithy.model.traits.TraitService

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/trait/UnexportedMemberTrait.java
```diff
@@ -0,0 +1,46 @@
...
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.AnnotationTrait;
+
+/**
+ * Provides a trait to decorate an member as unexported, so it will not be exported within the containing structure.
+ * This should only be used on top level input/output structure shapes, because internal members of non-input/output
+ * shapes are not visible outside of the API client's "types" package.
+ */
+public class UnexportedMemberTrait extends AnnotationTrait {
```
skotambkar, line 30: This trait is simple enough so may not require it. But all trait should have a `Trait test` associated with them.

transport/http/client.go
```diff
@@ -121,3 +121,16 @@ func WrapLogClient(logger interface{ Logf(string, ...interface{}) }, client Clie
 		return resp, nil
 	})
 }
+
+// NopClient provides a client that ignores the request, and returns a empty
+// successful HTTP response value.
+type NopClient struct{}
```
skotambkar, line 127: 1. Is this required for testing? or why should this be in `transport/http/client.go`?

2. If for testing, should this be modified to:

```
type NopClient struct{
	StatusCode int
	Header []http.Header
	Body   io.ReadCloser
}

```

That way we can create clients which can return error case and stuff. 

The Do implementation that you have below, can be used for the default returned values in case struct member isn't populated.

skotambkar (commented):
Few comments/questions/suggestions.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
jasdel: Sure, i'd initially had them separate so that a search member by predicate could be available, but it can always be added in later if needed.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
jasdel: Most likely not, protocol tests should be created for these, though the test probably have to be specific to the SDK.

It looks like this needs a little rework to make sure the lower casing happens after numbers have been stripped. off.  I'm guessing Smtihy allows member/shapes named `__123abc` :(

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/trait/UnexportedMemberTrait.java
jasdel: Yeah, any structure shape, which exception shapes are.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/trait/UnexportedMemberTrait.java
jasdel: interesting. I'm able to use it without issue. will look into the purpose that that provider.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/trait/UnexportedMemberTrait.java
jasdel: Yeah that makes sense, I did not get to tests for this. Also this trait needs validation that should it only can be applied to members of a structure.

transport/http/client.go
jasdel: This is required for the presign URL behavior, not tests.  The purpose of it is to have a no-op HTTP client that will do nothing but return success. Basically to prevent a middleware stack making a request with a presigned URL flow.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/trait/UnexportedMemberTrait.java
jasdel: Not sure the purpose of that metadata file, but it doesn't include all traits, e.g. `OptionalAuthTrait`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/trait/UnexportedMemberTrait.java
jasdel: hmm, looks like the annotation traits don't have tests in smithy.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
jasdel: yeah according to smithy's docs `__123abc` is a valid shape id :( https://awslabs.github.io/smithy/1.0/spec/core/model.html?highlight=membername#shape-id

jasdel: Ended up refactoring several places that where symbols weren't being used, and added case to the smithy-go tests to ensure the sdk handles `__123abc` correctly.

PR #219, opened 2020-10-16
Add Support for Slots to Stack Steps

skmcgrail:
Adds support for pre-configuring slot identifiers that middleware can use for inserting / adding themselves to a step.

### DIFF
--- a/middleware/ordered_group.go
+++ b/middleware/ordered_group.go
@@ -37,7 +37,7 @@ func (g *orderedIDs) IsSlot(id string) bool {
 }
 
 // Add injects the item to the relative position of the item group. Returns an
-// error if the item already exists.
+// error if the item already exists. RelativePosition is ignored if the ider fills a naed slot.
 func (g *orderedIDs) Add(m ider, pos RelativePosition) error {
 	id := m.ID()
 	if len(id) == 0 {
@@ -48,6 +48,8 @@ func (g *orderedIDs) Add(m ider, pos RelativePosition) error {
 		if err := g.order.Add(id, pos); err != nil {
 			return err
 		}
+	} else if _, ok := g.items[id]; ok {
+		return fmt.Errorf("already exists, %v", id)
 	}
 
 	g.items[id] = m
@@ -112,7 +114,7 @@ func (g *orderedIDs) Get(id string) (ider, bool) {
 }
 
 // Swap removes the item by id, replacing it with the new item. Returns error
-// if the original item doesn't exist.
+// if the original item doesn't exist, or if the ider does not match the slot name being swapped.
 func (g *orderedIDs) Swap(id string, m ider) (ider, error) {
 	if len(id) == 0 {
 		return nil, fmt.Errorf("swap from ID must not be empty")
--- a/middleware/ordered_group.go
+++ b/middleware/ordered_group.go
@@ -143,12 +143,13 @@ func (g *orderedIDs) Swap(id string, m ider) (ider, error) {
 		return nil, fmt.Errorf("swap to ID must not be empty")
 	}
 
-	if g.IsSlot(id) && id != iderID {
+	isSlot := g.IsSlot(id)
+	if isSlot && id != iderID {
 		return nil, fmt.Errorf("swap to ID must match slot being swapped")
-	}
-
-	if err := g.order.Swap(id, iderID); err != nil {
-		return nil, err
+	} else if !isSlot {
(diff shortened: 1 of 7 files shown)


middleware/ordered_group.go
```diff
@@ -21,28 +21,51 @@ type ider interface {
...
 
-	if err := g.order.Add(m.ID(), pos); err != nil {
+	if !g.IsSlot(id) {
+		if err := g.order.Add(id, pos); err != nil {
+			return err
+		}
+	}
+
+	g.items[id] = m
+	return nil
```
jasdel, line 54: Should this check to see if a middleware already exists at the slot? It looks like this logic will allow a middleware to overwrite another middleware as long as it targets a slot. I'd expect this to return an error with the an `id` value already exists in `g.items`.

middleware/ordered_group.go
```diff
@@ -76,18 +117,24 @@ func (g *orderedIDs) Swap(id string, m ider) (ider, error) {
...
+
+	iderID := m.ID()
+	if len(iderID) == 0 {
 		return nil, fmt.Errorf("swap to ID must not be empty")
 	}
 
-	if err := g.order.Swap(id, m.ID()); err != nil {
+	if g.IsSlot(id) && id != iderID {
+		return nil, fmt.Errorf("swap to ID must match slot being swapped")
+	}
```
jasdel, line 128: Would be good to update the doc for this function calling out the new error case.

middleware/ordered_group_test.go
```diff
@@ -190,3 +190,39 @@ func TestOrderedIDsGetOrder(t *testing.T) {
...
+	noError(t, o.AddSlot("second", After))
+	noError(t, o.InsertSlot("fourth", "second", After))
+	noError(t, o.Insert(mockIder("third"), "second", After))
+
+	expectIDs := []string{"first", "second", "third", "fourth"}
+	if e, a := expectIDs, o.List(); !reflect.DeepEqual(e, a) {
+		t.Errorf("expect %v order, got %v", e, a)
+	}
+
+	noError(t, o.Add(mockIder("second"), After))
```
jasdel, line 207: It would be good to have a test case that covers adding a `ider` with the same `id` multiple times, with subsequent calls failing.

middleware/ordered_group_test.go
```diff
@@ -190,3 +190,39 @@ func TestOrderedIDsGetOrder(t *testing.T) {
...
 	}
 }
+
+func TestOrderedIDsSlots(t *testing.T) {
+	o := newOrderedIDs()
+
+	noError(t, o.AddSlot("first", After))
+	noError(t, o.AddSlot("second", After))
+	noError(t, o.InsertSlot("fourth", "second", After))
+	noError(t, o.Insert(mockIder("third"), "second", After))
```
jasdel, line 200: Would be nice to have a swap positive and negative test case too.  Using test coverage with these are a great way to help more cases get hit.

jasdel (approved):

middleware/ordered_group_test.go
```diff
@@ -181,18 +181,17 @@ func TestOrderedIDsGetOrder(t *testing.T) {
...
 	noError(t, o.Add(mockIder("fifth"), After))
 
 	if e, a := []string{"first", "second", "third", "fourth", "fifth"}, o.List(); !reflect.DeepEqual(e, a) {
 		t.Errorf("expect %v order, got %v", e, a)
 	}
 
-	if o.AddSlot("first", After) == nil {
+	if o.AddSlot(After, "first") == nil {
 		t.Error("expect error, got nil")
 	}
```
jasdel: Change looks good, but need test cases to ensure correct behavior if the `ids...` variadic list is empty.

middleware/ordered_group.go
```diff
@@ -260,18 +282,26 @@ func (s *relativeOrder) Clear() {
...
+		}
+		copy(s.order[i+n:], src[i:])
+		copy(s.order[i:], ids)
 	case After:
 		if i == len(s.order)-1 {
-			s.order = append(s.order, id)
+			s.order = append(s.order, ids...)
 		} else {
-			s.order = append(s.order[:i+1], append([]string{id}, s.order[i+1:]...)...)
+			s.order = append(s.order[:i+1], append(ids, s.order[i+1:]...)...)
```
jasdel, line 304: Would be good to have a test that asserts this does the correct behavior if `ids...` is an empty slice.

PR #220, opened 2020-10-19
Add Values method to enum types

JordonPhillips:
This adds a function to generated enums that lists all the possible
values.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
```diff
@@ -57,5 +58,15 @@ public void run() {
                 writer.write("$L $L = $S", label, symbol.getName(), definition.getValue());
             }
         }).write("");
+
+        writer.writeDocs(String.format("Values returns all known values for %s. Note that this can be expanded in the "
+                + "future, and so it is only as up to date as the client.", symbol.getName()));
```
jasdel, line 63: This probably should also call out that the ordering of the enums in the returned slice may change, and not to rely on any particular ordering?

jasdel (approved):

PR #221, opened 2020-10-20
Add ability to define middleware step slots via the RuntimeClientPlugin 

skmcgrail:
Adds the ability to register runtime client plugins that provide registration of middleware step slots.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
@@ -19,11 +19,13 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Collectors;
+import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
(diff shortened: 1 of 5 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -223,6 +226,29 @@ private void generateClientInvokeOperation() {
...
+
+        for (RuntimeClientPlugin runtimeClientPlugin : runtimePlugins) {
+            if (!runtimeClientPlugin.matchesService(model, service)
+                    || !runtimeClientPlugin.getResolveFunction().isPresent()) {
+                Optional<StackSlotRegistrar> registrar = runtimeClientPlugin.getRegisterStackSlots();
+                if (!registrar.isPresent()) {
+                    continue;
+                }
+                builder = builder.merge(registrar.get());
+            }
```
jasdel, line 246: The `StackSlotRegistrar` builder has a both add and remove slot behavior. How should that get reconciled with the merging, or would it?

skmcgrail: That was just put there to match the existing builder pattern, but I could see that as being potentially confusing as the removal isn't a process directive, but only removes the entry from the given step if present. Likely we should just remove it from the build. The idea is that there are few plugins that register slots, and likely we shouldn't have a plugin that removes another slot that a plugin might rely on.

PR #222, opened 2020-10-23
Add HostnameImmutable context value for middleware

jasdel:
Adds a new `HostnameImmutable` context value getter/setter helpers to `transport/http` package. 

Updates EndpointTrait's HostPrefix handling to use the HostnameImmutable to skip modifying the endpoint.

Also refactors EndpointTrait HostPrefix handling to be an integration instead of implemented directly in the operation serializer.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
@@ -1,0 +1,194 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ *
+ *
+ */
+
+package software.amazon.smithy.go.codegen.integration;
+
+import java.util.ArrayList;
+import java.util.List;
+import software.amazon.smithy.codegen.core.SymbolProvider;
+import software.amazon.smithy.go.codegen.GoDelegator;
+import software.amazon.smithy.go.codegen.GoSettings;
+import software.amazon.smithy.go.codegen.GoStackStepMiddlewareGenerator;
+import software.amazon.smithy.go.codegen.GoWriter;
+import software.amazon.smithy.go.codegen.SmithyGoDependency;
+import software.amazon.smithy.go.codegen.SymbolUtils;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.pattern.SmithyPattern;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.traits.EndpointTrait;
+
+public class EndpointHostPrefixMiddleware implements GoIntegration {
+
+    private static final String MIDDLEWARE_ID = "EndpointHostPrefix";
+
+    List<RuntimeClientPlugin> runtimeClientPlugins = new ArrayList<>();
+    List<OperationShape> endpointPrefixOperations = new ArrayList<>();
+
+    @Override
+    public void processFinalizedModel(GoSettings settings, Model model) {
+        ServiceShape service = settings.getService(model);
+        endpointPrefixOperations = getOperationsWithEndpointPrefix(model, service);
+
+        endpointPrefixOperations.forEach((operation) -> {
+            String middlewareHelperName = getMiddlewareHelperName(operation);
+            runtimeClientPlugins.add(RuntimeClientPlugin.builder()
+                    .operationPredicate((m, s, o) -> o.equals(operation))
+                    .registerMiddleware(MiddlewareRegistrar.builder()
+                            .resolvedFunction(SymbolUtils.createValueSymbolBuilder(middlewareHelperName).build())
+                            .build())
+                    .build()
+            );
+        });
+    }
+
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins() {
+        return runtimeClientPlugins;
+    }
+
+    @Override
+    public void writeAdditionalFiles(
+            GoSettings settings,
+            Model model,
+            SymbolProvider symbolProvider,
+            GoDelegator delegator
+    ) {
+        endpointPrefixOperations.forEach((operation) -> {
+            delegator.useShapeWriter(operation, (writer) -> {
+                SmithyPattern pattern = operation.expectTrait(EndpointTrait.class).getHostPrefix();
+
+                writeMiddleware(writer, model, symbolProvider, operation, pattern);
+
+                String middlewareName = getMiddlewareName(operation);
+                String middlewareHelperName = getMiddlewareHelperName(operation);
+                writer.addUseImports(SmithyGoDependency.SMITHY_MIDDLEWARE);
+                writer.openBlock("func $L(stack *middleware.Stack) error {", "}",
+                        middlewareHelperName,
+                        () -> {
+                            writer.write(
+                                    "return stack.Serialize.Insert(&$L{}, `OperationSerializer`, middleware.Before)",
+                                    middlewareName);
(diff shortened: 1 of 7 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
```diff
@@ -0,0 +1,204 @@
...
+    public List<RuntimeClientPlugin> getClientPlugins() {
+        return runtimeClientPlugins;
+    }
+
+    /**
+     * Builds the set of runtime plugs used by the endpoint prefix middleware.
+     *
+     * @param settings codegen settings
+     * @param model    api model
+     */
```
skotambkar, line 53: Nit: Don't need these comments if overriding a method

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
```diff
@@ -0,0 +1,204 @@
...
+
+        operations.forEach((operation) -> {
+            delegator.useShapeWriter(operation, (writer) -> {
+                SmithyPattern pattern = operation.expectTrait(EndpointTrait.class).getHostPrefix();
+
+                writeMiddleware(writer, model, symbolProvider, operation, pattern);
+
+                String middlewareName = getMiddlewareName(operation);
+                String middlewareHelperName = getMiddlewareHelperName(operation);
+                writer.openBlock("func $L(stack *middleware.Stack) error {", "}",
```
skotambkar, line 97: Nit: missing `addUseImports` statement for middleware import

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * Gets a list of the operations decorated with the EndpointTrait.
+     *
+     * @param model   Model used for generation.
+     * @param service Service for getting list of operations.
+     * @return list of operations decorated with the EndpointTrait.
+     */
+    public static List<OperationShape> getOperationsWithEndpointPrefix(Model model, ServiceShape service) {
```
skotambkar, line 188: This can be `private` method

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
```diff
@@ -0,0 +1,204 @@
...
+        service.getAllOperations().stream().forEach((operationId) -> {
+            OperationShape operation = model.expectShape(operationId).asOperationShape().get();
+            if (!operation.hasTrait(EndpointTrait.ID)) {
+                return;
+            }
+
+            operations.add(operation);
+        });
+        return operations;
+    }
```
skotambkar, line 199: Instead of `getOperationsWithEndpointPrefix` method, returning a list of operations, suggest changing this method to `isOperationWithEndpointPrefix` which takes in an operation and return a boolean indicating if an operation has endpoint trait. 

Right now you call this method twice, you iterate all operations twice and then again iterate on returned operations list.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
```diff
@@ -0,0 +1,204 @@
...
+                    if (!segment.isLabel()) {
+                        w.write("prefix.WriteString($S)", segment.toString());
+                    } else {
+                        MemberShape member = input.getMember(segment.getContent()).get();
+                        String memberName = symbolProvider.toMemberName(member);
+                        String memberReference = "input." + memberName;
+
+                        // Theoretically this should never be nil or empty by this point unless validation has
+                        // been disabled.
+                        w.write("if $L == nil || len(*$L) == 0 {", memberReference, memberReference).indent();
```
skotambkar, line 158: QQ: Are we certain host prefix would not be numeric?

transport/http/middleware_metadata.go
```diff
@@ -0,0 +1,20 @@
+package http
+
+import "context"
+
+type (
+	hostnameImmutableKey struct{}
+)
+
+// GetHostnameImmutable retrieves the if the endpoint hostname should be
+// considered immutable or not.
```
skotambkar, line 10: Nit: extra `the` in comment

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
```diff
@@ -0,0 +1,209 @@
...
+    /**
+     * Builds the set of runtime plugs used by the endpoint prefix middleware.
+     *
+     * @param settings codegen settings
+     * @param model    api model
+     */
+    @Override
+    public void processFinalizedModel(GoSettings settings, Model model) {
+        ServiceShape service = settings.getService(model);
+        List<OperationShape> operations = getOperationsWithEndpointPrefix(model, service);
```
skotambkar, line 57: You should be able to store this operation list and reuse it in `writeAdditionalFiles` method

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
skotambkar: Posted another alternative above, where you can reuse the previously computed operation list

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
jasdel: This was made public so that tests could be generated in the SDK's codegen module.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
jasdel: Correct, host prefix labels must only be string values. But I've also removed this `len` check as its duplicates the validate method in the if else case.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
jasdel: The intention of this helper is a quick get list of operations that are match the filter and iterate over the result. Similar to how the knowledge indexes, and other collectors behave. This also would apply to other modules like the SDK's codegen that may use this helper to iterate through the operations with this trait in order to generate unit tests for operations using this trait.

Will investigate the is filter vs the collector pattern

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
```diff
@@ -0,0 +1,194 @@
...
+
+    @Override
+    public void processFinalizedModel(GoSettings settings, Model model) {
+        ServiceShape service = settings.getService(model);
+        endpointPrefixOperations = getOperationsWithEndpointPrefix(model, service);
+
+        endpointPrefixOperations.forEach((operation) -> {
+            String middlewareHelperName = getMiddlewareHelperName(operation);
+            runtimeClientPlugins.add(RuntimeClientPlugin.builder()
+                    .operationPredicate((m, s, o) -> o.equals(operation))
```
skotambkar, line 52: Missed this before, won't this operationPredicate always be true?

skotambkar (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
skotambkar: If I am looking at this correctly, may be having `endpointPrefixOperations` as a Set, rather than a List would be better here.  Then you don't have to iterate over the list.

jasdel: Spoke about this out of band. The individual predicates are needed since each operation has its own middleware and middleware helper.  If they all shared the same middleware and helper doing single runtime client plugin would be usable instead of multiple.

skmcgrail (approved):

PR #227, opened 2020-10-27
Camel-case using underscore separators

JordonPhillips:
The enum generator had been splitting and camel-casing based on
non-word characters, but that leaves underscores. This updates the
regex also split on underscores.

If the camel-casing were to produce a conflict, this will keep the
first camel-casing but leave subsequent potential conflicts
untouched.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
@@ -15,7 +15,10 @@
 
 package software.amazon.smithy.go.codegen;
 
+import java.util.LinkedHashSet;
 import java.util.Locale;
+import java.util.Set;
+import java.util.logging.Logger;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.model.shapes.StringShape;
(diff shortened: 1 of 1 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/EnumGenerator.java
```diff
@@ -46,14 +50,25 @@ public void run() {
...
                 String label = labelBuilder.toString();
+
+                // If camel-casing would cause a conflict, don't camel-case this enum value.
+                if (constants.contains(label)) {
+                    LOGGER.warning(String.format(
+                            "Multiple enums resolved to the same name, `%s`, using unaltered value for: %s",
+                            label, name));
+                    label = name;
+                }
+                constants.add(label);
```
jasdel, line 70: Is there a risk of a breaking change being introduced into the generated output when new enum values are added? For example an enum contains `Foo_123` and `Bar_123` names on day 1, then a new name, `Bar123` is added on day 2. Are there any protections that prevent `Bar_123` not being escaped on day 2?

JordonPhillips: In this example `Bar_123` would still be escaped if it is first in the list. So you'd end up with code gen ultimately failing due to a conflict. Alternatively we could just... keep appending underscores until it doesn't conflict anymore. There's still a risk that inserting into the middle would cause a breaking change though

jasdel: Using something like golang.org/x/exp/cmd/gorelease to check for breaking changes will help prevent releasing unexpected breaking changes. 

This is probably also an area that we should have smithy diff checking in the SDK that would signal us of this issue early on. Potentially suffixing with `_` is the alternative solution, but suffixing only really works if the order is stable between updates. e.g. no inserts in middle of list.

jasdel (approved):

PR #229, opened 2020-10-28
Fix list formatting

JordonPhillips:
This removes the indentation of lists, as indenting would only cause the line to be treated as a code block in godoc.


skotambkar:
Does this PR fix the issue seen at https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2 where there seems to be a newline after the bullet point. 

Indenting at the start of list does put it in a code block, which might be the issue.  Were you able to render the docs locally to verify the behavior?

JordonPhillips:
Screen Shot 2020-10-28 at 6 31 41 PM

JordonPhillips:
The above screenshot was taken after re-running generation with this commit, using the godoc tool

skotambkar (approved):

PR #233, opened 2020-11-01
moves host prefix addition to a build step middleware

skotambkar:
*Issue #, if available:*

Endpoint host prefix addition was being performed after a request was signed. This led to a incorrectly signed request. 

*Description of changes:*

[x] This changes moves the endpoint mutation to a build step middleware, that will be performed before signing.  

[x] We also add a `DisableEndpointHostPrefix` and `IsEndpointHostPrefixDisabled` method on transport/http which provides an ability to turn host prefixing off. This would be useful when applying customizations or for testing purposes.

Related to https://github.com/aws/aws-sdk-go-v2/pull/862 and https://github.com/aws/aws-sdk-go-v2/pull/863

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
@@ -35,6 +35,10 @@ import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.EndpointTrait;
 
+/**
+ * EndpointHostPrefixMiddleware adds middlewares to identify
+ * a host prefix and mutate the request URL host if permitted.
+**/
 public class EndpointHostPrefixMiddleware implements GoIntegration {
 
     private static final MiddlewareIdentifier MIDDLEWARE_ID = MiddlewareIdentifier.string("EndpointHostPrefix");
@@ -84,10 +88,9 @@ public class EndpointHostPrefixMiddleware implements GoIntegration {
                         middlewareHelperName,
                         () -> {
                             writer.write(
-                                    "return stack.Serialize.Insert(&$L{}, `OperationSerializer`, middleware.Before)",
+                                    "return stack.Serialize.Insert(&$L{}, `OperationSerializer`, middleware.After)",
                                     middlewareName);
                         });
-
             });
         });
     }
@@ -109,7 +112,8 @@ public class EndpointHostPrefixMiddleware implements GoIntegration {
             writer.addUseImports(SmithyGoDependency.SMITHY_HTTP_TRANSPORT);
             writer.addUseImports(SmithyGoDependency.FMT);
 
-            w.openBlock("if smithyhttp.GetHostnameImmutable(ctx) {", "}", () -> {
+            w.openBlock("if smithyhttp.GetHostnameImmutable(ctx) || "
+                    + "smithyhttp.IsEndpointHostPrefixDisabled(ctx) {", "}", () -> {
                 w.write("return next.$L(ctx, in)", generator.getHandleMethodName());
             }).write("");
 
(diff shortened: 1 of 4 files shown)


transport/http/middleware_host_prefix.go
```diff
@@ -0,0 +1,35 @@
...
+) {
+	req, ok := in.Request.(*Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown request type %T", req)
+	}
+
+	// Add the host prefix
+	if len(req.HostPrefix) != 0 && !IsEndpointHostPrefixDisabled(ctx) {
+		req.URL.Host = req.HostPrefix + req.URL.Host
+	}
```
jasdel, line 32: I think one thing we need to consider with this change, is if `HostPrefix` should be a member on the `smithyhttp.Request` at all. I'd venture to say it should not, since the correct endpoint cannot be determined without calling `smithyhttp.Requerst.Build`

jasdel: Can this skip the additional `Build` middleware and set `HostPrefix` in the serialize middleware since the new `IsEndpointHostPrefixDisabled` flag to disable behavior. It seems the additional Build middleware isn't really needed.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
```diff
@@ -108,10 +123,13 @@ private static void writeMiddleware(
             writer.addUseImports(SmithyGoDependency.SMITHY_HTTP_TRANSPORT);
             writer.addUseImports(SmithyGoDependency.FMT);
 
-            w.openBlock("if smithyhttp.GetHostnameImmutable(ctx) {", "}", () -> {
+            w.openBlock("if smithyhttp.GetHostnameImmutable(ctx) || "
+                    + "smithyhttp.IsEndpointHostPrefixDisabled(ctx) {", "}", () -> {
```
jasdel, line 127: Can this middleware set set the prefix on the endpoint url directly without the need for the `HostPrefix` member on Request?

skotambkar: Approached this issue to unblock on other PRs, but didn't spend time to investigate why this was designed the way it was, i.e. setting on request `hostprefix` and then performing mutation at the end.   Didn't change the behavior of setting `hostprefix` on smithy-http request .

transport/http/middleware_host_prefix.go
skotambkar: Keeping it in serialize didn't seem a good choice as we lack middleware slots atm. 

Serialize steps are getting easily crowded and maintaining order is difficult. 

Keeping it in Build ensure this is ran after everything is serialized and customized. I decided to add it as one of the first middlewares in build step. I think we need a better story on what goes in serialize steps

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
jasdel: It would be great if we can take a look at removing HostPrefix before adding additional complexity of additional middleware.

transport/http/middleware_host_prefix.go
jasdel: Understand the crowdyness of Serialize. I think thats the suggestion to remove HostPrefix all together.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
jasdel: Maybe @JordonPhillips  can provide more background why HostPrefix was used vs editing the endpoint host directly.

jasdel (approved):

PR #234, opened 2020-11-09
PaginationTrait Support

skmcgrail:
Depends on https://github.com/awslabs/smithy/pull/628

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
@@ -17,10 +17,12 @@ package software.amazon.smithy.go.codegen.integration;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Consumer;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.go.codegen.GoDelegator;
 import software.amazon.smithy.go.codegen.GoSettings;
+import software.amazon.smithy.go.codegen.GoValueAccessUtils;
 import software.amazon.smithy.go.codegen.GoWriter;
 import software.amazon.smithy.go.codegen.SmithyGoDependency;
 import software.amazon.smithy.go.codegen.SymbolUtils;
@@ -158,36 +160,28 @@ public class Paginators implements GoIntegration {
                         writer.write("return nil, err");
                     });
                     writer.write("p.firstPage = false");
-
-                    StringBuilder nilGuard = new StringBuilder();
-                    StringBuilder outputPath = new StringBuilder("result");
+                    writer.write("");
 
                     List<MemberShape> outputMemberPath = paginationInfo.getOutputTokenMemberPath();
-                    for (int i = 0; i < outputMemberPath.size(); i++) {
-                        MemberShape memberShape = outputMemberPath.get(i);
-                        outputPath.append(".");
-                        outputPath.append(symbolProvider.toMemberName(memberShape));
+                    MemberShape tokenMember = outputMemberPath.get(outputMemberPath.size() - 1);
+                    Consumer<String> setToken = (container) -> {
+                        writer.write("p.nextToken = $L", container + "."
+                                + symbolProvider.toMemberName(tokenMember));
+                    };
 
-                        // Don't need to check the last member here as we won't dereference it
-                        if (i != outputMemberPath.size() - 1) {
-                            if (i != 0) {
-                                nilGuard.append(" && ");
-                            }
-                            nilGuard.append(outputPath);
-                            nilGuard.append(" != nil");
-                        }
+                    for (int i = outputMemberPath.size() - 2; i >= 0; i--) {
+                        MemberShape memberShape = outputMemberPath.get(i);
+                        Consumer<String> inner = setToken;
+                        setToken = (container) -> {
+                            GoValueAccessUtils.writeIfNonZeroValueMember(model, symbolProvider, writer, memberShape,
+                                    container, inner);
+                        };
                     }
 
                     writer.write("prevToken := p.nextToken");
-                    Runnable setToken = () -> {
-                        writer.write("p.nextToken = $L", outputPath);
-                    };
-                    if (nilGuard.length() > 0) {
-                        writer.write("p.nextToken = nil");
-                        writer.openBlock("if $L {", "}", nilGuard.toString(), setToken::run);
-                    } else {
-                        setToken.run();
-                    }
+                    writer.write("p.nextToken = nil");
+                    setToken.accept("result");
+                    writer.write("");
 
                     if (model.expectShape(paginationInfo.getInputTokenMember().getTarget()).isStringShape()) {
                         writer.openBlock("if p.options.StopOnDuplicateToken && "
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
@@ -20,12 +20,14 @@ import java.util.Optional;
 import java.util.function.Consumer;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
+import software.amazon.smithy.go.codegen.CodegenUtils;
 import software.amazon.smithy.go.codegen.GoDelegator;
 import software.amazon.smithy.go.codegen.GoSettings;
 import software.amazon.smithy.go.codegen.GoValueAccessUtils;
 import software.amazon.smithy.go.codegen.GoWriter;
 import software.amazon.smithy.go.codegen.SmithyGoDependency;
 import software.amazon.smithy.go.codegen.SymbolUtils;
+import software.amazon.smithy.go.codegen.knowledge.GoPointableIndex;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.PaginatedIndex;
 import software.amazon.smithy.model.knowledge.PaginationInfo;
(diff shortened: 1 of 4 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
```diff
@@ -0,0 +1,254 @@
...
+
+        if (paginationInfos.size() == 0) {
+            return;
+        }
+
+        writerFactory.accept("paginators.go", settings.getModuleName(), writer -> {
+            paginationInfos.forEach((paginationInfo) -> {
+                generateOperationPaginator(model, symbolProvider, writer, paginationInfo);
+            });
+        });
```
jasdel, line 68: Thoughts on using paginators file vs operation file? Since they are per operation.  Using the operation shape is an easy way to get a writer for shape.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
```diff
@@ -0,0 +1,254 @@
...
+            Symbol optionsSymbol
+    ) {
+        String inputMember = symbolProvider.toMemberName(paginationInfo.getInputTokenMember());
+
+        writer.writeDocs(String.format("%s is a paginator for %s", paginatorSymbol, operationSymbol));
+        writer.openBlock("type $T struct {", "}", paginatorSymbol, () -> {
+            writer.write("options $T", optionsSymbol);
+            writer.write("client $T", interfaceSymbol);
+            writer.write("params $P", inputSymbol);
+            writer.write("nextToken *string");
```
jasdel, line 114: type of `nextToken` probably should be $P of symbol of next token member. Especially to support things like DDB's next token.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
```diff
@@ -0,0 +1,254 @@
...
+                        writer.write("params: params,");
+                        writer.write("firstPage: true,");
+                    });
+                });
+        writer.write("");
+
+        writer.writeDocs("HasMorePages returns a boolean indicating whether more pages are available");
+        writer.openBlock("func (p $P) HasMorePages() bool {", "}", paginatorSymbol, () -> {
+            writer.openBlock("if p.done {", "}", () -> writer.write("return false"));
+            writer.write("return p.firstPage || (p.nextToken != nil && len(*p.nextToken) > 0)");
```
jasdel, line 142: This won't support the DDB paginators.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
```diff
@@ -0,0 +1,254 @@
...
+                    if (nilGuard.length() > 0) {
+                        writer.openBlock("if $L {", "}", nilGuard.toString(), setToken::run);
+                    } else {
+                        setToken.run();
+                    }
+
+                    writer.openBlock("if p.options.StopOnDuplicateToken && "
+                            + "params.$L != nil && p.nextToken != nil && "
+                            + "*params.$L == *p.nextToken {", "", inputMember, inputMember, () -> {
+                        writer.write("p.done = true");
```
jasdel, line 199: Instead of an extra `done` member could just set `p.nextToken` to nil.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
```diff
@@ -0,0 +1,254 @@
...
+                        writer.openBlock("} else {", "}", () -> writer.write("p.done = false"));
```
jasdel, line 200: If a paginator stops because of duplicate tokens and user asked it to stop, this should be considered terminal stop.  Probably no need to allow resume. Doing so seems to be a recipe for unexpected behavior and errors.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
```diff
@@ -0,0 +1,254 @@
...
+                MemberShape limitShape = pageSizeMember.get();
+                limitShape.getMemberTrait(model, DocumentationTrait.class).ifPresent(documentationTrait -> {
+                    writer.writeDocs(documentationTrait.getValue());
+                });
+                writer.write("Limit $P", symbolProvider.toSymbol(limitShape));
+                writer.write("");
+            });
+            writer.writeDocs("Set to true if pagination should stop if the service returns "
+                    + "the same taken passed to it");
+            writer.write("StopOnDuplicateToken bool");
```
jasdel, line 229: Minor nit, with doc.  Maybe something to clarify that the same token returned for subiquent requests. Basically to clarify that the sdk won't keep a history of tokens, and only compares against the previous and most recent.

jasdel (commented):
Changes look good, but had two main comments:

1.) We probably want the `nextToken` member to be the member symbol of the API model. This complicates checking for empty string, but that's solvable by special casing string to check for empty string, and nil for everything else. My [work with the dense list map change uses a helper](https://github.com/jasdel/smithy-go/blob/23f470b1267ba22aae5dccd13bc998adcb28c1ec/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoValueAccessUtils.java#L111-L148), that regretfully isn't available here yet that would simplify this. 

2.) The `done` member being able to reactivate is odd, i think.  If the SDK detects end of pagination, for any reason, the paginator should forcibly stop there. Arguably it probably  should also return an error if it is called when it is known that there are no more pages.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
```diff
@@ -0,0 +1,250 @@
...
+                        writer.openBlock("if $L {", "}", nilGuard.toString(), setToken::run);
+                    } else {
+                        setToken.run();
+                    }
+
+                    writer.openBlock("if p.options.StopOnDuplicateToken && "
+                            + "params.$L != nil && p.nextToken != nil && "
+                            + "*params.$L == *p.nextToken {", "}", inputMember, inputMember, () -> {
+                        writer.write("p.nextToken = nil");
+                    });
```
jasdel, line 196: I'm not sure this will do the correct behavior. It looks like the `p.nextToken` will only be updated if `outputPath` value is not nil.  This would mean when `outputPath` has a nil value, `p.nextToken` won't be updated. Potentially looping for ever.

To address this i think this check could be simplified, to always set `p.nextToken` with the value of `outputPath`, then do the checks for duplicate tokens afterwards.

```go
prevToken := p.nextToken
p.nextToken = result.Foo

if p.options.StopOnDuplicateToken && prevToken != nil && p.nextToken != nil && *prevToken == *p.nextToken {
     p.nextToken = nil
}
```

Also side note. The `p.options.StopOnDuplicateToken` probably should only be generated when `nextToken` type is a `string` shape. We don't want to get into comparing list/maps like in DynamoDB's paginators.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
```diff
@@ -0,0 +1,250 @@
...
+            if (pageSizeMember != null) {
+                pageSizeMember.getMemberTrait(model, DocumentationTrait.class).ifPresent(documentationTrait -> {
+                    writer.writeDocs(documentationTrait.getValue());
+                });
+                writer.write("Limit $P", symbolProvider.toSymbol(pageSizeMember));
+                writer.write("");
+            }
+            writer.writeDocs("Set to true if pagination should stop if the service returns a pagination token that "
+                    + "matches the most recent token provided to the service.");
+            writer.write("StopOnDuplicateToken bool");
```
jasdel, line 223: We probably want to only generated the `StopOnDuplicateToken` on string `nextToken` shapes since those are the ones that are comparable.

jasdel (changes requested):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
skmcgrail: Well `outputPath` can be nested in the response shape so you must check for nils to avoid panics. The solution here would be if you can't traverse to the token location in the path then just nil out next token to mark the paging as done.

skmcgrail: Well there is a couple of things here, yes you are correct about the corner case here for StopDuplicateToken. But there is a corner case for there being a member that is nil along the path, in that case we need to set next token to nil as well.

skmcgrail: As an example here is a case where the token is nested in a member of the output structure:
```go
prevToken := p.nextToken
	if result.EngineDefaults != nil {
		p.nextToken = result.EngineDefaults.Marker
	} else {
		p.nextToken = nil
	}

	if p.options.StopOnDuplicateToken && prevToken != nil && p.nextToken != nil && *prevToken == *p.nextToken {
		p.nextToken = nil
	}
```

Here is the simple case where we don't need to have a guard check:
```go
	prevToken := p.nextToken
	p.nextToken = result.Marker
	if p.options.StopOnDuplicateToken && prevToken != nil && p.nextToken != nil && *prevToken == *p.nextToken {
		p.nextToken = nil
	}
```

jasdel: Ah, good point, didn't see the case where there were multiple levels of members! I guess the alternative would be to always set `p.nextToken` to nil up front, but the else is fine.

jasdel (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Paginators.java
skmcgrail: Ah yeah I can make the change to nil it up front after storing the reference, removes the else branch and cleans it up a bit.

jasdel (commented):
change looks good. Looks like just waiting on updated smithy change.

PR #237, opened 2020-12-01
Adds waiter support to smithy-go

skotambkar:
While reviewing, please also review the generated waiters for weather service: https://gist.github.com/skotambkar/73f4f0825f9960fa29834753f1f59c30. (updated)

*Description of changes:*
  * Adds support for `waitable` trait. 

TODO:
[x] move waiter middleware to be waiter-specific and code generated. This is required to remove the use of empty interfaces on waiter state mutator options. 
[x] Test waiter generation with weather service.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+                            "WaiterStateMutator is mutator function option that can be used to override the "
+                                    + "service defined waiter-behavior based on operation output, or returned error. "
+                                    + "The mutator function is used by the waiter to decide if a state is retryable "
+                                    + "or a terminal state.\n\nBy default service-modeled waiter state mutators "
+                                    + "will populate this option. This option can thus be used to define a custom "
+                                    + "waiter state with fall-back to service-modeled waiter state mutators."
+                    );
+                    writer.write(
+                            "WaiterStateMutator func(context.Context, $P, $P, error) "
+                                    + "(bool, error)", inputSymbol, outputSymbol);
```
jasdel, line 211: This isn't really a a waiter state mutator, since it doesn't change the states. This option checks the response to determine behavior of the waiter.

The intention of this option is a function that determines if the response puts the waiter into a fail state (exit with error), success state (exit without error), or retry state (can retry).

I think a name like `ContinueWaiting`, or `Retryable` would be more clear of the intention and capabilities. Since the `bool` returned specifies if the waiter could continue to retry, and the `error` is for some failure state. I think `Retryable` would make this more similar to the design of the SDK's attempt retryer.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+                () -> {
+                    writer.write("options := $T{}", waiterOptionsSymbol);
+                    writer.addUseImports(SmithyGoDependency.TIME);
+
+                    // set defaults
+                    writer.write("options.MinDelay = $L * time.Second", waiter.getMinDelay());
+                    writer.write("options.MaxDelay = $L * time.Second", waiter.getMaxDelay());
+                    // set defaults not defined in model
+                    writer.write("options.MaxWaitTime = $L * time.Second", DEFAULT_MAX_WAIT_TIME_IN_SECONDS);
+                    writer.write("options.MaxAttempts = $L ", DEFAULT_MAX_ATTEMPTS);
```
jasdel, line 274: Does smithy not model `MaxAttempts`? I believe this was a value that was modeled by the v1 SDK's waiter definitions.

https://github.com/aws/aws-sdk-go/blob/98d117b72099e244e566a83870e141cde6f7d8a7/models/apis/s3/2006-03-01/waiters-2.json#L4-L8

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+                clientSymbol, WAITER_INVOKER_FUNCTION_NAME, inputSymbol, waiterOptionsSymbol,
+                () -> {
+                    writer.write("options := $T{}", waiterOptionsSymbol);
+                    writer.write("options.MinDelay = w.options.MinDelay");
+                    writer.write("options.MaxDelay = w.options.MaxDelay");
+                    writer.write("options.MaxWaitTime = w.options.MaxWaitTime");
+                    writer.write("options.MaxAttempts = w.options.MaxAttempts");
+                    writer.write("options.EnableLogger = w.options.EnableLogger");
+                    writer.write("options.WaiterStateMutator = w.options.WaiterStateMutator");
+
```
jasdel, line 334: This isn't really needed since `options` is a value, not pointer member. Creating a local variable of `options` from the `w.options` is equivalent to this since all values of the options structure are value.

```go
options := w.options
```

If we end up ever adding pointer values that need deeper copies similar to the API Client's options struct a `clone` method created when needed.`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+                            "}", () -> {
+                                writer.write("fn(&options)");
+                            });
+
+                    writer.openBlock("_, err := w.client.$T(ctx, params, func (o *Options) { ", "})",
+                            operationSymbol, () -> {
+                                writer.openBlock("o.APIOptions = append(o.APIOptions, "
+                                                + "func(stack *middleware.Stack) error {", "})",
+                                        () -> {
+                                            writer.openBlock("return stack.Finalize.Add(&$L{", "}, middleware.Before)",
```
jasdel, line 345: Since this middleware needs to go before the retryer middleware, it should probably callout that explicitly by name and use `Insert` instead of `Add`.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+            w.write("var remainingTime = m.maxWaitTime");
+
+            // start loop
+            w.write("for {");
+
+            w.writeDocs("check number of attempts");
+            w.write("if m.maxAttempts == attempt { break }");
+            w.write("");
+            w.write("attempt++");
+            w.write("");
```
jasdel, line 428: Should this also check for max wait time, e.g. remainingTime?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+
+            w.write("attemptInput := in");
+            w.write("attemptInput.Request = m.requestCloner(attemptInput.Request)");
+
+            w.write("");
+
+            // start retry only behavior
+            w.write("if attempt > 1 {");
```
jasdel, line 436: unfinished comment?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+            w.write("remainingTime = remainingTime - delay");
+            w.write("");
+
+            w.writeDocs("rewind transport stream");
+            w.write("if rewindable, ok := in.Request.(interface{ RewindStream() error }); ok {");
+            w.write("if err := rewindable.RewindStream(); err != nil {");
+            w.addUseImports(SmithyGoDependency.FMT);
+            w.write("return out, metadata, fmt.Errorf(\"failed to rewind transport stream for retry, %w\", err)\n}");
+            w.write("}");
+            w.write("");
```
jasdel, line 455: `w.openBlock` might make these nested layers bit easier to read.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+            w.write("}");
+
+            w.openBlock("if m.enableLogger {", "}", () -> {
+                w.write("logger.Logf(logging.Debug, \"max retry attempts exhausted, max %d\", attempt)");
+            });
+
+            w.addUseImports(SmithyGoDependency.FMT);
+            w.write("return out, metadata, "
+                    + "fmt.Errorf(\"exhausted all retry attempts while waiting for the resource state\")");
+
```
jasdel, line 506: is a log message really needed if an error will will duplicate it?

jasdel: Also this section seems to be also returning on the first attempt regardless if retryable or not.

i.e i'd expect a `continue` or similar if `retryable` is true.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+            w.writeDocs("attempt request");
+            w.write("out, metadata, err = next.HandleFinalize(ctx, attemptInput)");
+            w.write("");
+
+            w.writeDocs("check for state mutation");
+            w.write("output := out.Result.($P)", outputSymbol);
+            w.write("retryable, err := m.waiterStateMutator(ctx, m.params, output, err)");
+            w.write("");
+
+            w.openBlock("if !retryable || err != nil {", "}", () -> {
```
jasdel, line 486: nit not really anything being mutated, but checking if retryable.

also don't really need the extra `write("")` before the error check so they are

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+            w.write("}");
+
+            w.write("");
+            w.writeDocs("attempt request");
+            w.write("out, metadata, err = next.HandleFinalize(ctx, attemptInput)");
+            w.write("");
+
+            w.writeDocs("check for state mutation");
+            w.write("output := out.Result.($P)", outputSymbol);
+            w.write("retryable, err := m.waiterStateMutator(ctx, m.params, output, err)");
```
jasdel, line 483: Should have error handling for failed type assertion of `out.Result.($P)`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,835 @@
...
+                            case "output":
+                                writer.addUseImports(SmithyGoDependency.GO_JMESPATH);
+                                writer.addUseImports(SmithyGoDependency.FMT);
+
+                                Matcher.OutputMember outputMember = (Matcher.OutputMember) matcher;
+                                String path = outputMember.getValue().getPath();
+                                String expectedValue = outputMember.getValue().getExpected();
+                                PathComparator comparator = outputMember.getValue().getComparator();
+                                writer.openBlock("if err == nil {", "}", () -> {
+                                    writer.write("pathValue, err :=  jmespath.Search($S, output)", path);
```
jasdel, line 577: is jmespath just used a place holder until the smithy code gen can be done for jmespath? so SDK doesn't need to rely on this at runtime.

jasdel: If at all possible the SDK should be using code generation for these compariators, with no runtime reflection with jmespath or similar.

jasdel:
Probably https://gist.github.com/skotambkar/e4d4ff1404df8f78c598a9471f9212a4#file-listcitieswaiter-go-L7-L32 needs a `APIOptions` member to provide additional api options like stack mutators.

waiters/waiter.go
```diff
@@ -0,0 +1,39 @@
...
+	"time"
+)
+
+// ComputeDelay will compute delay between waiter attempts using the min delay, max delay, remaining time and
+// current attempt count. Will return a delay value in seconds.
+func ComputeDelay(minDelay, maxDelay, remainingTime time.Duration, attempt int64) time.Duration {
+	delay := time.Duration(math.Min(
+		maxDelay.Seconds(),
+		minDelay.Seconds()*math.Pow(2, float64(attempt-1))),
+	) * time.Second
```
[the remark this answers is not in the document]
skotambkar: Tried reusing component that helps avoid math.Pow usage.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
skotambkar: Created https://github.com/awslabs/smithy-go/issues/238 to track this

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,652 @@
...
+                    writer.write("options.MinDelay = $L * time.Second", waiter.getMinDelay());
+                    writer.write("options.MaxDelay = $L * time.Second", waiter.getMaxDelay());
+                    writer.write("options.Retryable = $L", generateRetryableName(waiterName));
+                    writer.write("");
+
+                    writer.openBlock("for _, fn := range optFns {",
+                            "}", () -> {
+                                writer.write("fn(&options)");
+                            });
+
```
skmcgrail, line 237: Should we reset any of the defaults back if a user gives an invalid range? Like a negative value for min or max delay?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
skmcgrail: I was under the impression that out MVP would not require jmespath? Are we saying that we are potentially going to ship with the dependency?

waiters/waiter.go
```diff
@@ -0,0 +1,25 @@
+package waiters
```
skmcgrail, line 1: Should this package name be singular `waiter`?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/InterfaceGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.PaginatedTrait;
+import software.amazon.smithy.waiters.WaitableTrait;
+
+/**
+ * Generates API client Interfaces as per API operation.
+ */
+public class InterfaceGenerator implements GoIntegration {
```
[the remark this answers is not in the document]
skotambkar: Prefered `InterfaceGenerator` over `OperationInterfaceGenerator` as shorter. Also wanted to keep it generic naming to scale in the future.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
skotambkar: Since waiters are mostly sleeping, think it's better to error out if invalid delays override the default values. Switching back to default might cost them more than they wanted.   

Similar rationale behind why maxWaitTime is required now.  I can add validation to exit early if `maxdelay` and `mindelay` values are invalid. Thoughts?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
jasdel: I think for now, and come back to it at a later given this is all internal unexported behavior.

waiter/waiter.go
```diff
@@ -0,0 +1,41 @@
...
+		return 0, nil
+	}
+
+	// [0.0, 1.0) * 2 ^ attempt-1
+	ri := 1 << uint64(attempt-1)
+
+	delay := time.Duration(math.Min(
+		maxDelay.Seconds(),
+		minDelay.Seconds()*float64(ri)),
+	) * time.Second
```
jasdel, line 34: This is susceptible to integer overflow if number of attempts is large enough.

To compute the max attempt ceiling for max delay this should use Log2, and then check the number of attempts against the max attempt ceiling.

```go
// Get attempt ceiling to prevent integer overflow.
attemptCeiling := math.Log2(maxDelay.Seconds())
if attempt > int(attemptCeiling) {
     attempt = attemptCeiling
}

// Compute exponential delay based on attempt.
ri := 1 << uint64(attempt-1)
delaySeconds: = float(ri)

// remaining work
if remainingTime-delay <= minDelay {
	delay = remainingTime - minDelay
}

return delay, nil
```

jasdel (changes requested):

waiter/waiter.go
jasdel: Algorithm was wrong for attempt ceiling its should be, 

It should be `minDelay * 2^x = maxDelay`.  Which to solve for x need to do

```
attemptCeiling = (math.Log(float64(maxDelay / minDelay)) / math.Log(2)) + 1
```

https://play.golang.org/p/wgLPZ1rF5Hr

jasdel: Also since this would be using Log, something should ensure that `minDelay` and `maxDelay` are greater than zero.

waiter/waiter.go
```diff
@@ -0,0 +1,70 @@
...
+	attemptCeiling := (math.Log(float64(maxDelay/minDelay)) / math.Log(2)) + 1
+
+	if attempt > int64(attemptCeiling) {
+		delay = maxDelay
+	} else {
+		// Compute exponential delay based on attempt.
+		// [0.0, 1.0) * 2 ^ attempt-1
+		ri := 1 << uint64(attempt-1)
+		// compute delay
+		delay = smithytime.DurationMin(maxDelay, minDelay*time.Duration(ri))
```
jasdel, line 61: `DurationMin` shouldn't be needed, it is impossible for `minDelay*time.Duration(ri)` to be greater than `maxDelay`

waiter/waiter.go
```diff
@@ -0,0 +1,70 @@
...
+	}()
+
+	// Get attempt ceiling to prevent integer overflow.
+	attemptCeiling := (math.Log(float64(maxDelay/minDelay)) / math.Log(2)) + 1
+
+	if attempt > int64(attemptCeiling) {
+		delay = maxDelay
+	} else {
+		// Compute exponential delay based on attempt.
+		// [0.0, 1.0) * 2 ^ attempt-1
```
jasdel, line 58: Nit this comment about `[0.0, 1.0)` should be moved down to the jitter logic, not relevant here.

waiter/waiter.go
```diff
@@ -0,0 +1,70 @@
...
+		return 0, true, nil
+	}
+
+	// check if this is the last attempt possible and compute delay accordingly
+	defer func() {
+		if remainingTime-delay <= minDelay {
+			delay = remainingTime - minDelay
+			done = true
+		}
+	}()
```
jasdel, line 49: Since there are no early exits with this implementation this defer can be removed, and its logic put before the return.

waiter/waiter.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// as we use log, ensure min delay and maxdelay are atleast 1 ns
+	if minDelay < 1 {
+		minDelay = 1
+	}
+
+	// if max delay is less than 1 ns, return 0 as delay
+	if maxDelay < 1 {
+		return 0, true, nil
+	}
```
jasdel, line 41: Should the SDK support a `maxDelay` of < 1 ?  This effectively would be no delays which i don't think are intended to be supported, and 0, is treated as a value that will be substituted with a default

waiter/waiter_test.go
```diff
@@ -0,0 +1,122 @@
...
+			if e, a := c.expectedMinAttempts, len(delays); e > a {
+				t.Logf("%v", delays)
+				t.Fatalf("expected minimum attempts to be %v, got %v", e, a)
+			}
+
+			for i, expectedDelay := range c.expectedMaxDelays {
+				if e, a := expectedDelay*time.Second, delays[i]; e < a {
+					t.Fatalf("attempt %d : expected delay to be less than %v, got %v", i+1, e, a)
+				}
+			}
```
jasdel, line 91: What do you think about updating this to check both min and max to ensure that the value always falls within  the [min, jitter+delay] range.

waiter/waiter.go
```diff
@@ -0,0 +1,70 @@
...
+		// [0.0, 1.0) * 2 ^ attempt-1
+		ri := 1 << uint64(attempt-1)
+		// compute delay
+		delay = smithytime.DurationMin(maxDelay, minDelay*time.Duration(ri))
+	}
+
+	if delay != minDelay {
+		// randomize to get jitter between min delay and delay value
+		delay = time.Duration(rand.Int63n(int64(delay-minDelay))) + minDelay
+	}
```
jasdel, line 67: This should probably use `crypto/rand` instead of `math/rand`.  Math rand will be predictable order, and results. probably not big deal, but SDK retry uses `crypto/rand` for this reason.

Consider taking a similar path to the SDK's [internal/rand#Float64](https://github.com/aws/aws-sdk-go-v2/blob/8baa5f464a5dee3456ad62d9caa978e3c2572c8b/internal/rand/rand.go#L19-L27) where it also takes a rand reader allowing it to be stubbed out for testing.

jasdel (changes requested):

jasdel (changes requested):

waiter/waiter.go
```diff
@@ -0,0 +1,66 @@
...
+		// Compute exponential delay based on attempt.
+		ri := 1 << uint64(attempt-1)
+		// compute delay
+		delay = minDelay * time.Duration(ri)
+	}
+
+	if delay != minDelay {
+		// randomize to get jitter between min delay and delay value
+		// [0.0, 1.0) * [minDelay, delay]
+		d, err := rand.CryptoRandInt63n(int64(delay - minDelay))
```
jasdel, line 52: minor comment clean up needed, `// [0.0, 1.0) * [minDelay, delay]` isn't needed.

waiter/waiter_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+			for i, expectedDelay := range c.expectedMaxDelays {
+				if e, a := expectedDelay*time.Second, delays[i]; e < a {
+					t.Fatalf("attempt %d : expected delay to be less than %v, got %v", i+1, e, a)
+				}
+
+				if e, a := c.minDelay, delays[i]; e > a && c.maxWaitTime > c.minDelay {
+					t.Fatalf("attempt %d : expected delay to be more than %v, got %v", i+1, e, a)
+				}
+			}
```
jasdel, line 102: To help ensure `jitter` is being applied, consider adding a check to assert that `delay[i]` is less than `expectedDelay`. One way to help ensuring stable test would be to mock out `smithy-go/rand#Reader` with a mock reader with a static seed, like [math/rand#New](https://golang.org/pkg/math/rand/#New)

```go
import "github.com/awslabs/smithy-go/rand"
import mathrand "math/rand"

// ...

rand.Reader = mathrand.New(mathrand.NewSource(1))
```

Then the jitter produced by `ComputeDelay` will be consistent for the test.  Probably will need to set `smithy-go/rand#Reader` inside of the `t.Run` loop since maps are not ordered, and inconsistent order will mess with the expected results.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,690 @@
...
+                        "%s calls the waiter function for %s waiter. The maxWaitDur is the maximum wait duration "
+                                + "the waiter will wait. The maxWaitDur is required and must be greater than zero.",
+                        WAITER_INVOKER_FUNCTION_NAME, waiterName)
+        );
+        writer.openBlock(
+                "func (w $P) $L(ctx context.Context, params $P, maxWaitDur time.Duration, optFns ...func($P)) error {",
+                "}",
+                clientSymbol, WAITER_INVOKER_FUNCTION_NAME, inputSymbol, waiterOptionsSymbol,
+                () -> {
+                    writer.openBlock("if maxWaitDur == 0 {", "}", () -> {
```
jasdel, line 299: nit since `time.Duration` is an signed int64, this should be `<= 0`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,690 @@
...
+                    writer.write("options := w.options");
+
+                    writer.openBlock("for _, fn := range optFns {",
+                            "}", () -> {
+                                writer.write("fn(&options)");
+                            });
+                    writer.write("");
+
+                    // validate values for MaxDelay from options
+                    writer.openBlock("if options.MaxDelay == 0 {", "}", () -> {
```
jasdel, line 313: probably want to handle negative as well `<= 0`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -0,0 +1,690 @@
...
+                    writer.write("var attempt int64");
+                    writer.openBlock("for {", "}", () -> {
+                        writer.write("");
+
+                        // handle retry delay computation, sleep. Skip if minDelay is lesser than or equal to 0.
+                        writer.openBlock("if attempt > 0 && options.MinDelay > 0 {", "}", () -> {
+                            writer.openBlock("if remainingTime < options.MinDelay || remainingTime <= 0 {", "}", () -> {
+                                writer.write("break");
+                            });
+                            writer.write("");
```
jasdel, line 346: kind of an edge case here, but if `MinDelay` is `0`, then i think this will loop forever, until the context is canceled. since `remainingTime` is only decremented based on delay.  To address this, consider moving `remainingTime` check a level higher

In addition, should also decrement `remainingTime` based on the duration the request took to execute, in addition to the `delay` sleep for.  Without this `remainingTime` only tracks delay decrements not the time the operations took in addition. Which means that the context timeout would also always trigger before the remainingTime is < minDelay.  Ideally we'd prefer the remainingTime < minDelay to be the case that causes the waiter to return.

waiter/waiter_test.go
skotambkar: > consider adding a check to assert that delay[i] is less than expectedDelay
we do check that delay[i] is less than expectedDelay.

jasdel (approved):

PR #239, opened 2020-12-10
Generate Union Documentation and Output Usage Example

skmcgrail:
* Adds additional package documentation to indicate that types that satisfy union interface types.
* Adds an example for package documentation that indicates how a union can be used when received in output responses.

Weather example: https://gist.github.com/skmcgrail/fcea7b51cb2bd8da6f4e28db64ae6fe1

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDelegator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDelegator.java
@@ -70,7 +70,7 @@ public final class GoDelegator {
     /**
      * Gets a previously created writer or creates a new one if needed.
      *
-     * @param shape Shape to create the writer for.
+     * @param shape          Shape to create the writer for.
      * @param writerConsumer Consumer that accepts and works with the file.
      */
     public void useShapeWriter(Shape shape, Consumer<GoWriter> writerConsumer) {
@@ -81,7 +81,7 @@ public final class GoDelegator {
     /**
      * Gets a previously created writer or creates a new one for the a Go test file for the associated shape.
      *
-     * @param shape Shape to create the writer for.
+     * @param shape          Shape to create the writer for.
      * @param writerConsumer Consumer that accepts and works with the file.
      */
     public void useShapeTestWriter(Shape shape, Consumer<GoWriter> writerConsumer) {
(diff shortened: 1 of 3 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoDelegator.java
```diff
@@ -93,16 +93,39 @@ public void useShapeTestWriter(Shape shape, Consumer<GoWriter> writerConsumer) {
...
+        StringBuilder b = new StringBuilder(filename);
+        b.insert(filename.lastIndexOf(".go"), "_public_test");
+        filename = b.toString();
+
+        symbol = symbol.toBuilder()
+                .definitionFile(filename)
+                .namespace(symbol.getNamespace() + "_test", symbol.getNamespaceDelimiter())
+                .build();
 
         useShapeWriter(symbol, writerConsumer);
```
jasdel, line 122: minor semantics, Should `Public` be `Exported` since that is the phrasing in Go?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/UnionGenerator.java
```diff
@@ -78,6 +89,42 @@ public void run() {
...
+            writer.openBlock("switch v := union.(type) {", "}", () -> {
+                for (MemberShape member : members) {
+                    Symbol targetSymbol = symbolProvider.toSymbol(model.expectShape(member.getTarget()));
+                    Symbol memberSymbol = SymbolUtils.createValueSymbolBuilder(symbolProvider.toMemberName(member),
+                            symbol.getNamespace()).build();
+
+                    writer.openBlock("case *$T:", "", memberSymbol, () -> {
+                        writer.write("_ = v.Value // Value is $L", targetSymbol.getName());
+                    });
+                }
```
jasdel, line 114: might be helpful to also document the type of member `Value`?

skmcgrail: `targetSymbol.getName()` is the type, so `string`, `[]byte` or `NameOfStructType` etc

jasdel (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/UnionGenerator.java
```diff
@@ -78,6 +89,42 @@ public void run() {
...
+
+            writer.writeDocs("type switches can be used to check the union value");
+            writer.openBlock("switch v := union.(type) {", "}", () -> {
+                for (MemberShape member : members) {
+                    Symbol targetSymbol = symbolProvider.toSymbol(model.expectShape(member.getTarget()));
+                    Symbol memberSymbol = SymbolUtils.createValueSymbolBuilder(symbolProvider.toMemberName(member),
+                            symbol.getNamespace()).build();
+
+                    writer.openBlock("case *$T:", "", memberSymbol, () -> {
+                        writer.write("_ = v.Value // Value is $L", targetSymbol.getName());
```
jasdel, line 112: Instead of target symbol name, maybe this should be the type of the target symbol for value?

jasdel: Only difference between using $T vs getName is that $T probably will include the `types` package if needed?

skmcgrail: I believe this would have to modified some, as I believe the default out for symbol would be the full path including the `github.com/aws/aws-sdk-go-v2` portion of the path.

PR #241, opened 2020-12-16
Normalize map keys for shapes targeted by httpPrefixHeaders during deserialization 

skmcgrail:
Normalizes map keys for shapes targeted by httpPrefixHeaders during deserialization. Also adds a comment when writing docs that are affected by this.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
 import java.util.StringJoiner;
 import java.util.function.BiFunction;
 import java.util.logging.Logger;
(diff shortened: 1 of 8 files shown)


jasdel (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/UnionGenerator.java
```diff
@@ -109,7 +109,7 @@ public void generateUnionExamples(GoWriter writer) {
                             symbol.getNamespace()).build();
 
                     writer.openBlock("case *$T:", "", memberSymbol, () -> {
-                        writer.write("_ = v.Value // Value is $L", targetSymbol.getName());
+                        writer.write("_ = v.Value // Value is $T", targetSymbol);
```
skotambkar, line 112: nit: Think `$L ` would have worked correctly if you used just `targetSymbol` instead of getName

skotambkar (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/UnionGenerator.java
jasdel: `$L` will write and the full import path + types, `github.com/aws/aws-sdk-go-v2/service/foo/types#SomeType`, which arguably is never the correct behavior, but its what $L does. Should look at updating $L to do "right" thing.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
```diff
@@ -290,24 +294,60 @@ boolean writePackageShapeDocs(Shape shape) {
...
+            }
+        }
+
+        if (member.getMemberTrait(model, RequiredTrait.class).isPresent()) {
+            if (hasDocs) {
+                writeDocs("");
+            }
+            writeDocs("This member is required.");
+            hasDocs = true;
+        }
```
jasdel, line 334: minor nit, if there are no docs, this would be 

```
// value conforms to media type...
// this field is required
```

Whereas if there were docs it would be
```
// something really cool to say
//
// value conforms to media type...
//
// this field is required
```

jasdel (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
skmcgrail: `hasDocs` gets set to true when of these sections is first hit if its not already true. So these would be spaced as expected.

jasdel: Thanks missed that being set! You're right

PR #242, opened 2020-12-16
Refactor encoding packages to be nested under common package

jasdel:
Updates the encoding packages, `httpbinding`, `json` and `xml` to all be nested under a common package, `encoding`


skotambkar:
why is `time` under encoding?

jasdel:
time was moved mainly because it had encoding and decode helpers. :/ though that doesn't fit well with the `SleepWithContext`. Will take a look at this.

encoding/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package encoding provides utilities for encoding values for specific
+// document encodings.
+
+package encoding
```
skotambkar, line 4: Nit: At a later point, we can consider updating this page to have link to available encoding packages

mediatype.go
```diff
@@ -1,9 +0,0 @@
-package smithy
-
-// The MediaType interface is intended to be implemented by string and
-// byte array types whose values can be defined by RFC638 media types.
-type MediaType interface {
-	// Describes the contents of the string or byte array using a media type
-	// as defined by RFC6838.
-	MediaType() string
-}
```
skotambkar, line 9: Intentional?

skotambkar (approved):

mediatype.go
jasdel: yeah this type wasn't used at all and was an artifact that was never deleted from an old design.

PR #251, opened 2021-01-06
 middleware: Add utility for stack scoped stack values

jasdel:
Adds utilities for values that are intended to be scoped to individual stacks. Provides utilities for bulk clearing these values as well.

Related to aws/aws-sdk-go-v2#914

* [x] Needs unit tests.
* [x] Should RecordMiddleware be stack scoped?

---
#### Updated to be stack scoped
* hostnameImmutableKey
* hostPrefixDisableKey

#### Not changed
* Logger

### DIFF
--- a/middleware/stack_values.go
+++ b/middleware/stack_values.go
@@ -1,0 +1,100 @@
+package middleware
+
+import (
+	"context"
+	"reflect"
+	"strings"
+)
+
+// WithStackValue adds a key value pair to the context that are intended to be
+// scoped to a stack. Use ClearStackValues to get a new context with all stack
+// values cleared.
+func WithStackValue(ctx context.Context, key, value interface{}) context.Context {
+	md, _ := ctx.Value(stackValuesKey{}).(*stackValues)
+
+	md = withStackValue(md, key, value)
+	return context.WithValue(ctx, stackValuesKey{}, md)
+}
+
+// ClearStackValues returns a context without any stack values.
+func ClearStackValues(ctx context.Context) context.Context {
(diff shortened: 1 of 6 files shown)


middleware/stack_values.go
```diff
@@ -0,0 +1,101 @@
...
+
+// GetStackValues returns the value pointed to by the key within the stack
+// values, if it is present.
+func GetStackValue(ctx context.Context, key interface{}) interface{} {
+	v := ctx.Value(stackValuesKey{})
+	if v == nil {
+		return nil
+	}
+
+	md := v.(*stackValues)
```
skmcgrail, line 32: I think we actually need to check that md itself is also not nil.

skmcgrail: As a small example on the different between a nil empty interface and a interface containing a pointer to something that is nil:

```package main

type Foo struct {}

func (f *Foo) DoThing() {
	if f == nil {
		panic("Foo is nil")
	}
}

func main() {
	var x interface{}

	if x == nil {
		println("x is nil")
	}

	x = (*Foo)(nil)

	if x != nil {
		println("x is not nil")
	}

	f := x.(*Foo)
	f.DoThing()
}
```

This prints
```
x is nil
x is not nil
panic: Foo is nil
```

jasdel: Simplified this to the following which solves this issue.

```go
	md, _ := ctx.Value(stackValuesKey{}).(*stackValues)
	if md == nil {
		return nil
	}
```

skmcgrail (approved):

PR #253, opened 2021-01-07
Add support for validating union members in required trait chains.

skmcgrail:
Adds support for validating union members that are part of a required trait chain.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ValidationGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ValidationGenerator.java
@@ -36,7 +36,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.Symbol;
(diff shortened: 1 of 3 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ValidationGenerator.java
```diff
@@ -227,13 +233,49 @@ private void generateShapeValidationFunctions(
...
                     case UNION:
-                        // TODO: Implement Union support
+                        UnionShape unionShape = shape.asUnionShape().get();
+                        Symbol unionSymbol = symbolProvider.toSymbol(unionShape);
+
+                        Set<MemberShape> memberShapes = unionShape.getAllMembers().values().stream()
+                                .filter(memberShape ->
+                                        shapesWithHelpers.contains(model.expectShape(memberShape.getTarget()).getId()))
+                                .collect(Collectors.toCollection(TreeSet::new));
+
```
jasdel, line 249: Will this filter the member shapes to only those will with validation traits?

jasdel (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ValidationGenerator.java
skmcgrail: Correct, this filters to only include members who refer to a shape that either have a member marked as required, or are an intermediate shape that can contain references to another shape that has such a trait.

PR #261, opened 2021-01-15
codegen: Update generated serde to skip unmodeled input/outputs

### LINKED ISSUE #262: codegen: SyntheticClone's getArchetype should return Optional<ShapeId>
The `SynthenticClone`'s `getArchetype` method should return an `Optional<ShapeId>`. To more correctly show this is an optional member.

_Originally posted by @skmcgrail in https://github.com/aws/smithy-go/pull/261#discussion_r558622430_

jasdel:
Updates the SDK's generated serializers and deserializers to skip unmodeled
input and output shapes for operations.

This issue caused generated deserializers to attempt to deserialize
operation responses that were not modeled, resulting in an error.
Whereas the SDK should of ignored the response if one was present for
operations without a modeled output.

Related to https://github.com/aws/aws-sdk-go-v2/issues/1047
Fixes #262

#### Input
Unmodeled Input shape serializer functions are not generated.

This update changes API protocol operation serializers for HTTP binding protocols to skip serialization of the request document if the operation's input was not modeled.

This change is **not** applied to HTTP RPC protocol serializations. This is due to the AwsQuery protocol requiring some protocol specific elements to always be serialized regardless if an input is modeled or not.

#### Output
Unmodeled output shape serializer functions are not generated.

This update changes all API protocol operation deserializers to discard the response body if the operation's output was not modeled.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
@@ -160,7 +160,7 @@ public final class CodegenUtils {
         }
 
         SyntheticClone synthClone = optional.get();
-        return synthClone.getArchetype() == null;
+        return !synthClone.getArchetype().isPresent();
     }
 
     /**
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
@@ -23,6 +23,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Optional;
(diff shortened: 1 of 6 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenUtils.java
```diff
@@ -145,6 +146,22 @@ public static String getSyntheticTypeNamespace() {
...
+     * @return if the shape is synthetic clone, but not based on a specific shape.
+     */
+    public static boolean isStubSyntheticClone(Shape shape) {
+        Optional<SyntheticClone> optional = shape.getTrait(SyntheticClone.class);
+        if (!optional.isPresent()) {
+            return false;
+        }
+
+        SyntheticClone synthClone = optional.get();
+        return synthClone.getArchetype() == null;
```
skmcgrail, line 163: Minor, but we probably want to circle back on this later and have this getter return this as a Optional<T>

skotambkar (approved):

PR #266, opened 2021-02-04
Add LoggerFunc to logging package.

skmcgrail:
Adds `LoggerFunc` type to the logging package. Allows for easily satisfying the `Logger` interface by wrapping a function.

### DIFF
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -21,6 +21,14 @@ type Logger interface {
 	Logf(classification Classification, format string, v ...interface{})
 }
 
+// LoggerFunc is a wrapper around a function to satisfy the Logger interface.
+type LoggerFunc func(classification Classification, format string, v ...interface{})
+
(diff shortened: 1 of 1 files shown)


logging/logger.go
```diff
@@ -21,6 +21,14 @@ type Logger interface {
 	Logf(classification Classification, format string, v ...interface{})
 }
 
+// LoggerFunc is a wrapper around a function to satisfy the Logger interface.
+type LoggerFunc func(classification Classification, format string, v ...interface{})
```
skotambkar, line 25: what does `format` represent here? Slightly confusing naming. Should it be msg or something?

skmcgrail: `format` as in the format string. Given that this is has the `f` suffix similar to `fmt.Printf` etc this is the format string that the values will be templated into.

skotambkar (approved):

PR #272, opened 2021-02-17
New Config Field Resolver Design

skmcgrail:
This redesign config field resolver to allow them to apply to two locations: the New client constructor or invokeOperation when an operation is invoked. Each resolver can then target either the initial set of client field options prior to user mutation, or a resolver can target the final mutated version of options (should be used sparingly).


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -222,9 +244,27 @@ private void generateClientInvokeOperation() {
 
             generateConstructStack();
             writer.write("options := c.options.Copy()");
+            writer.write("");
```
jasdel, line 247: Minor nit, if this newline isn't there wouldn't have as many files/modules changed in generated code.

jasdel (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
skmcgrail: Will update

PR #282, opened 2021-04-02
Fix RingBuffer panic due to out of bounds index

skmcgrail:
Fixes an index out of bounds in the RingBuffer implementation.

Fixes https://github.com/aws/aws-sdk-go-v2/issues/1200

### DIFF
--- a/io/ringbuffer.go
+++ b/io/ringbuffer.go
@@ -45,7 +45,7 @@ func (r *RingBuffer) Write(p []byte) (int, error) {
 		r.end++
 		r.size++
 	}
-	return r.size, nil
+	return len(p), nil
 }
 
 // Read copies the data on the ring buffer into the byte slice provided to the method.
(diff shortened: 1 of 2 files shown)


io/ringbuffer.go
```diff
@@ -45,7 +45,7 @@ func (r *RingBuffer) Write(p []byte) (int, error) {
 		r.end++
 		r.size++
 	}
-	return r.size, nil
+	return len(p), nil
```
skotambkar, line 48: Thanks for the change. Correct me if I am wrong, this change makes `Write` function more language idiomatic  as  it returns the number of bytes written. 

In case of ringBuffer, since it is a revolving buffer, the number of bytes written is always going to be the the length of bytes provided. My intention here was to convey the size of ringbuffer instead. 

If we do this change, may be we should export a `Size()` function on ringbuffer?

skmcgrail: Sure, probably want the signature to be`Len() int` to match others.

skotambkar (approved):

jasdel (approved):

purpleidea:
No call out of thanks in the commit for tracking down bugs? ;) Oh well.

PR #283, opened 2021-04-06
Generated Manifest of Build Artifacts

skmcgrail:
Creates a generated manifest of the build output artifacts. Introduces the ability to handle selection of the minimum required standard library, and minimum version required for a given set of dependencies. This manifest will be used to ultimately remove the responsibilities of generating the `go.mod` from the smithy-go code base and move that concern into the SDK. 

Example `generated.json` from the `smithy-go-codegen-test` test code:
```json
{
    "dependencies": {
        "github.com/aws/smithy-go": "v1.3.0",
        "github.com/jmespath/go-jmespath": "v0.4.0"
    },
    "files": [
        "api_client.go",
        "api_op_BadName.go",
        "api_op_GetCity.go",
        "api_op_GetCityAnnouncements.go",
        "api_op_GetCityImage.go",
        "api_op_GetCurrentTime.go",
        "api_op_GetForecast.go",
        "api_op_ListCities.go",
        "doc.go",
        "generated.json",
        "go.mod",
        "types/enums.go",
        "types/errors.go",
        "types/types.go",
        "types/types_exported_test.go",
        "validators.go"
    ],
    "go": "1.15",
    "module": "weather"
}
```


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
```diff
@@ -0,0 +1,157 @@
...
+        Optional<SymbolDependency> minStandard = dependencies.stream()
+                .filter(symbolDependency ->
+                        symbolDependency.getDependencyType().equals(GoDependency.Type.STANDARD_LIBRARY.toString()))
+                .max(Comparator.comparing(SymbolDependency::getVersion));
+
+        Map<StringNode, Node> manifestNodes = new HashMap<>();
+
+        Map<String, String> minimumDependencies = gatherMinimumDependencies(nonStdLib);
+
+        Map<StringNode, Node> dependencyNodes = new HashMap<>();
```
skotambkar, line 93: Should we use TreeMap's here to generate a stable list?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private static int compareGoPseudoVersions(String x, String y) {
+        if (x.equals(INCOMPATIBLE_VERSION) && y.equals(INCOMPATIBLE_VERSION)) {
+            return 0;
+        }
+        return parsePseudoVersions(x).compareTo(parsePseudoVersions(y));
+    }
+
+    private static String parsePseudoVersions(String extra) {
```
skotambkar, line 148: Could you put a comment here, to help understand the format of PseudoVersion this function is parsing.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
```diff
@@ -0,0 +1,157 @@
...
+        return SymbolDependency.gatherDependencies(symbolStream, (dx, dy) -> {
+            SemanticVersion sx = SemanticVersion.parseVersion(dx.getVersion());
+            SemanticVersion sy = SemanticVersion.parseVersion(dx.getVersion());
+
+            // This *shouldn't* happen in Go since the Go module import path must end with the major version component.
+            // Exception is the case where the major version is 0 or 1.
+            if (sx.getMajor() != sy.getMajor() && !(sx.getMajor() == 0 || sy.getMajor() == 0)) {
+                throw new CodegenException(String.format("Dependency %s has conflicting major versions",
+                        dx.getPackageName()));
+            }
```
skotambkar, line 126: Comment calls out `Exception is the case where major version is 0 or 1`. 

Do we need an extra condition - where we check for major version as 1 (similar to how we have for major version as 0).

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SemanticVersion.java
```diff
@@ -0,0 +1,327 @@
...
+     * Parse a semantic version string into a {@link SemanticVersion}.
+     *
+     * @param version the semantic version string
+     * @return the SemanticVersion representing the parsed value
+     */
+    public static SemanticVersion parseVersion(String version) {
+        String[] components = version.split("\\.", 3);
+
+        if (components.length != 3) {
+            throw newInvalidSemanticVersion();
```
skotambkar, line 137: Nit: can we return the version value in the error?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SemanticVersion.java
```diff
@@ -0,0 +1,327 @@
...
+        }
+
+        char[] parseArr = components[0].toCharArray();
+        int position = 0;
+
+        StringBuilder prefixBuilder = new StringBuilder();
+        while (position < parseArr.length && !Character.isDigit(parseArr[position])) {
+            prefixBuilder.append(parseArr[position]);
+            position++;
+        }
```
skotambkar, line 147: This seems to assume first component will follow `(a-zA-Z)*[0-9]*`, if so, may be we should break the loop when we find the first character that is a digit.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SemanticVersion.java
```diff
@@ -0,0 +1,327 @@
...
+        int patch;
+        if (extraStart > -1) {
+            patch = parseNumericIdentifier(components[2].substring(0, extraStart).toCharArray(), 0);
+        } else {
+            patch = parseNumericIdentifier(components[2].toCharArray(), 0);
+        }
+
+        String extra = null;
+        if (extraStart > -1) {
+            extra = components[2].substring(extraStart);
```
skotambkar, line 171: Nit: can be merged under line 163 check

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SemanticVersion.java
skmcgrail: Prefixes are optional, (this is just to support Go's prefixing with `v`), and I don't see why any sort of break here is required. This will only loop while the current position is not a digit. So this will effectively strip any leading text on the major version. Test cases cover this.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
```diff
@@ -0,0 +1,157 @@
...
+            List<SymbolDependency> dependencies
+    ) {
+        Path manifestFile = fileManifest.getBaseDir().resolve("generated.json");
+
+        if (Files.exists(manifestFile)) {
+            try {
+                Files.delete(manifestFile);
+            } catch (IOException e) {
+                throw new CodegenException("Failed to delete existing generated.json file", e);
+            }
```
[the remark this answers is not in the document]
skmcgrail: I'm not familiar with this exception type, what is the context? Is this a standard Java exception you found or one in Smithy?

codegen/config/checkstyle/checkstyle.xml
```diff
@@ -30,9 +30,9 @@
     </module>
 
     <!-- Files must contain a copyright header. -->
-    <module name="Header">
+    <module name="RegexpHeader">
         <property name="header"
-                  value="/*\n * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.\n"/>
+                  value="/*\n * Copyright \d\d\d\d Amazon.com, Inc. or its affiliates. All Rights Reserved.\n"/>
```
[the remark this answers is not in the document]
skmcgrail: Then we need to unnecessarily update this check every year, I think checking that a 4-digit year is present is more then sufficient for our needs.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SemanticVersion.java
skotambkar: Correct, I read the while condition incorrectly. This seems to do the right thing.

codegen/config/checkstyle/checkstyle.xml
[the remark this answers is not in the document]
skotambkar: I think that's a good enforcement for checkstyle. Don't have to update it often.  Plus what's the point of checkstyle if we end up putting incorrect years. (Ps: Smithy does enforce this)

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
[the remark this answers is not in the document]
skotambkar: It's a smithy exception.

codegen/config/checkstyle/checkstyle.xml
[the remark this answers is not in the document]
skmcgrail: Code review would catch incorrect years, second there is zero risk to the code of a year is off or incorrect, and finally the copyright dates are not actually required according to guidance. So if we want to let’s just remove the dates off them and not have to worry about this going forward.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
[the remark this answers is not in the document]
skmcgrail: The Java doc states that it is for when a node did not match the expected type or have a particular member.

codegen/config/checkstyle/checkstyle.xml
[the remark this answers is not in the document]
jasdel: Agree using the generic `\d` pattern allows more flexible "fire and forget" behave relying on the IDE to fill in the style correctly.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
```diff
@@ -0,0 +1,157 @@
...
+            List<SymbolDependency> dependencies
+    ) {
+        Stream<SymbolDependency> nonStdLib = dependencies.stream()
+                .filter(symbolDependency ->
+                        !symbolDependency.getDependencyType().equals(GoDependency.Type.STANDARD_LIBRARY.toString()));
+
+        Optional<SymbolDependency> minStandard = dependencies.stream()
+                .filter(symbolDependency ->
+                        symbolDependency.getDependencyType().equals(GoDependency.Type.STANDARD_LIBRARY.toString()))
+                .max(Comparator.comparing(SymbolDependency::getVersion));
```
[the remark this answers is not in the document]
jasdel: This would be done via a `for`/`foreach` loop right? Does Java prefer to do the streaming pattern like in current code, or more effecent for loop.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
jasdel: Assuming this file would be checked in as a part of the repo, its content would need to be stable.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -220,6 +220,8 @@ void execute() {
 
         LOGGER.fine("Generating go.mod file");
         GoModGenerator.writeGoMod(settings, fileManifest, SymbolDependency.gatherDependencies(dependencies.stream()));
+        LOGGER.fine("Generating generated.json file");
+        ManifestWriter.writeManifest(settings, fileManifest, dependencies);
```
jasdel, line 224: next step is to remove the `go.mod` generation, once we have tooling to use this generated.json file correct?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
```diff
@@ -0,0 +1,157 @@
...
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.node.StringNode;
+
+/**
+ * Generates a manifest description of the generated code, minimum go version, and minimum dependencies required.
+ */
+public final class ManifestWriter {
+    public static final String INCOMPATIBLE_VERSION = "+incompatible";
+
+    private ManifestWriter() {
```
jasdel, line 45: This concept is something we should suggest to Smithy team upstream since there is probably value for all the generated SDKs.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
jasdel: I think the idea is to allow a library dependency to be promoted form v0.x to v1.x if something else in the codegen depends on the v1.x version. Also to call out that There is no way to include both v0.x and v1.x of a module in the same application.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SemanticVersion.java
jasdel: Also which dependency if possible, but that might be not available at this scope.

codegen/smithy-go-codegen/src/test/java/software/amazon/smithy/go/codegen/SemanticVersionTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+public class SemanticVersionTest {
+
+    @Test
+    public void testSemanticVersion() {
+        SemanticVersion version = SemanticVersion.parseVersion("1.2.3");
+        assertThat(version.toString(), Matchers.equalTo("1.2.3"));
+    }
```
jasdel, line 31: Would be nice to have test that uses `gohash` format, and also test case for invalid/incomplete version

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SemanticVersion.java
skmcgrail: Yeah unfortunately not known in this scope.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
skmcgrail: Yes that is correct, I want to have the pieces together in the SDK repo for generating the `go.mod` from the `generated.json` first before we remove it from the generated artifacts here. That way we don't break out existing tooling. So there will be a follow-up PR to remove the `go.mod` when we are ready to do so.

jasdel:
` "module": "weather"` item in manifest should be the full module with import. Since there will be dependencies on this path in the `types` or other nested packages.

can have separate serviceId or similar for the folder name or toolin could split top folder of import module path. to get this path.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
```diff
@@ -0,0 +1,141 @@
...
+
+        Map<StringNode, Node> manifestNodes = new HashMap<>();
+
+        Map<String, String> minimumDependencies = gatherMinimumDependencies(nonStdLib.stream());
+
+        Map<StringNode, Node> dependencyNodes = new HashMap<>();
+        for (Map.Entry<String, String> entry : minimumDependencies.entrySet()) {
+            dependencyNodes.put(StringNode.from(entry.getKey()),
+                    StringNode.from(entry.getValue()));
+        }
```
jasdel, line 111: will these be in stable order, or is a `TreeHash` needed for that?

jasdel (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
skmcgrail: Yes these are stable, by means of line 128 `ObjectNode.objectNode(manifestNodes).withDeepSortedKeys();`

PR #291, opened 2021-05-04
Create operation input/output synthetic shapes before integrations preprocess model

jasdel:
Updates smithy-go codegen to clone operation top level input and output to be performed before other integrations can preprocess models. This ensures that the preprocessors are working on the correct operation top level shapes.


skotambkar:
Should preprocess be the one that actually adds unique operation input/output synthetic shapes? The intention for preprocess is it handles/makes changes to the resolved model  before any processing/usage takes place.

Adding unique operation input/output synthetic shapes seems like a preprocess step.

skotambkar (approved):

jasdel:
I think the intention here is that input/output shape is a requirement of the smithy-go codegen, not extra functionality that can optionally be added. Where as all of the integrations preprocess, are effectively optional.

PR #307, opened 2021-06-15
Update time parsing to not be as strict for HTTPDate and DateTime

### LINKED ISSUE #302: Incorrect time formatting
`FormatDateTime` is unsafe!

It uses the formatting string `dateTimeFormat = "2006-01-02T15:04:05.99Z"` which essentially just appends the `Z` character onto the user's local time. 

Possible fixes:

- Swap the format string to  `"2006-01-02T15:04:05.99Z07:00"` or maybe even `time.RFC3339` (probably breaking)
- Explicitly cast to UTC time (maybe breaking - but probably just bug fixing)
```
func FormatDateTime(value time.Time) string {
      return value.UTC().Format(dateTimeFormat)
 }
```
- Document that the function should be called with a UTC time (probably too conservative)

Not sure what your original intentions were so I'll leave this to you.

I originally found this after spending an hour or so debugging why calls to `cloudwatch.GetMetricData` weren't returning any data in the v2 aws SDK.

jasdel:
Updates the parsing of HTTPDate and DateTime time formats to not be as strict, and allow more variants of the time format. Including UTC offsets for DateTime.

Also updates the Format* methods of smithy-go/time package to always serialize the date as a UTC value. Without this, the value would be serialized as it is, dropping the UTC offset.

Updates the `transport/http` package's `ParseTime` function to use the updated `time.ParseHTTPDate` function since they overlapped.

Fixes #302 by changing time to UTC before formatting so no local offset time is lost.


skmcgrail (approved):

time/time_test.go
```diff
@@ -8,20 +8,67 @@ import (
...
 	}
 
-	parseTime, err := ParseDateTime(dateTime)
-	if err != nil {
-		t.Fatalf("expected no error, got %v", err)
-	}
+	for name, c := range cases {
+		t.Run(name, func(t *testing.T) {
+			formattedTimeValue := FormatDateTime(c.TimeValue)
+			t.Logf("formated: %v -> %v", c.TimeValue, formattedTimeValue)
```
skotambkar, line 45: Nit: don't need to log

time/time_test.go
```diff
@@ -8,20 +8,67 @@ import (
...
 
-	if e, a := refTime, parseTime; !e.Equal(a) {
-		t.Errorf("expected %v, got %v", e, a)
+			// Round Trip time value ensure format and parse are compatible.
+			parsedTimeValue, err := ParseDateTime(formattedTimeValue)
```
skotambkar, line 48: Can ParseDateTime function do fomatting if required?

time/time_test.go
```diff
@@ -41,6 +88,84 @@ func TestHTTPDate(t *testing.T) {
...
+		},
+		"shortened year with leading zero day": {
+			date:   "Thu, 05 Feb 21 11:04:03 GMT",
+			expect: time.Date(2021, 2, 5, 11, 04, 03, 0, time.UTC),
+		},
+	}
+
+	for name, tt := range cases {
+		t.Run(name, func(t *testing.T) {
+			result, err := ParseHTTPDate(tt.date)
```
skotambkar, line 154: Should ParseHTTPDate function call format?

time/time_test.go
jasdel: clarified offline the Parse functions wouldn't need to perform formatting, since `time.Time` value returned is the "universal" format.

PR #310, opened 2021-06-28
Smithy Document Shape Support

skmcgrail:
Adds support for Smithy Document shapes and supporting types for protocols to implement support.

AWS Protocol Support: https://github.com/aws/aws-sdk-go-v2/pull/1320

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
@@ -1,0 +1,373 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.go.codegen;
+
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.Consumer;
+
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.go.codegen.integration.ProtocolGenerator;
+import software.amazon.smithy.go.codegen.integration.ProtocolGenerator.GenerationContext;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.DocumentShape;
+
+/**
+ * Generates the Smithy document type for the service client.
+ */
+public final class ProtocolDocumentGenerator {
+    public static final String DOCUMENT_INTERFACE_NAME = "Interface";
+    public static final String NO_DOCUMENT_SERDE_TYPE_NAME = "noSmithyDocumentSerde";
+    public static final String NEW_LAZY_DOCUMENT = "NewLazyDocument";
+    public static final String INTERNAL_NEW_DOCUMENT_MARSHALER_FUNC = "NewDocumentMarshaler";
+    public static final String INTERNAL_NEW_DOCUMENT_UNMARSHALER_FUNC = "NewDocumentUnmarshaler";
+    public static final String INTERNAL_IS_DOCUMENT_INTERFACE = "IsInterface";
+    public static final String UNMARSHAL_SMITHY_DOCUMENT_METHOD = "UnmarshalSmithyDocument";
+    public static final String MARSHAL_SMITHY_DOCUMENT_METHOD = "MarshalSmithyDocument";
+
+    private static final String SERVICE_SMITHY_DOCUMENT_INTERFACE = "smithyDocument";
+    private static final String IS_SMITHY_DOCUMENT_METHOD = "isSmithyDocument";
+
+
+    private final GoSettings settings;
+    private final GoDelegator delegator;
+    private final Set<DocumentShape> documentShapes = new TreeSet<>();
+    private final Model model;
+
+    public ProtocolDocumentGenerator(
+            GoSettings settings,
+            Model model,
+            GoDelegator delegator
+    ) {
+        this.settings = settings;
+        this.model = model;
+        this.delegator = delegator;
+    }
+
+    /**
+     * Get the set of document shapes for the service.
+     *
+     * @return the service's document shapes.
+     */
+    public Set<DocumentShape> getDocumentShapes() {
+        return documentShapes;
+    }
+
+    /**
+     * Returns whether the service has any document shapes.
+     *
+     * @return whether the service has one or more document types.
+     */
+    public boolean hasDocumentShapes() {
+        return getDocumentShapes().size() > 0;
+    }
+
+    /**
+     * Add a document shape to the generator.
+     *
+     * @param shape the document shape to add.
+     */
+    public void addDocumentShape(DocumentShape shape) {
+        documentShapes.add(shape);
+    }
+
+    /**
+     *
+     */
+    public void generateStandardTypes() {
+        generateNoSerdeType();
+        generateInternalDocumentInterface();
+        generateDocumentPackage();
+    }
+
+    private void generateDocumentPackage() {
+        if (!hasDocumentShapes()) {
+            return;
+        }
+
+        writeDocumentPackage("document.go", writer -> {
+            writer.write("type $L = $T", DOCUMENT_INTERFACE_NAME,
+                            getInternalDocumentSymbol(DOCUMENT_INTERFACE_NAME))
+                    .write("");
+
+            writer.openBlock("func $L(v interface{}) $T {", "}", NEW_LAZY_DOCUMENT,
+                            getDocumentSymbol(DOCUMENT_INTERFACE_NAME), () -> {
+                                writer.write("return $T(v)",
+                                        getInternalDocumentSymbol(INTERNAL_NEW_DOCUMENT_MARSHALER_FUNC));
+                            })
+                    .write("");
+        });
+    }
+
+    private void generateNoSerdeType() {
+        Symbol noSerde = SymbolUtils.createValueSymbolBuilder("NoSerde",
+                SmithyGoDependency.SMITHY_DOCUMENT).build();
+
+        delegator.useShapeWriter(settings.getService(model), writer -> {
+            writer.write("type $L = $T", NO_DOCUMENT_SERDE_TYPE_NAME, noSerde);
+        });
+
+        delegator.useFileWriter("./types/types.go", settings.getModuleName() + "/types", writer -> {
+            writer.write("type $L = $T", NO_DOCUMENT_SERDE_TYPE_NAME, noSerde);
+        });
+    }
+
+    private void generateInternalDocumentInterface() {
+        if (!hasDocumentShapes()) {
+            return;
+        }
+
+        writeInternalDocumentPackage("document.go", writer -> {
+            Symbol serviceSmithyDocumentInterface = getInternalDocumentSymbol(SERVICE_SMITHY_DOCUMENT_INTERFACE);
+
+            writer.openBlock("type $T interface {", "}", serviceSmithyDocumentInterface,
+                            () -> writer.write("$L()", IS_SMITHY_DOCUMENT_METHOD))
+                    .write("");
+
+            Symbol internalDocumentInterface = getInternalDocumentSymbol(DOCUMENT_INTERFACE_NAME);
+            Symbol smithyDocumentMarshaler = SymbolUtils.createValueSymbolBuilder("SmithyDocumentMarshaler",
+                    SmithyGoDependency.SMITHY_DOCUMENT).build();
+            Symbol smithyDocumentUnmarshaler = SymbolUtils.createValueSymbolBuilder("SmithyDocumentUnmarshaler",
+                    SmithyGoDependency.SMITHY_DOCUMENT).build();
+
+            writer.openBlock("type $T interface {", "}", internalDocumentInterface, () -> {
+                writer.write("$T", serviceSmithyDocumentInterface);
+                writer.write("$T", smithyDocumentMarshaler);
+                writer.write("$T", smithyDocumentUnmarshaler);
+            }).write("");
+
+            writer.write("var _ $T = ($P)(nil)",
+                    serviceSmithyDocumentInterface, internalDocumentInterface);
+            writer.write("var _ $T = ($P)(nil)",
+                    smithyDocumentMarshaler, internalDocumentInterface);
+            writer.write("var _ $T = ($P)(nil)",
+                    smithyDocumentUnmarshaler,
+                    internalDocumentInterface);
+            writer.write("");
+        });
+    }
+
+    /**
+     * Generates the internal document Go package for the service client. Delegates the logic for document marshaling
+     * and unmarshalling types to the provided protocol generator using the given context.
+     *
+     * @param protocolGenerator the protocol generator.
+     * @param context           the protocol generator context.
+     */
+    public void generateInternalDocumentTypes(ProtocolGenerator protocolGenerator, GenerationContext context) {
+        if (!hasDocumentShapes()) {
+            return;
+        }
+
+        delegator.useFileWriter(getInternalDocumentFilePath("document.go"), getInternalDocumentPackage(), writer -> {
+            Symbol marshalerSymbol = getInternalDocumentSymbol("documentMarshaler",
+                    true);
+            Symbol unmarshalerSymbol = getInternalDocumentSymbol("documentUnmarshaler", true);
+
+            Symbol isDocumentInterface = getInternalDocumentSymbol(INTERNAL_IS_DOCUMENT_INTERFACE);
+
+            writeInternalDocumentImplementation(
+                    writer,
+                    marshalerSymbol,
+                    () -> {
+                        protocolGenerator.generateProtocolDocumentMarshalerUnmarshalDocument(context.toBuilder()
+                                .writer(writer)
+                                .build());
+                    },
+                    () -> {
+                        protocolGenerator.generateProtocolDocumentMarshalerMarshalDocument(context.toBuilder()
+                                .writer(writer)
+                                .build());
+                    });
+            writeInternalDocumentImplementation(writer,
+                    unmarshalerSymbol,
+                    () -> {
+                        protocolGenerator.generateProtocolDocumentUnmarshalerUnmarshalDocument(context.toBuilder()
+                                .writer(writer)
+                                .build());
+                    },
+                    () -> {
+                        protocolGenerator.generateProtocolDocumentUnmarshalerMarshalDocument(context.toBuilder()
+                                .writer(writer)
+                                .build());
+                    });
+
+            Symbol documentInterfaceSymbol = getInternalDocumentSymbol(DOCUMENT_INTERFACE_NAME);
+
+            writer.openBlock("func $L(v interface{}) $T {", "}", INTERNAL_NEW_DOCUMENT_MARSHALER_FUNC,
+                    documentInterfaceSymbol, () -> {
+                        protocolGenerator.generateNewDocumentMarshaler(context.toBuilder()
+                                .writer(writer)
+                                .build(), marshalerSymbol);
+                    }).write("");
+
+            writer.openBlock("func $L(v interface{}) $T {", "}", INTERNAL_NEW_DOCUMENT_UNMARSHALER_FUNC,
+                    documentInterfaceSymbol, () -> {
+                        protocolGenerator.generateNewDocumentUnmarshaler(context.toBuilder()
+                                .writer(writer)
+                                .build(), unmarshalerSymbol);
+                    }).write("");
+
+            writer.openBlock("func $T(v Interface) (ok bool) {", "}", isDocumentInterface, () -> {
+                writer.openBlock("defer func() {", "}()", () -> {
+                    writer.openBlock("if err := recover(); err != nil {", "}", () -> writer.write("ok = false"));
+                });
+                writer.write("v.$L()", IS_SMITHY_DOCUMENT_METHOD);
+                writer.write("return true");
+            }).write("");
+        });
+    }
+
+    private void writeInternalDocumentImplementation(
+            GoWriter writer,
+            Symbol typeSymbol,
+            Runnable unmarshalMethodDefinition,
+            Runnable marshalMethodDefinition
+    ) {
+        writer.openBlock("type $T struct {", "}", typeSymbol, () -> {
+            writer.write("value interface{}");
+        });
+        writer.write("");
+
+        writer.openBlock("func (m $P) $L(v interface{}) error {", "}", typeSymbol, UNMARSHAL_SMITHY_DOCUMENT_METHOD,
+                unmarshalMethodDefinition);
+        writer.write("");
+
+        writer.openBlock("func (m $P) $L() ([]byte, error) {", "}", typeSymbol, MARSHAL_SMITHY_DOCUMENT_METHOD,
+                marshalMethodDefinition);
+        writer.write("");
+
+        writer.write("func (m $P) $L() {}", typeSymbol, IS_SMITHY_DOCUMENT_METHOD);
+        writer.write("");
+
+        writer.write("var _ $T = ($P)(nil)", getInternalDocumentSymbol(DOCUMENT_INTERFACE_NAME, true), typeSymbol);
+        writer.write("");
+    }
+
+    private void writeDocumentPackage(String fileName, Consumer<GoWriter> writerConsumer) {
+        delegator.useFileWriter(getDocumentFilePath(fileName), getDocumentPackage(), writerConsumer);
+    }
+
+    private void writeInternalDocumentPackage(String fileName, Consumer<GoWriter> writerConsumer) {
+        delegator.useFileWriter(getInternalDocumentFilePath(fileName), getInternalDocumentPackage(), writerConsumer);
+    }
+
+    private String getInternalDocumentPackage() {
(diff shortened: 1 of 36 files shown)


document/json/encoder.go
```diff
@@ -0,0 +1,318 @@
...
+	// Handle both pointers and interface conversion into types
+	rv = serde.ValueElem(rv)
+
+	switch rv.Kind() {
+	case reflect.Invalid:
+		if tag.OmitEmpty {
+			return nil
+		}
+		vp.GetValue().Null()
+		return nil
```
jasdel, line 84: would this case be hit if the `IsZeroValue` check above is triggered?

document/json/encoder.go
```diff
@@ -0,0 +1,318 @@
...
+
+	default:
+		return e.encodeScalar(vp, rv)
+	}
+}
+
+func (e *Encoder) encodeZeroValue(vp valueProvider, rv reflect.Value) error {
+	switch rv.Kind() {
+	case reflect.Invalid:
+		vp.GetValue().Null()
```
jasdel, line 107: I think this duplicates the `Invalid` case in `encode`. the case in `encode` probably isn't needed?

document/json/encoder_test.go
```diff
@@ -0,0 +1,82 @@
...
+	})
+}
+
+func TestNewEncoderUnsupportedTypes(t *testing.T) {
+	type customTime time.Time
+
+	cases := []interface{}{
+		time.Now().UTC(),
+		customTime(time.Now().UTC()),
+	}
```
jasdel, line 48: Should this include cases for functions and channels?

document/json/json.go
```diff
@@ -0,0 +1,25 @@
+package json
+
+func NewEncoder(optFns ...func(options *EncoderOptions)) *Encoder {
```
jasdel, line 3: nit documentation

document/json/decoder_test.go
```diff
@@ -0,0 +1,630 @@
...
+		t.Fatalf("expect error, got nil")
+	}
+}
+
+func TestNewDecoderUnsupportedTypes(t *testing.T) {
+	cases := []struct {
+		input             []byte
+		value             interface{}
+		disableJSONNumber bool
+	}{
```
jasdel, line 566: Would be helpful to also have a test case that the `Decoder` cannot unmarshal into a non-empty interface.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -395,6 +407,107 @@ public Config build() {
...
+        public Void nullNode(NullNode node) {
+            writer.writeInline("nil");
+            return null;
+        }
+
+        @Override
+        public Void numberNode(NumberNode node) {
+            if (node.isNaturalNumber()) {
+                writer.writeInline("$L", node.getValue());
+            } else {
```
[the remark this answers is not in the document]
skmcgrail: The else branch here covers Floats in the case that it was a BigDecimal type  according to codegen. Will need to investigate adding big.Int for the natural number case here.

document/document.go
```diff
@@ -0,0 +1,71 @@
...
+var _ noSerde = (*NoSerde)(nil)
+
+// IsNoSerde returns whether the given type implements the no smithy document serde interface.
+func IsNoSerde(x interface{}) bool {
+	_, ok := x.(noSerde)
+	return ok
+}
+
+// Number is a arbitrary precision numerical value
+type Number string
```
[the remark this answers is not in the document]
skmcgrail: Yup I can add helpers for converting to those types.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -395,6 +407,107 @@ public Config build() {
...
+                } else if (value instanceof Double) {
+                    writer.writeInline("float64($L)", value.doubleValue(), value);
+                } else {
+                    writer.addUseImports(SmithyGoDependency.BIG);
+                    writer.writeInline("func () *big.Float {\n"
+                            + "\tf, ok := (&big.Float{}).SetString($S)\n"
+                            + "\tif !ok { panic(\"failed to parse string to float: \" + $S) }\n"
+                            + "\treturn f\n"
+                            + "}()", value, value);
+                }
```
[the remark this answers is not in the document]
skmcgrail: They do not currently, can raise that with the Smithy team, overall I think document types is lacking in sufficient protocol tests.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
```diff
@@ -0,0 +1,373 @@
...
+            }).write("");
+
+            writer.write("var _ $T = ($P)(nil)",
+                    serviceSmithyDocumentInterface, internalDocumentInterface);
+            writer.write("var _ $T = ($P)(nil)",
+                    smithyDocumentMarshaler, internalDocumentInterface);
+            writer.write("var _ $T = ($P)(nil)",
+                    smithyDocumentUnmarshaler,
+                    internalDocumentInterface);
+            writer.write("");
```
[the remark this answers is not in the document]
skmcgrail: Technically no harm in it being here but can move to a separate file.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ProtocolGenerator.java
```diff
@@ -252,90 +258,171 @@ static String getDeserializeMiddlewareName(ShapeId operationShapeId, ServiceShap
...
+     * @param context   the generation context
      * @param operation the operation shape to retrieve errors for
      * @return map of error names to {@link ShapeId}
      */
     default Map<String, ShapeId> getOperationErrors(GenerationContext context, OperationShape operation) {
         return HttpProtocolGeneratorUtils.getOperationErrors(context, operation);
     }
 
+    default void generateProtocolDocumentMarshalerUnmarshalDocument(GenerationContext context) {
+        throw new CodegenException("document types not implemented for protocol");
```
mtdowling, line 270: Can these all say which protocol?

mtdowling: And can they get docs?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ProtocolGenerator.java
```diff
@@ -252,90 +258,171 @@ static String getDeserializeMiddlewareName(ShapeId operationShapeId, ServiceShap
...
+
+    default void generateProtocolDocumentUnmarshalerUnmarshalDocument(GenerationContext context) {
+        throw new CodegenException("document types not implemented for protocol");
+    }
+
+    default void generateProtocolDocumentUnmarshalerMarshalDocument(GenerationContext context) {
+        throw new CodegenException("document types not implemented for protocol");
+    }
+
+    default void generateNewDocumentMarshaler(GenerationContext context, Symbol marshalerSymbol) {
```
mtdowling, line 285: How are these different than the above? I guess docs would help here too a lot.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ProtocolGenerator.java
```diff
@@ -252,90 +258,171 @@ static String getDeserializeMiddlewareName(ShapeId operationShapeId, ServiceShap
...
+        private final List<GoIntegration> integrations;
+        private final String protocolName;
+        private final GoDelegator delegator;
+
+        private GenerationContext(Builder builder) {
+            this.settings = SmithyBuilder.requiredState("settings", builder.settings);
+            this.model = SmithyBuilder.requiredState("model", builder.model);
+            this.service = SmithyBuilder.requiredState("service", builder.service);
+            this.symbolProvider = SmithyBuilder.requiredState("symbolProvider", builder.symbolProvider);
+            this.writer = builder.writer;
```
mtdowling, line 317: Why is the writer optional?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
```diff
@@ -0,0 +1,373 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mtdowling, line 2: nit: 2021 on all the files

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
```diff
@@ -0,0 +1,373 @@
...
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.go.codegen.integration.ProtocolGenerator;
+import software.amazon.smithy.go.codegen.integration.ProtocolGenerator.GenerationContext;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.DocumentShape;
+
+/**
+ * Generates the Smithy document type for the service client.
+ */
+public final class ProtocolDocumentGenerator {
```
mtdowling, line 31: Does this class need to be public?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
```diff
@@ -0,0 +1,373 @@
...
+import java.util.function.Consumer;
+
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.go.codegen.integration.ProtocolGenerator;
+import software.amazon.smithy.go.codegen.integration.ProtocolGenerator.GenerationContext;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.DocumentShape;
+
+/**
+ * Generates the Smithy document type for the service client.
```
mtdowling, line 29: Can you give more info on what this generates? Do you generate a document type per type defined in the model? Do you generate a single document type used by every document type, and this mostly just generates the serde code specific to the protocol?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
```diff
@@ -0,0 +1,373 @@
...
+    public boolean hasDocumentShapes() {
+        return getDocumentShapes().size() > 0;
+    }
+
+    /**
+     * Add a document shape to the generator.
+     *
+     * @param shape the document shape to add.
+     */
+    public void addDocumentShape(DocumentShape shape) {
```
mtdowling, line 83: Isn't the set of document shapes knowable in the constructor? Something like:

```
context.getModel.getDocumentShapes();
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
```diff
@@ -0,0 +1,373 @@
...
+        delegator.useShapeWriter(settings.getService(model), writer -> {
+            writer.write("type $L = $T", NO_DOCUMENT_SERDE_TYPE_NAME, noSerde);
+        });
+
+        delegator.useFileWriter("./types/types.go", settings.getModuleName() + "/types", writer -> {
+            writer.write("type $L = $T", NO_DOCUMENT_SERDE_TYPE_NAME, noSerde);
+        });
+    }
+
+    private void generateInternalDocumentInterface() {
```
mtdowling, line 128: Does each service have its own document interface?

document/json/encoder.go
```diff
@@ -0,0 +1,318 @@
...
+	"reflect"
+
+	"github.com/aws/smithy-go/document"
+	"github.com/aws/smithy-go/document/internal/serde"
+	smithyjson "github.com/aws/smithy-go/encoding/json"
+)
+
+type EncoderOptions struct{}
+
+type Encoder struct {
```
JordonPhillips, line 15: Why is this necessary? You appear to be using the standard library json marshaler anyway.

document/json/decoder.go
```diff
@@ -0,0 +1,333 @@
...
+	"math/big"
+	"reflect"
+
+	"github.com/aws/smithy-go/document"
+	"github.com/aws/smithy-go/document/internal/serde"
+)
+
+type DecoderOptions struct{}
+
+type Decoder struct {
```
JordonPhillips, line 15: Why is this needed as opposed to the default json unmarshaler?

JordonPhillips (changes requested):

document/internal/serde/field.go
```diff
@@ -0,0 +1,292 @@
+package serde
```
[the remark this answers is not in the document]
skmcgrail: The only reasonable way I can think of would require us to just make this a publicly accessible Go package in smithy-go and we would have to maintain the public interfaces and forward compatability. It's certainly something we could do, and it would likely probably best done as a stand-alone module. Would we want to do that now or later as a separate task?

document/internal/serde/reflect.go
```diff
@@ -0,0 +1,22 @@
...
+	BigFloat             reflect.Type
+	BigInt               reflect.Type
+	DocumentNumber       reflect.Type
+	MapStringToInterface reflect.Type
+	Time                 reflect.Type
+}{
+	BigFloat:             reflect.TypeOf((*big.Float)(nil)).Elem(),
+	BigInt:               reflect.TypeOf((*big.Int)(nil)).Elem(),
+	DocumentNumber:       reflect.TypeOf((*document.Number)(nil)).Elem(),
+	MapStringToInterface: reflect.TypeOf((map[string]interface{})(nil)),
```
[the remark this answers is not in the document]
skmcgrail: Not need in this instance given how the decoder / encoder is implemented.

document/internal/serde/serde.go
```diff
@@ -0,0 +1,97 @@
...
+	if !v.IsValid() {
+		return nil
+	}
+	if v.Kind() == reflect.Ptr {
+		return PtrToValue(v.Interface())
+	}
+	return v.Interface()
+}
+
+func IsZeroValue(v reflect.Value) bool {
```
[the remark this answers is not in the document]
skmcgrail: This is here to detect if a value is a zero-value, this doesn't prevent serialization of the zero value unless the user instructed the document marshaler via a struct tag like`document:",omitempty"`, where omitempty indicates that zero-values should not be serialized.

document/json/decoder.go
skmcgrail: While document types are JSON-like we want to decouple them from the standard library's expectations and behaviors around the types it supports. (For example cases where Smithy has chosen not to support like Time serialization and blob serialization). By explicitly defining our own Encoder/Decoder interface and struct field tags we will be to evolve the interfaces further when modeled document types are added at a later date. Additionally, this enables us to have tighter control over type usage, such as being able to block usage of generated Smithy types as their document representation could change at a later date when modeled type support is added.

document/json/decoder.go
```diff
@@ -0,0 +1,333 @@
...
+	case reflect.Struct:
+		if rv.CanInterface() && document.IsNoSerde(rv.Interface()) {
+			return &document.UnmarshalTypeError{
+				Value: fmt.Sprintf("unsupported type"),
+				Type:  rv.Type(),
+			}
+		}
+	case reflect.Interface:
+		rv.Set(reflect.MakeMap(serde.ReflectTypeOf.MapStringToInterface))
+		rv = rv.Elem()
```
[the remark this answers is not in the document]
skmcgrail: Yes this should :)

document/json/encoder.go
skmcgrail: Going to resolve this comment, see earlier response on the Decoder for the teams reasoning.

document/json/encoder.go
skmcgrail: Since `serde.ValueElem(rv)` unwraps pointers and interfaces to retrieve the element the resulting value in `rv` could still be Invalid.

document/json/encoder.go
skmcgrail: See my reply to your earlier comment with regards to how Invalid is valid in both these cases.

document/json/encoder_test.go
skmcgrail: The current behavior for channels and functions is to completely skip them and not serialize them (with no error returned), which is the behavior of the `encoding/json` library. I feel like it makes sense to at least attempt to match that behavior for those types. Time is a bit of a special case since we want to avoid making any sort of opinionated stance on their handling at this time, whereas channels and functions would more then likely never be supported.

document/document.go
```diff
@@ -0,0 +1,71 @@
+package document
+
+import (
+	"strconv"
+)
+
+type SmithyDocumentMarshaler interface {
```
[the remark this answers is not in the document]
skmcgrail: Yes, next revision will include documentation, just wanted to get this out for initial feedback on the overall implementation.

document/document.go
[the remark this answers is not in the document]
skmcgrail: This is used in the Encoder/Decoder. This is a similar concept to the standard library's json.Number in that a user can use this type to capture a number literal into a string and then use one of the function methods to coerce to your desired type (int64, float64, etc).

document/errors.go
```diff
@@ -0,0 +1,74 @@
...
+// Error returns the string representation of the error satisfying the error
+// interface.
+func (e *UnmarshalError) Error() string {
+	return fmt.Sprintf("unmarshal failed, cannot unmarshal %q into %s, %v",
+		e.Value, e.Type.String(), e.Err)
+}
+
+// An InvalidMarshalError is an error type representing an error
+// occurring when marshaling a Go value type to an AttributeValue.
+type InvalidMarshalError struct {
```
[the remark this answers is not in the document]
skmcgrail: There isn't any cases on the Marshal code path where we wrap and underlying error currently which is why this type doesn't exist specifically.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/DocumentShapeDeserVisitor.java
```diff
@@ -411,7 +411,7 @@ protected final void generateDeserFunction(
             BiConsumer<GenerationContext, Shape> functionBody
     ) {
         SymbolProvider symbolProvider = context.getSymbolProvider();
-        GoWriter writer = context.getWriter();
+        GoWriter writer = context.getWriter().get();
```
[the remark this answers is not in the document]
skmcgrail: The ProtocolGenerator context is kind of a bit of a grab bag currently, in some instances a writer is to be expected, in other instances the delegator that is stored on the context is used and no writer is expected. This was just to make that contact a bit clearer that it may not be there. 

This was just something that I noticed when I was refactoring context to make it immutable after construction.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ProtocolGenerator.java
skmcgrail: I refer to this in an earlier comment, but since the generation context is a bit of a grab bag, the writer is not required in all instances of the context's usage, thus is why it is optional.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ProtocolUtils.java
```diff
@@ -210,4 +210,8 @@ public static void writeSerDelegateFunction(
 
         lambda.accept(acceptVar);
     }
+
+    public static void writeDocumentInterfaceCheck(GenerationContext context, GoWriter writer) {
```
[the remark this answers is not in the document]
skmcgrail: This is an older artifact and didn't get cleaned up, will remove in next revision.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
skmcgrail: Yes, these exported constants are needed so that protocol implementation parts make symbol references to the types.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
skmcgrail: Will add some clarifying code examples sections to this class documentation. To summarize what is being created here: Each service client generates its own document type interface definition. This ensures that document types that are sent or received from the service are bound to the client package and it's associated client protocol. This prevents the situation where a user recieves

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
```diff
@@ -0,0 +1,373 @@
...
+            writer.openBlock("func $L(v interface{}) $T {", "}", NEW_LAZY_DOCUMENT,
+                            getDocumentSymbol(DOCUMENT_INTERFACE_NAME), () -> {
+                                writer.write("return $T(v)",
+                                        getInternalDocumentSymbol(INTERNAL_NEW_DOCUMENT_MARSHALER_FUNC));
+                            })
+                    .write("");
+        });
+    }
+
+    private void generateNoSerdeType() {
```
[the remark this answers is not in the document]
skmcgrail: This type is actually needed in all service client structures types to avoid making backwards incompatible changes when modeled smithy document types are later added at a future date. Since Smithy makes no opinionated stance currently on things like JSON object keys should be formatted (lower-case, camelCase, UPPER CASE) etc, this NoSerde type is embedded in a generated smithy structure types from the model to prevent their usage as document types. That way if an existing structure is later annotated as modeled document type, we don't have a situation where we make a potential breaking change in the representation of the underlying type.  (object key casing being an example).

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
skmcgrail: explained in an earlier comment above, going to resolve this in favor of that thread.

skmcgrail:
I believe all outstanding feedback has been capture in the `Implement Feedback` commit that I have force-pushed. Please review when you have the chance, where applicable I left some conversations unresolved to give you the opportunity to read the response, feel free to resolve if satisfied.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
```diff
@@ -0,0 +1,520 @@
...
+            writer.write("type $L = $T", DOCUMENT_INTERFACE_NAME,
+                            getInternalDocumentSymbol(DOCUMENT_INTERFACE_NAME))
+                    .write("");
+
+            writer.writeDocs(String.format("You create document type using the %s function and passing it the Go "
+                    + "type to be marshaled and sent to the service. The document marshaler supports semantics similar "
+                    + "to the encoding/json Go standard library.", NEW_LAZY_DOCUMENT));
+            writer.writeDocs("");
+            writer.writeDocs("For more information see the accompanying package documentation and linked references.");
+            writer.openBlock("func $L(v interface{}) $T {", "}", NEW_LAZY_DOCUMENT,
```
jasdel, line 154: Should this document the struct tag to use for custom member names?

document/document.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+// Marshaler is an interface for a type that marshals a document to its protocol-specific byte representation and
+// returns the resulting bytes. A non-nil error will be returned if an error is encountered during marshaling.
+//
+// Marshal supports basic scalars (int,uint,float,bool,string), big.Int, and big.Float, maps, slices, and structs.
+// Anonymous nested types are flattened based on Go anonymous type visibility.
+//
+// When defining struct types. the `document` struct tag can be used to control how the value will be
+// marshaled into the resulting protocol document.
```
jasdel, line 16: Maybe this is where the earlier comment I had about linking to struct tags should link to? What do you think about linking to the smithy-go documentation within the code's docstring?

document/internal/serde/field.go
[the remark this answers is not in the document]
jasdel: Makes sense. That sounds like something we could add later as some kind of type walker helper if we decided its valuable in a maintainable way. We probably don't want to do that here.

jasdel (approved):

JordonPhillips (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolDocumentGenerator.java
skmcgrail: The doc.go for the clients `document` package has a `doc.go` from this template file https://github.com/aws/smithy-go/blob/a87ef15659d457c607421ee90635d35b2e6def4c/codegen/smithy-go-codegen/src/main/resources/software/amazon/smithy/go/codegen/document_doc.go.template which details more details and links to the smithy-go document package.

document/document.go
skmcgrail: https://github.com/aws/smithy-go/blob/a87ef15659d457c607421ee90635d35b2e6def4c/codegen/smithy-go-codegen/src/main/resources/software/amazon/smithy/go/codegen/document_doc.go.template links to this.

skmcgrail:
Rebased on `main`.

PR #319, opened 2021-07-21
Add method to get successful output from waiters

haines:
*Issue #, if available:*
https://github.com/aws/aws-sdk-go-v2/issues/1338

*Description of changes:*

This PR adds a `WaitForOutput` method to waiters that returns the successful operation's output. This allows the user to avoid making an extra API request to find the full result of the operation that they were waiting for - in my case, I want to access the `StandardOutputContent` on the `GetCommandInvocationOutput` using the `ssm.NewCommandExecutedWaiter`.

I opted for adding a method rather than changing the existing `Wait` method to avoid breaking the interface, but if a breaking change is ok then I'm happy to update this to just change the signature of `Wait` to include the output.

I've regenerated the SDK code locally and confirmed that the new method works as intended.


haines:
@jasdel @skmcgrail could you please take a look?

jasdel (approved):
Thanks for taking the time to create this PR. The change looks good.  We'll get this merged in and included in the next release of the smithy-go module. I've also created a AWS SDK for Go v2 PR that makes use of this change.

PR #336, opened 2021-10-28
Consume Response.Body in `errorCloseResponseBodyMiddleware.HandleDeserialize` and `closeResponseBody.HandleDeserialize`

rennnosuke:
# Description of changes
Consume response body to reuse TCP connection in default HTTP client's transport.

>
	// The http Client and Transport guarantee that Body is always
	// non-nil, even on responses without a body or responses with
	// a zero-length body. It is the caller's responsibility to
	// close Body. The default HTTP client's Transport may not
	// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
	// not read to completion and closed.
>
https://github.com/golang/go/blob/2ff1074f5ac48e67ee079dafc9f0e314bc454510/src/net/http/response.go#L59

# Related Issue([aws/aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2))
https://github.com/aws/aws-sdk-go-v2/issues/1317

### DIFF
--- a/transport/http/middleware_close_response_body.go
+++ b/transport/http/middleware_close_response_body.go
@@ -3,6 +3,7 @@ package http
 import (
 	"context"
 	"fmt"
+	"io"
 
 	"github.com/aws/smithy-go/middleware"
 )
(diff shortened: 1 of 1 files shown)


skmcgrail:
Hey @rennnosuke thanks for putting to this PR. This change is actually already present in https://github.com/aws/smithy-go/pull/331 and will be part of that feature shipment in an upcoming release.

PR #351, opened 2022-01-18
Update smithy-go to pin updated version of smithy

jasdel:
Updates smithy-go to pin to the smithy 1.17.0 version.

Updates the code generation to handle (de)serialization of quoted header list values.

### DIFF
--- a/transport/http/headerlist.go
+++ b/transport/http/headerlist.go
@@ -2,48 +2,120 @@ package http
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
+	"unicode"
 )
 
 func splitHeaderListValues(vs []string, splitFn func(string) ([]string, error)) ([]string, error) {
-	for i := 0; i < len(vs); i++ {
-		if len(vs[i]) == 0 {
-			continue
-		}
+	values := make([]string, 0, len(vs))
 
+	for i := 0; i < len(vs); i++ {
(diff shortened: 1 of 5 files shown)


JordonPhillips:
You may as well go all the way to 1.16 while you're at it

transport/http/headerlist.go
```diff
@@ -2,48 +2,120 @@ package http
...
 }
 
-func commaSplit(v string) ([]string, error) {
-	return strings.Split(v, ","), nil
+func quotedCommaSplit(v string) (parts []string, err error) {
+	v = strings.TrimSpace(v)
+
+	expectMore := true
+	for i := 0; i < len(v); i++ {
+		if unicode.IsSpace(rune(v[i])) {
```
skmcgrail, line 36: np: I think the cast to `rune` here is not required?  `v[i]` is already a rune when you index into a string.

skmcgrail (approved):

transport/http/headerlist.go
jasdel: This is one of those odd differences between `v[i]` iteration vs `range`.  https://go.dev/play/p/ZPPf0pkQ8iQ 

`range` of string will return a `int` and `rune`(aka `int32`), but  a string index `v[i]`  will return `int` and `byte`(aka `uint8`)

PR #360, opened 2022-03-17
codegen: Update REST streaming request payload content-type usage

jasdel:
Updates the SDK's codegen for REST HTTP request payload to always
include content-type header for operations with streaming request
payloads. Prior the content-type header would only be added if the
streaming request io.Reader was not nil.


jasdel:
This cannot be merged until https://github.com/aws/aws-sdk-go-v2/pull/1630 failing test cases with content-type are fixed.

PR #362, opened 2022-04-11
Add support for Smithy httpBearerAuth authentication trait

jasdel:
Adds support for the [Smithy httpBearerAuth](https://awslabs.github.io/smithy/1.0/spec/core/auth-traits.html#httpbearerauth-trait) authentication trait to smithy-go. This allows the SDK to support the bearer authentication flow for API operations decorated with `httpBearerAuth`. An API client will need to be provided with its own `bearer.TokenProvider` implementation or use the `bearer.StaticTokenProvider` implementation.

### DIFF
--- a/auth/bearer/token.go
+++ b/auth/bearer/token.go
@@ -19,7 +19,7 @@ func (t Token) Expired(now time.Time) bool {
 	if !t.CanExpire {
 		return false
 	}
-	return now.Equal(t.Expires) || now.After(t.Expires)
+	return now.Round(0).Equal(t.Expires) || now.After(t.Expires)
 }
 
 // TokenProvider provides interface for retrieving bearer tokens.
--- a/auth/bearer/token.go
+++ b/auth/bearer/token.go
@@ -1,0 +1,49 @@
+package bearer
+
+import (
+	"context"
+	"time"
+)
+
+// Token provides a type wrapping a bearer token and expiration metadata.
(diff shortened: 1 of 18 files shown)


auth/bearer/token.go
```diff
@@ -0,0 +1,49 @@
...
+	Expires   time.Time
+}
+
+// IsExpired returns if the token's Expires time is before or equal to the time
+// provided. If CanExpires is false, Expired will always return false.
+func (t Token) Expired(now time.Time) bool {
+	if !t.CanExpire {
+		return false
+	}
+	return now.Round(0).Equal(t.Expires) || now.After(t.Expires)
```
skmcgrail, line 22: Should we do `now = now.Round(0)` before the return? That way `Equal` and `After` are both wall-clock checks instead of monotonic.

auth/bearer/token_cache.go
```diff
@@ -0,0 +1,221 @@
...
+	context.Context
+}
+
+func (s *suppressedContext) Deadline() (deadline time.Time, ok bool) {
+	return time.Time{}, false
+}
+
+func (s *suppressedContext) Done() <-chan struct{} { return nil }
+
+func (s *suppressedContext) Err() error { return nil }
```
skmcgrail, line 221: Don't believe this is used anymore.

skmcgrail (approved):

Issue #326, opened 2021-09-07
Include `json` tag in structs if field is annotated with @jsonName


jabalsad:
As title says, in its current form `json` tags are not included in the generated Go code.

We use `jsonName` extensively throughout our Smithy models, but the Go code is unusable without the `json` struct tags.

If you can point me towards where it needs to be added, I can look at submitting a pull request.

Many thanks.

jasdel:
Thanks for reaching out @jabalsad. The Shape type generators are generally protocol agnostic. This allows API clients to not be tightly coupled to an API protocol at its surface public interface. With that said, [StructureGenerator](https://github.com/aws/smithy-go/blob/main/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/StructureGenerator.java) would be a good place to investigate adding a builder so that the `StructGenerator` can be configured to optionally generate Go struct tags on struct members.

lucix-aws:
Tracked by #458 - the json family of protocol generators should handle this.

PR #372, opened 2022-07-28
Add jsonName and json struct tag support

NukaCody:
*Issue #326, if available:*

*Description of changes:*

If you specify `jsonName("somevalue")` trait in your smithy model, then assume you want a json struct tag.

I think that's the most intuitive approach to take this but definitely open to other ideas. Something like you have to specify  in `smithy-build.json` some value such as `useJsonStructTag: true`. But may be more work for not much reward.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/StructureGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/StructureGenerator.java
@@ -25,6 +25,7 @@ import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.ErrorTrait;
+import software.amazon.smithy.model.traits.JsonNameTrait;
 import software.amazon.smithy.model.traits.StreamingTrait;
 import software.amazon.smithy.utils.MapUtils;
 import software.amazon.smithy.utils.SetUtils;
@@ -113,7 +114,17 @@ final class StructureGenerator implements Runnable {
                                 .orElse(memberSymbol);
                     }
 
-                    writer.write("$L $P", memberName, memberSymbol);
+                    String jsonTag = "";
+                    if (member.hasTrait(JsonNameTrait.ID)) {
+                        jsonTag = member.getMemberTrait(model, JsonNameTrait.class).get().getValue();
+                    }
+
+                    if (!jsonTag.isEmpty()) {
+                        writer.write("$L $P `json:\"$L\"`", memberName, memberSymbol, jsonTag);
+                    } else {
+                        writer.write("$L $P", memberName, memberSymbol);
+                    }
+
                 });
 
         runnable.run();
--- a/codegen/smithy-go-codegen-test/model/main.smithy
+++ b/codegen/smithy-go-codegen-test/model/main.smithy
@@ -159,6 +159,7 @@ structure GetCityInput {
     // has to be marked as required.
     @required
     @httpLabel
+    @jsonName("city-id")
     cityId: CityId,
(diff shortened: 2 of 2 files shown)


NukaCody:
Also, I have no clue what the letters in `writer.write("$L $P \json:\"$L\"", memberName, memberSymbol, jsonTag);` mean lol. I tried $C and kept running into a `InvocationTargetException` so I went with L and it worked

jasdel:
We've been hesitant to add struct tags to the API types. Mainly because the SDK strives to ensure serialization in an implementation detail of the API and SDK. An API could always change it's protocol and that would put the struct tags at risk of breaking. The public types purposely do not expose these.

We're happy to continue this discussion and hear your use case about the need for this feature.  Also it's important to note that API protocol serialization isn't Go JSON standard especially in cases such as byte slices and time.

NukaCody:
I can understand that. If I capture it right, looking through an example. [EC2RunInstances](https://github.com/aws/aws-sdk-go-v2/blob/8df2ee085105c481fc1f39086d023b22d457245d/service/ec2/api_op_RunInstances.go#L71) and it's [Input](https://github.com/aws/aws-sdk-go-v2/blob/8df2ee085105c481fc1f39086d023b22d457245d/service/ec2/api_op_RunInstances.go#L86) make no assumption on protocol. It's not till [invokeOperation](https://github.com/aws/aws-sdk-go-v2/blob/8df2ee085105c481fc1f39086d023b22d457245d/service/ec2/api_client.go#L172) and I think more specifically [HTTP semantics start here](https://sourcegraph.com/github.com/aws/smithy-go/-/blob/transport/http/request.go?L26)

Our use case is a little bit different. We have a folder structure like this:

```bash
.
├── model
│   ├── build/smithyprojections/model/source/go-codegen
└── externalLib # Both use model
└── infra # Both use model
```

Our model is used in two different contexts, one as supplying an actual API in the traditional sense of an SDK talking to an API endpoint in `Infra`. Another as a sort of "shared type" folder. In `externalLib` we have to build a custom marshaller/unmarshaller that transforms a specification (that we can't control) into cloudformation. The jsonName tag makes writing the custom marshaller easier.

I think jsonName() is also a two-way door. If we need to backtrack, we can always `Find and Replace jsonName(*) to ""` then regenerate the model. Although I do see how that would break `externalLib`'s marshalling if we did that. I think that's a risk we are willing to accept though.

I'm a bit smooth brain and didn't follow this: 

>  Also it's important to note that API protocol serialization isn't Go JSON standard especially in cases such as byte slices and time.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/StructureGenerator.java
```diff
@@ -113,7 +114,17 @@ public void renderStructure(Runnable runnable, boolean isInputStructure) {
...
+                    if (member.hasTrait(JsonNameTrait.ID)) {
+                        jsonTag = member.getMemberTrait(model, JsonNameTrait.class).get().getValue();
+                    }
+
+                    if (!jsonTag.isEmpty()) {
+                        writer.write("$L $P `json:\"$L\"`", memberName, memberSymbol, jsonTag);
+                    } else {
+                        writer.write("$L $P", memberName, memberSymbol);
+                    }
+
```
jvschneid, line 127: Consider using optionals here instead of all this conditionals, something like:

```
writer.writeInline("$L $P", memberName, memberSymbol);

member.getMemberTrait(model, JsonNameTrait.class).map(JsonNameTrait::getValue)
      .map(jsonName -> writer.writeInline(" `json:\"$L\"`", jsonName));
                        
writer.write("");
```

jvschneid (changes requested):

lucix-aws:
I'm going to close this - this should be handled via protocol generation, specifically we'll get json protocol support (including `jsonName`) by addressing #458.

PR #375, opened 2022-08-08
Dynamo putitem escaping

RanVaknin:
*Issue #, if available:*
https://github.com/aws/aws-sdk-go-v2/issues/1764

*Description of changes:*
- fixed escaping on `writeKey` function. 
- Added test coverage for that
- Fixed typo
- Added test coverage for `escapeStringBytes`

### DIFF
--- a/encoding/json/escape_test.go
+++ b/encoding/json/escape_test.go
@@ -23,16 +23,16 @@ func TestEscapeStringBytes(t *testing.T) {
 			input:    []byte(`hello\\world`),
 		},
 		"new line": {
-			expected: `"foo\\nbar"`,
-			input:    []byte(`foo\nbar`),
+			expected: `"foo\nbar"`,
+			input:    []byte("foo\nbar"),
 		},
 		"carriage return": {
-			expected: `"foo\\rbar"`,
-			input:    []byte(`foo\rbar`),
+			expected: `"foo\rbar"`,
+			input:    []byte("foo\rbar"),
 		},
 		"tab": {
-			expected: `"foo\\tbar"`,
-			input:    []byte(`foo\tbar`),
+			expected: `"foo\tbar"`,
+			input:    []byte("foo\tbar"),
 		},
 	}
 	for name, c := range cases {
--- a/encoding/json/escape_test.go
+++ b/encoding/json/escape_test.go
@@ -6,16 +6,44 @@ import (
 )
 
 func TestEscapeStringBytes(t *testing.T) {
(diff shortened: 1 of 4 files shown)


encoding/json/escape_test.go
```diff
@@ -0,0 +1,49 @@
...
+			expected: `"foo\""`,
+			input:    []byte(`foo"`),
+		},
+		"double escape": {
+			expected: `"hello\\\\world"`,
+			input:    []byte(`hello\\world`),
+		},
+		"new line": {
+			expected: `"foo\\nbar"`,
+			input:    []byte(`foo\nbar`),
```
skmcgrail, line 27: This test isn't actually testing what you are are expecting. This is only testing whether `\` gets escaped as `\\`. The issue is you are using [raw string literals](https://tip.golang.org/ref/spec#String_literals) when constructing your byte slices. So you aren't actually testing that newline is getting escaped in this case. You need to be using an actual string literal to your byte slice `[]byte("foo\nbar")` which should output the expected value of `"foo\nbar"`

encoding/json/escape_test.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+		"carriage return": {
+			expected: `"foo\\rbar"`,
+			input:    []byte(`foo\rbar`),
```
skmcgrail, line 31: See comment w.r.t. new line testing.

encoding/json/escape_test.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+		"tab": {
+			expected: `"foo\\tbar"`,
+			input:    []byte(`foo\tbar`),
```
skmcgrail, line 35: See comment w.r.t. new line testing.

skmcgrail (changes requested):
See comments with regards to some of the test cases in your `TestEscapeStringBytes` test.

encoding/json/escape_test.go
RanVaknin: fixed

encoding/json/escape_test.go
RanVaknin: fixed

encoding/json/escape_test.go
RanVaknin: fixed

skmcgrail:
Did you validate this with the AWS SDK to confirm that this correctly resolved the issue seen with the DynamoDB client?

skmcgrail (approved):

PR #381, opened 2022-08-30
codegen: Updates codegen to support non-model writing, and lazy writable

jasdel:
Updates the smithy-go codegen to include support for writing file manifests for non-model based Go code generation. This allows code generation for standalone Go modules that are not smithy model based. Including Go module unit tests.

Adds GoWritable for lazy code writing. Allows a Writable to be returned from a method improving the ability to do inline code generation.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
@@ -60,6 +60,8 @@ public final class GoWriter extends AbstractCodeWriter<GoWriter> {
     private static final Logger LOGGER = Logger.getLogger(GoWriter.class.getName());
     private static final int DEFAULT_DOC_WRAP_LENGTH = 80;
 
+    private static final Pattern ARGUMENT_NAME_PATTERN = Pattern.compile("\\$([a-z][a-zA-Z_0-9]+)(:\\w)?");
+
     private final String fullPackageName;
     private final ImportDeclarations imports = new ImportDeclarations();
     private final List<SymbolDependency> dependencies = new ArrayList<>();
@@ -263,8 +265,7 @@ public final class GoWriter extends AbstractCodeWriter<GoWriter> {
     }
 
     private void validateContext(String template, Map<String, Object> scope) {
-        var pattern = Pattern.compile("\\$([a-z][a-zA-Z_0-9]+)(:\\w)?");
-        var matcher = pattern.matcher(template);
+        var matcher = ARGUMENT_NAME_PATTERN.matcher(template);
 
         var foundKeys = new HashSet<String>();
         while (matcher.find()) {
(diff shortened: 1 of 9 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
```diff
@@ -50,31 +55,236 @@
...
+        }
+        scope.forEach((k, v) -> {
+            if (v == null) {
+                throw new CodegenException("Template argument " + k + " cannot be null");
+            }
+        });
+    }
+
+    private void validateContext(String template, Map<String, Object> scope) {
+        var pattern = Pattern.compile("\\$([a-z][a-zA-Z_0-9]+)(:\\w)?");
```
skmcgrail, line 266: I'm not sure if Java would optimize this by default via static analysis, but can probably move this to be a `private final static` class field?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ManifestWriter.java
```diff
@@ -117,25 +140,64 @@ private static Node buildManifestFile(
...
+    }
+
+    public static Builder builder() {
+        return new Builder();
     }
 
+    public static class Builder implements SmithyBuilder<ManifestWriter> {
+        private String moduleName;
+        private FileManifest fileManifest;
+        private List<SymbolDependency> dependencies;
```
skmcgrail, line 169: Using Smithy's `BuilderRef` for these types can be helpful. https://github.com/skmcgrail/smithy/blob/rules-engine/smithy-rules-engine/src/main/java/software/amazon/smithy/rulesengine/traits/EndpointTestsTrait.java#L93

Ensures the list is always copied from the builder, is initialized to an empty list on copy if it was never set etc.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/testutils/ExecuteCommand.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.logging.Logger;
+import software.amazon.smithy.utils.SmithyBuilder;
+
+/**
+ * Utility for invoking command line utilities.
+ */
+public final class ExecuteCommand {
+    private static final Logger LOGGER = Logger.getLogger(ExecuteCommand.class.getName());
+
+    private final String[] command;
```
skmcgrail, line 34: Not sure if we have it setup, but usually `String[]` will be a spotbugs item. Usually want this to be a `List<String>`.

skmcgrail (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/testutils/ExecuteCommand.java
jasdel: Interesting, Spotbugs doesn't call this out as an issue. Would it be better for this to be a List instead?  Its a array since that is the input the underlying exec method takes.

skmcgrail (approved):

PR #383, opened 2022-08-31
Add support for intEnum code generation

syall:
*Issue #, if available:*

N/A

---

*Description of changes:*

Add support for intEnum code generation:

- Update smithy-go-codegen-test with intEnum test
- Add unit test for intEnum shape

---

*Testing:*

Ran `make smithy-clean smithy-build smithy-publish-local smithy-clean`.

---


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -327,4 +328,12 @@ public Void serviceShape(ServiceShape shape) {
         });
         return null;
     }
+
+    @Override
+    public Void intEnumShape(IntEnumShape shape) {
+        if (!shape.getEnumValues().isEmpty()) {
```
sugmanue, line 334: This condition is always `true`, we do validate that enums have at least one member.

codegen/smithy-go-codegen/src/main/resources/software/amazon/smithy/go/codegen/smithy-tests/int-enum-shape-test/expected/types/enums.go
```diff
@@ -0,0 +1,66 @@
+// Code generated by smithy-go-codegen DO NOT EDIT.
```
jasdel, line 1: Not relevant to this PR specifically, but the `codegen` directory should be broken off into an empty go module so any go files nested under codegen are out of scope of go commands executed on the smithy-go root module.

jasdel (approved):

PR #387, opened 2022-09-08
Endpoint Types

skmcgrail:


transport/fieldset.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// NewFieldSet returns an initalized FieldSet.
+func NewFieldSet() *FieldSet {
+	return &FieldSet{
+		header: &Fields{},
+	}
+}
+
+// $get retrieves the header Field by name, by performing a case-insensitive lookup.
```
jasdel, line 17: Looks like minor codegen bug here with `$get` not rendering correct name for doc strings in this file.

transport/internal/fieldset/main.go
```diff
@@ -0,0 +1,102 @@
...
+		{{ (privateSymbol $name) }}: &Fields{},
+	{{ end -}}
+	}
+}
+
+{{ range $name, $_ := $.Locations -}}
+{{ $privateName := (privateSymbol $name) -}}
+{{ $publicName := (publicSymbol $name) -}}
+{{ $get := (join "Get" $publicName) -}}
+// $get retrieves the {{ $name }} Field by name, by performing a case-insensitive lookup.
```
jasdel, line 51: `$get` should be wrapped in `{{ }}`

transport/internal/fieldset/main.go
```diff
@@ -0,0 +1,102 @@
...
+func (fs *FieldSet) {{ $get }}(name string) Field {
+	return fs.{{ $privateName }}.Get(name)
+}
+
+{{ $has := (join "Has" $publicName) -}}
+// $has returns whether a matching {{ $name }} Field exists for the given name using case-insensitive lookup.
```
jasdel, line 57: Similar to `$get` for other doc strings

transport/properties.go
```diff
@@ -0,0 +1,59 @@
...
+// Properties provides storing and reading metadata values. Keys may be any
+// comparable value type. Get and set will panic if key is not a comparable
+// value type.
+//
+// Properties uses lazy initialization, and Set method must be called as an
+// addressable value, or pointer. Not doing so may cause key/value pair to not
+// be set.
+type Properties struct {
+	values map[interface{}]interface{}
+}
```
jasdel, line 12: This file in `transport` isn't needed since the `Properties` are defined in the root package now?

jasdel (changes requested):

jasdel (approved):

PR #394, opened 2022-10-21
[issues-391] Fix document typecheck in encoding  for nested types

isaiahvita:
This resolves github issue 391 where encoding of document types is not checking to see if nested types are explicitly marked for no serialization or deserialization. In order to to check for every layer of nesting, I moved the noserde check to the recursive layer from the original top-level function. This mimics how it correctly works in decoding

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/document/json/encoder.go
+++ b/document/json/encoder.go
@@ -20,10 +20,6 @@ type Encoder struct {
 
 // Encode returns the JSON encoding of v.
 func (e *Encoder) Encode(v interface{}) ([]byte, error) {
-	if document.IsNoSerde(v) {
-		return nil, fmt.Errorf("unsupported type: %v", v)
-	}
-
 	encoder := smithyjson.NewEncoder()
 
 	if err := e.encode(jsonValueProvider(encoder.Value), reflect.ValueOf(v), serde.Tag{}); err != nil {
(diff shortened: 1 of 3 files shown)


document/json/encoder.go
```diff
@@ -324,3 +324,10 @@ func isValidJSONNumber(s string) bool {
 	// Make sure we are at the end.
 	return s == ""
 }
+
+func (e *Encoder) unsupportedType(v interface{}) error {
+	if document.IsNoSerde(v) {
+		return &document.InvalidUnmarshalError{Type: reflect.TypeOf(v)}
```
isaiahvita, line 330: is this backwards incompatible? should i be performing the same error behavior as L24?

document/json/encoder.go
```diff
@@ -20,10 +20,6 @@ type Encoder struct {
 
 // Encode returns the JSON encoding of v.
 func (e *Encoder) Encode(v interface{}) ([]byte, error) {
-	if document.IsNoSerde(v) {
-		return nil, fmt.Errorf("unsupported type: %v", v)
-	}
-
```
skmcgrail, line 26: Wondering if this check should stay? Otherwise I think lines 71-78 in the `encode` function could still serialize the zero value of a generated Smithy type as an empty object which would be weird?

isaiahvita: in my previous implementation, the `document.IsNoSerde` func was still being called, i had just wrapped it in another function `isUnsupported`. however, in my latest commit, I replaced that to more closely mimic decode which actually directly calls `document.IsNoSerde` when decoding a struct

skmcgrail (approved):

PR #400, opened 2022-11-30
add smithy-ci gradle task

AndrewFossAWS:
*Description of changes:*

> In the past, Smithy changes have been released without integration testing with Smithy-based SDK code generators. In addition, SDK code generators often fall behind in Smithy versions. As a result, we don’t have visibility on code generators’ failures until they are upgraded to the latest smithy version. 

> Upgrading versions and building all the code generators manually is also a labor-intensive task. Hence, the goal is to design a continuous integration system for testing the latest Smithy implementation with the SDK code generators.

This PR adds  `smithy-ci` gradle task to be executed by Smithy integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Release (2022-12-02)
+
+* No change notes available for this release.
+
 # Release (2022-10-24)
 
 ## Module Highlights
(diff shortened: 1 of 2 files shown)


syall (approved):
Approving, equivalent to `make smithy-clean smithy-build smithy-publish-local`.

PR #401, opened 2022-12-07
Add `ErrorCodeOverride` to error structs

eddy-aws:
*Description of changes:*
Makes changes to AwsJson protocol generator.
* Introduces `ErrorCodeOverride` field to error structs
* Returns `ErrorCodeOverride` if not empty. Else returns default error code
* Introduces `generateErrorShape` method to allow override default error block for unmodeled errors

For example, AWS code bases would have something like the following to override `generateErrorShape`:

```java
    @Override
    protected Set<StructureShape> generateErrorShapes(
        GenerationContext context, OperationShape operation, Symbol responseType) {
        if (isAwsQueryCompatibleTraitFound(context)) {
            return HttpProtocolGeneratorUtils.generateErrorDispatcher(
                context, operation, responseType, this::writeErrorMessageCodeDeserializer,
                this::getOperationErrors, (writer) -> AwsJsonRpc1_0.awsQueryCompatibleDefaultBlockWriter(writer));
        } else {
            return HttpProtocolGeneratorUtils.generateErrorDispatcher(
                context, operation, responseType, this::writeErrorMessageCodeDeserializer,
                this::getOperationErrors);
        }
    }

    private static void awsQueryCompatibleDefaultBlockWriter(GoWriter writer) {
        writer.openBlock("default:", "", () -> {
            writer.openBlock("genericError := &smithy.GenericAPIError{", "}", () -> {
                    writer.write("Code: getAwsQueryErrorCode(response),");
                writer.write("Message: errorMessage,");
            });
            writer.write("return genericError");
        });
    }
```

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/StructureGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/StructureGenerator.java
@@ -38,7 +38,7 @@ final class StructureGenerator implements Runnable {
             "ErrorMessage", "string",
             "ErrorFault", "string"
     );
-    private static final Set<String> ERROR_MESSAGE_MEMBER_NAMES = SetUtils.of("ErrorMessage", "Message");
+    private static final Set<String> ERROR_MEMBER_NAMES = SetUtils.of("ErrorMessage", "Message", "ErrorCodeOverride");
 
     private final Model model;
     private final SymbolProvider symbolProvider;
@@ -144,12 +144,12 @@ final class StructureGenerator implements Runnable {
             // The message is the only part of the standard APIError interface that isn't known ahead of time.
             // Message is a pointer mostly for the sake of consistency.
             writer.write("Message *string").write("");
-
+            writer.write("ErrorCodeOverride *string").write("");
 
             for (MemberShape member : shape.getAllMembers().values()) {
                 String memberName = symbolProvider.toMemberName(member);
                 // error messages are represented under Message for consistency
-                if (!ERROR_MESSAGE_MEMBER_NAMES.contains(memberName)) {
+                if (!ERROR_MEMBER_NAMES.contains(memberName)) {
                     writer.write("$L $P", memberName, symbolProvider.toSymbol(member));
                 }
             }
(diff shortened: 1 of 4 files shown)


syall:
There should not be any AWS-specific code in generic smithy code generators.

isaiahvita:
agree with @syall, this logic should be in the codegen module in the V2 package. only generic service (not AWS specific) code generation logic should be in `smithy-go`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/StructureGenerator.java
```diff
@@ -174,7 +180,13 @@ private void renderErrorStructure() {
 
         String errorCode = protocolGenerator == null ? shape.getId().getName(service)
                 : protocolGenerator.getErrorCode(service, shape);
-        writer.write("func (e *$L) ErrorCode() string { return $S }", structureSymbol.getName(), errorCode);
+        if (withQueryCompatible) {
+            renderQueryCompatibleErrorCodeStructure(writer, structureSymbol, errorCode);
+            writer.write(
+                "func (e *$L) SetErrorCode(code string) { e.QueryErrorCode = code }", structureSymbol.getName());
+        } else {
+            writer.write("func (e *$L) ErrorCode() string { return $S }", structureSymbol.getName(), errorCode);
```
isaiahvita, line 188: instead of the AWS check here, i think we should modify `ErrorCode()` to check if there is a `Code` field and if it is non-empty then use it. otherwise do the existing default behavior.

then in the V2 package, if the aws query compatible trait is included then add the `SetErrorCode` function

eddy-aws:
> just to clarify the intent behind:
> 
> ```
> Introduces generateErrorShape method to allow override default error block for unmodeled errors
> ```

I have updated the description with some additional details

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/StructureGenerator.java
```diff
@@ -39,6 +39,7 @@ final class StructureGenerator implements Runnable {
             "ErrorFault", "string"
     );
     private static final Set<String> ERROR_MESSAGE_MEMBER_NAMES = SetUtils.of("ErrorMessage", "Message");
+    private static final Set<String> ERROR_CODE_MEMBER_NAMES = SetUtils.of("Code");
```
syall, line 42: Can these be merged into a single `ERROR_MEMBER_NAMES` set?

syall (changes requested):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SymbolVisitor.java
```diff
@@ -122,6 +122,7 @@ final class SymbolVisitor implements SymbolProvider, ShapeVisitor<Symbol> {
                 .put("ErrorFault", "ErrorFault_")
                 .put("Unwrap", "Unwrap_")
                 .put("Error", "Error_")
+                .put("ErrorCodeOverride", "ErrorCodeOverride_")
```
syall, line 125: Added `ErrorCodeOverride` to reserved error members.

syall (approved):
Rebased changes, but will need to Squash commits before merging.

isaiahvita (approved):

PR #404, opened 2022-12-23
feat(auth): support http api key auth trait

eduardomourar:
*Issue #, if available:* N/A

*Description of changes:*

This adds codegen support for services and operations that use the `@httpApiKeyAuth trait`.

It will include a new auth middleware to deal with API Keys. The generated client will have that middleware automatically included. This value will be injected into the header/query parameter that's specified in the trait by the new middleware.

Similar implementation as done in the TypeScript codegen (more details https://github.com/awslabs/smithy-typescript/pull/473 and https://github.com/aws/aws-sdk-js-v3/pull/4012).


syall:
Hi @eduardomourar! From first glance at the CI actions:

- Moving the `Message` interface from `auth/bearer` to `auth` [is a backwards-incompatible API change](https://github.com/aws/smithy-go/actions/runs/3769309288/jobs/6442574207#step:5:17).
    - Is it possible to implement the feature with `Message` in the same package?
- There are [some checkstyle validations caught, but can be fixed](https://github.com/aws/smithy-go/actions/runs/3769309290/jobs/6442574311#step:6:53).
    - Checkstyle checks can be ran locally by running `./gradlew check` in the `codegen/` directory.

eduardomourar:
> Hi @eduardomourar! From first glance at the CI actions:
> 
> * Moving the `Message` interface from `auth/bearer` to `auth` [is a backwards-incompatible API change](https://github.com/aws/smithy-go/actions/runs/3769309288/jobs/6442574207#step:5:17).
>   
>   * Is it possible to implement the feature with `Message` in the same package?
> * There are [some checkstyle validations caught, but can be fixed](https://github.com/aws/smithy-go/actions/runs/3769309290/jobs/6442574311#step:6:53).
>   
>   * Checkstyle checks can be ran locally by running `./gradlew check` in the `codegen/` directory.

I have pushed changes taking into account both suggestions.

auth/auth.go
```diff
@@ -0,0 +1,27 @@
...
+	// The current auth configuration that has been set by any auth middleware and
+	// that will prevent from being set more than once.
+	CURRENT_AUTH_CONFIG key = "currentAuthConfig"
+)
+
+// An HTTP-specific authentication scheme that sends an arbitrary
+// auth value in a header or query string parameter.
+// As described in the Smithy documentation:
+// https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
+type HttpAuthDefinition struct {
```
gosar, line 15: this seems to be `HttpApiKeyAuthDefinition`, no? I see the same name used in the corresponding TypeScript PR which is merged in.

auth/auth.go
```diff
@@ -0,0 +1,27 @@
+package auth
+
+type key string
+
+const (
+	// The current auth configuration that has been set by any auth middleware and
+	// that will prevent from being set more than once.
+	CURRENT_AUTH_CONFIG key = "currentAuthConfig"
```
gosar, line 8: Is this a CURRENT_AUTH_CONFIG or CURRENT_API_KEY_AUTH_CONFIG?
Is this actually going to be set by _any_ auth middleware? sigv4 middleware?

auth/auth.go
eduardomourar: This is a generic auth definition used for both API Key and [Bearer token](https://smithy.io/1.0/spec/core/auth-traits.html#httpbearerauth-trait) (ideally it should be used for for [httpBasicAuth](https://smithy.io/1.0/spec/core/auth-traits.html#httpbasicauth-trait) and [httpDigestAuth](https://smithy.io/1.0/spec/core/auth-traits.html#httpdigestauth-trait) traits as well).

auth/auth.go
eduardomourar: This auth config should be used any auth middleware to prevent us from setting the auth more than once (e.g. you would send Bearer token in the Authorization header and the API key in the query parameter in the same request). I think sigv4 middleware would not require us setting this config, because it is a special case. For it, I believe we would only allow one auth middleware (for a particular operation) in the codegen itself. I don't believe we can design an operation in the Smithy model that can allow sigv4 with any other auth mechanism, right?

codegen/smithy-go-codegen-test/model/main.smithy
```diff
@@ -7,8 +7,10 @@ use smithy.waiters#waitable
 
 /// Provides weather forecasts.
 @httpBearerAuth
+@httpApiKeyAuth(scheme: "Apikey", name: "Authorization", in: "header")
 @fakeProtocol
 @paginated(inputToken: "nextToken", outputToken: "nextToken", pageSize: "pageSize")
+@auth([httpBearerAuth, httpApiKeyAuth])
```
syall, line 13: general note: `@auth` order is not implemented in smithy-go yet. When testing locally, changing the order as shown produced the same order of auth middleware resolution as defined in `AddAuthPlugin.java`:

```smithy
@auth([httpApiKeyAuth, httpBearerAuth])
```

auth/auth.go
```diff
@@ -0,0 +1,27 @@
+package auth
```
syall, line 1: Add a note about the package being unstable and subject to breaking changes, whether it's here or in a `auth/docs.go` file.

syall (changes requested):
Hi @eduardomourar, sorry for the late review!

I added some comments to the current state of the PR, most notably saying that the `@httpApiKeyAuth` trait support in the PR is experimental and unstable, subject to breaking changes.

auth/auth.go
```diff
@@ -0,0 +1,27 @@
...
+
+type key string
```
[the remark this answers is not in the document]
eduardomourar: actually had to create own type because of this static check error:
> should not use built-in type string as key for value; define your own type to avoid collisions (SA1029) go-staticcheck

PR #405, opened 2022-12-24
feat(codegen): create go.mod file automatically

eduardomourar:
*Issue #, if available:* N/A

*Description of changes:*

In order to simplify the management of the generated Go files and well as allowing some customization, this will introduce the following changes:

- A property called `moduleVersion` in the codegen setttings will allow consumers to pass the version that they manage user their own tooling outside Smithy
- The file `go.mod` will automatically be generated whenever a settings called `generateGoMod` is set to true ~~based on the template located at `codegen/smithy-go-codegen/src/main/resources/software/amazon/smithy/go/codegen/go.mod.template`~~
- ~~The file `go_module_metadata.go` will generated containing the version of the module provided by the user~~


syall:
Hi @eduardomourar, thanks for the contribution!

I was wondering if it would be possible to provide an example of how this feature would work, and how this change was tested?

eduardomourar:
At Stedi, we are starting to generate the Go client SDKs based on our own Smithy models. In order to use our tooling for managing the modules' version, it made more sense to create, at least, the go.mod as part of the codegen. We have tested this internally with all services, but we are not ready to make this repository public just yet. I will try to either create a sample repo or maybe include some basic test to this PR.

RanVaknin:
Hi @eduardomourar 

We really appreciate your contirbution! Right now no one is able to review your PR. We will try to address is as part of 2023 efforts.

Thanks again!
Ran

syall:
Hey @eduardomourar! I had a few suggestions for this feature:

1. Toggle `go.mod` generation through a `generateGoMod` boolean in `GoSettings.java` (defaulting to `false`) to avoid overwriting existing consumers that have their own `go.mod` workflows.
2. Use `go` commands to generate `go.mod` files (reference #283, where `go.mod` generation was first removed from `smithy-go`).
3. For `go.mod`, use a `goDirective` string in `GoSettings.java` (defaulting to `1.15`) instead of `DEFAULT_MINIMUM_GO_VERSION` in the PR (see [the go module docs](https://go.dev/ref/mod#go-mod-file-go) for reference).
4. For the `go_module_metadata.go` generation, this is an opinionated file that probably should not be included in the smithy-go repo. If it were included, it should not be in the smithy-go plugins SPI list, but instead used at a smithy-go consumer level (e.g. [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/smithy-aws-go-codegen/src/main/resources/META-INF/services/software.amazon.smithy.go.codegen.integration.GoIntegration)).

syall:
The `go.mod` unit test probably failed since `go.mod` isn't written into the smithy file manifest, but is instead written by `go` commands.

To assert on `go.mod` contents and creation, the actual path of `go.mod` would have to be referenced.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoSettings.java
```diff
@@ -33,10 +34,16 @@ public final class GoSettings {
...
+    private static final String MODULE_VERSION = "moduleVersion";
+    private static final String GENERATE_GO_MOD = "generateGoMod";
+    private static final String GO_DIRECTIVE = "goDirective";
 
     private ShapeId service;
     private String moduleName;
     private String moduleDescription = "";
+    private String moduleVersion;
+    private Boolean generateGoMod = false;
+    private String goDirective = "1.15";
```
syall, line 46: Could the values for these be extracted into static fields, e.g. for `goDirective`:

```java
private static final String DEFAULT_GO_DIRECTIVE = "1.15";
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoSettings.java
```diff
@@ -47,13 +54,16 @@ public final class GoSettings {
...
+            Arrays.asList(SERVICE, MODULE_NAME, MODULE_DESCRIPTION, MODULE_VERSION, GENERATE_GO_MOD, GO_DIRECTIVE));
 
         settings.setService(config.expectStringMember(SERVICE).expectShapeId());
         settings.setModuleName(config.expectStringMember(MODULE_NAME).getValue());
         settings.setModuleDescription(config.getStringMemberOrDefault(
                 MODULE_DESCRIPTION, settings.getModuleName() + " client"));
-
+        settings.setModuleVersion(config.getStringMemberOrDefault(MODULE_VERSION, null));
+        settings.setGenerateGoMod(config.getBooleanMemberOrDefault(GENERATE_GO_MOD, false));
+        settings.setGoDirective(config.getStringMemberOrDefault(GO_DIRECTIVE, "1.15"));
```
syall, line 66: These settings could reference static fields instead of hardcoding / duplicating the default values.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -255,6 +255,9 @@ void execute() {
         List<SymbolDependency> dependencies = writers.getDependencies();
         writers.flushWriters();
 
+        LOGGER.fine("Generating go.mod file");
```
syall, line 258: This log should be moved to `GoModGenerator.java` after the `generateGoMod` check.

syall (changes requested):
Left a few comments, but overall looks good.

---

The `go.mod` output in `smithy-go-codegen-test` gradle subproject looks correct at path `codegen/smithy-go-codegen-test/build/smithyprojections/smithy-go-codegen-test/source/go-codegen/go.mod`:

```
module github.com/aws/smithy-go/internal/tests/service/weather

go 1.18

require (
        github.com/aws/smithy-go v1.4.0
        github.com/jmespath/go-jmespath v0.4.0
)
```

syall (approved):

aajtodd (approved):

PR #411, opened 2023-01-28
upgrade smithy to 1.27.2

syall:
*Issue #, if available:*

N/A

---

*Description of changes:*

Upgrade smithy to 1.27.2.

Tested with `make smithy-clean smithy-build smithy-publish-local`.

---

### DIFF
--- a/codegen/gradle.properties
+++ b/codegen/gradle.properties
@@ -1,2 +1,2 @@
-smithyVersion=1.26.0
+smithyVersion=1.27.2
 smithyGradleVersion=0.6.0


codegen/gradle.properties
```diff
@@ -1,2 +1,2 @@
-smithyVersion=1.26.0
+smithyVersion=[1.27.0,1.28.0[
```
isaiahvita, line 1: is this valid gradle syntax? with the hanging open bracket?

isaiahvita: discussed offline, despite the latter way being more readable/explicit IMO, the former way written in the PR is more gradle-idiomatic. and i agree for most cases its better to follow convention. so im approving.

isaiahvita (dismissed):

codegen/gradle.properties
aajtodd: Why a range in the first place? Why not a specific version here? We control the smithy version and when we upgrade, I'm not sure I'm understanding why we are declaring a range of versions to be compatible with.

syall: Updated to `1.27.2`

lucix-aws (approved):

PR #415, opened 2023-03-17
Add standard lib functions with utility error construct

isaiahvita:
This PR adds generic standard library functions for endpoints2.0.

* An error/endpoints package was added in order to contain a utility ErrorCollector class that can provide detailed logging for errors encountered in rule evaluation. It was made public so that the additional SDK functions can depend on this as well.
* The standard library functions were added in an internal package as required by the EP20 SEP.


isaiahvita:
wrong PR again

PR #388, opened 2022-09-08
Adds code generation of Smithy modeled endpoint metadata

jasdel:
*Issue #, if available:*

*Description of changes:*

Adds support for generating endpoint metadata from API model.

TODO:
* [x] Parameters can have default values, need to assign before validation
* [x] Generate modeled endpoint tests
* [x] Implement remaining builtin functions
* [x] Parameter option docstrings are not getting generated correct
* [x] Modeled test endpoint generator duplicates options of regular endpoint generation. This should be simplified.
* [ ] Better logging of ErrorCollector
* [ ] Endpoint properties should be use typed keys not string
    * [ ] `AuthScheme` key should should private to the API client module
    * [ ] Consider any new `Properties` key add to set of generated struct in endpoint package.
* [x] Document endpoint/builtin types
* [x] ~`builtin.SubString` is defined as returning an Optional value, but the `valid-rules/substring.json` and s3 rules use the result as value not optional.~
* [x] Option<T> results when used in conditionals are will always cause conditional statements to be false. Not clear if this applies to other use cases as well.
* [ ] Some rules include duplicate return statements
```
service/s3/internal/endpoints/smithy_endpoints.go:687:6: unreachable code
service/s3/internal/endpoints/smithy_endpoints.go:3304:4: unreachable code
```
* [ ] `test_case_12/TestEndpointCase7` expects `0x7F`, non-printable character `DEL`, to be valid for `SubString`. This seems odd. Need to clarify behavior.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #416, opened 2023-03-17
Add standard lib functions

isaiahvita:
This PR adds generic standard library functions for endpoints2.0.

An error/endpoints package was added in order to contain a utility ErrorCollector class that can provide detailed logging for errors encountered in rule evaluation. It was made public so that the additional SDK functions can depend on this as well.
The standard library functions were added in an internal package as required by the EP20 SEP.

EDIT: the core functions `isSet` `not` and `getAttr` will be implemented directly in codegen generation in a forthcoming PR. for example,

```
        @Override
        public GoWriter.Writable visitIsSet(Expr expr) {
            return (GoWriter w) -> {
                w.write("$W != nil", new ExprGenerator(scope).generate(expr));
            };
        }

        @Override
        public GoWriter.Writable visitNot(Expr expr) {
            return (GoWriter w) -> {
                w.write("!($W)", new ExprGenerator(scope).generate(expr));
            };
        }

        @Override
        public GoWriter.Writable visitBoolEquals(Expr left, Expr right) {
            return (GoWriter w) -> {
                var generator = new ExprGenerator(scope);
                w.write("$W == $W", generator.generate(left), generator.generate(right));
            };
        }

        @Override
        public GoWriter.Writable visitStringEquals(Expr left, Expr right) {
            return (GoWriter w) -> {
                var generator = new ExprGenerator(scope);
                w.write("$W == $W", generator.generate(left), generator.generate(right));
            };
        }

        @Override
        public GoWriter.Writable visitGetAttr(GetAttr getAttr) {
            var target = new ExprGenerator(scope).generate(getAttr.getTarget());
            var path = (GoWriter.Writable) (GoWriter w) -> {
                getAttr.getPath().stream().toList().forEach((part) -> {
                    if (part instanceof GetAttr.Part.Key) {
                        w.writeInline(".$L", getBuiltinMemberName(((GetAttr.Part.Key) part).key()));
                    } else if (part instanceof GetAttr.Part.Index) {
                        w.writeInline(".Get($L)", ((GetAttr.Part.Index) part).index());
                    }
                });
            };

            return (GoWriter w) -> w.writeInline("$W$W", target, path);
        }

        @Override
        public GoWriter.Writable visitLibraryFunction(FunctionDefinition fnDef, List<Expr> args) {
            return new FnGenerator(scope).generate(fnDef, args);
        }
```


internal/endpoints/rulesfn/strings_test.go
```diff
@@ -0,0 +1,71 @@
+package rulesfn
+
+import (
+	"strings"
+	"testing"
+	smithyerrep "github.com/aws/smithy-go/error/endpoints"
+	"github.com/aws/smithy-go/ptr"
+)
+
+func TestSubStrin(t *testing.T) {
```
aajtodd, line 10: **fix**: `TestSubStrin` -> `TestSubString`

internal/endpoints/rulesfn/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package rulesfn provides endpoint rule functions for evaluating endpoint
+// resolution rules.
+
+package rulesfn
```
aajtodd, line 4: **Correctness**: Not sure where to place this comment but having rule functions in an `internal` package won't work. These functions will need to be referenced eventually by generated service clients which means they need to be accessible outside of the `smithy-go` package.

error/endpoints/error_collector.go
```diff
@@ -0,0 +1,49 @@
+package endpoints
+
+import "strings"
+
+// ErrorCollector provides collection of errors that occurred while evaluating
+// endpoint rule functions.
+type ErrorCollector struct {
```
aajtodd, line 7: I'm not sure I'm understanding the need for this. Why would we not just return an error from any given rules function? Everytime we add an error to a collector in a rules function we bail and stop processing. This seems unnecessary and not idiomatic.

internal/endpoints/rulesfn/uri.go
```diff
@@ -0,0 +1,132 @@
...
+		Path:           u.Path,
+		NormalizedPath: normalizedPath,
+		IsIp:           isIP,
+	}
+}
+
+// URL provides the structure describing the parts of a parsed URL returned by
+// [ParseURL].
+type URL struct {
+	Scheme         string
```
aajtodd, line 102: **fix**: Missing member documentation

aajtodd (changes requested):
Overall this looks really good. I have some concerns/questions around `ErrorCollector` and `internal` as well as some minor feedback.

One other question: I'm assuming that the other functions `isSet`, `not`, and `getAttr` will be implemented directly in codegen generation (i.e. inlined in codegen)?

error/endpoints/error_collector.go
lucix-aws: That aside, I don't think `ErrorCollector` exists for any reason other than for us to check specific failure cases in testing. I'm not saying we should outright remove it, though.

lucix-aws (approved):
Approved contigent on addressing the `internal/` issue as pointed out by Aaron.

internal/endpoints/rulesfn/doc.go
isaiahvita: ah right, ok will make this public. that actually might work out fine with a top-level `endpoints` package and i can put the `error_collector.go` file in there (rather than introducing a top-level `errors` package)

error/endpoints/error_collector.go
isaiahvita: My understanding is that the purpose of the `ErrorCollector` is for more robust error logging and more robust test cases. For example, `isValidHostLabel`, the SEP states that this function is required to return a `bool`. While bailing on `false` would be acceptable (as far as the SEP is concerned), the caller would get no further context as to why this thing failed (at the level in which we call `isValidHostLabel`). the `ErrorCollector` provides the caller (either the actual consumer or a test case) with context as to why it failed.
Additionally, while I agree that passing a separate object for error reporting purposes may not be idiomatic, this is due to requirements imposed on us by the EP20 SEP. And, I think consumers may accept a hit on language-idiom compliance in order to get better error reporting.
And while, the `ErrorCollector` may not be required to write unit tests, without it, we wouldnt be able to ensure that a particular branch was tested (i.e. an input failed for a particular reason and not just failed for whatever reason).

isaiahvita: With that said, if necessary, this can totally be revisited as more of the implementation gets built out -- since this is an incremental merge into a feature branch.

isaiahvita:
> One other question: I'm assuming that the other functions isSet, not, and getAttr will be implemented directly in codegen generation (i.e. inlined in codegen)?

@aajtodd Good question, yah thats correct and i should have said this in the PR description.

private/README.md
```diff
@@ -0,0 +1,2 @@
+## Smithy Go Private packages ##
+`private` is a collection of packages used internally by Smithy Go, and is subject to have breaking changes. This package is not `internal` so that if you really need to use its functionality, and understand breaking changes will be made, you are able to.
```
lucix-aws, line 2: I think just the first sentence is fine.

lucix-aws (approved):

private/README.md
aajtodd: > I think just the first sentence is fine.

Ya, I would either drop the rest or if you feel to need to say more I'd change it to something like `This package is not `internal` because it is an implementation detail of generated code.`

private/endpoints/rulesfn/error_collector.go
```diff
@@ -0,0 +1,49 @@
+package rulesfn
+
+import "strings"
+
+// ErrorCollector provides collection of errors that occurred while evaluating
+// endpoint rule functions.
+type ErrorCollector struct {
```
aajtodd, line 7: **Question**: I still don't understand who is the consumer of errors collected by this abstraction? How do we see this playing out in practice?

aajtodd (approved):

error/endpoints/error_collector.go
isaiahvita: removed error collector now per standup discussion, will revisit later if i start to feel strongly about it

private/README.md
isaiahvita: @lucix-aws sure, i can do that. i was just copying how both the v1 and v2 repository use a root level internal directory. but i dont have any objections with the way you suggested above.

@aajtodd sure, removed the requested lines.

private/endpoints/rulesfn/error_collector.go
isaiahvita: @aajtodd you can see a full example [here](https://github.com/aws/smithy-go/pull/388) if you click Example generated output in one of the comments. However, per standup discussion, im removing for now and will revisit later if i start to feel strongly about it.

PR #423, opened 2023-04-11
Add Go Synthetic Trait equality methods

syall:
*Issue #, if available:*

N/A.

*Description of changes:*

When using smithy diff or trait comparisons for the Go `Synthetic`
trait, the following exception is thrown in the `createNode()`
implementation:

```text
"attempted to serialize runtime only trait"
```

This is due to inheriting from `AbstractTrait`'s equality methods which
use the `ToNode()` method which calls the `createNode()` method.

This change adds an `equals()` and `hashCode()` implementation for the
Go `Synthetic` trait that does not use `ToNode()` or `createNode()`.

`ToNode()` cannot be overriden since `AbstractTrait`'s implementation is
marked as `final`.

*Testing:*

1. Running `make smithy-clean smithy-build smithy-publish-local smithy-clean` succeeded.
2. Running `make generate` in aws/aws-sdk-go-v2 produced no diffs.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/Synthetic.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/Synthetic.java
@@ -56,6 +56,25 @@ public final class Synthetic extends AbstractTrait implements ToSmithyBuilder<Sy
     }
 
     @Override
+    public boolean equals(Object other) {
+        if (other == this) {
+            return true;
+        } else if (!(other instanceof Synthetic)) {
+            return false;
+        } else {
+            Synthetic b = (Synthetic) other;
+            return toShapeId().equals(b.toShapeId()) && archetype.equals(b.getArchetype());
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        return toShapeId().hashCode() * 17 + Node.objectNode()
+            .withOptionalMember(ARCHETYPE, archetype.map(ShapeId::toString).map(Node::from))
+            .hashCode();
+    }
+
+    @Override
     public SmithyBuilder<Synthetic> toBuilder() {
         Builder builder = builder();
         getArchetype().ifPresent(builder::archetype);


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/Synthetic.java
```diff
@@ -55,6 +54,25 @@ protected Node createNode() {
         throw new CodegenException("attempted to serialize runtime only trait");
     }
 
+    @Override
+    public boolean equals(Object other) {
+        if (other == this) {
+            return true;
+        } else if (!(other instanceof Trait)) {
+            return false;
+        } else if (hashCode() != other.hashCode()) { // take advantage of hashcode caching
```
sugmanue, line 63: It does not look like the hash code is being cached below, and this should also test that the class is `Synthetic` instead of just `Trait`. Having said that and given that there's no really a good way to tell any instance apart I'd make them all equal or all different on its reference value (I suspect that the later will be preferred in particular when trying to figure out if this has been processed). E.g.,

Making each instance equivalent:

```
public boolean equal(Object other) {
   if (other == this) {
    return true;
  }
  return (other instanceof Synthetic);
}

public int hashCode() {
    return toShapeId().hashCode();
}
```
Making each instance different:

```
public boolean equal(Object other) {
   return other == this;
}

public int hashCode() {
    return System.identityHashCode(this);
}
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/Synthetic.java
```diff
@@ -55,6 +54,25 @@ protected Node createNode() {
...
+            return false;
+        } else {
+            Trait b = (Trait) other;
+            return toShapeId().equals(b.toShapeId());
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        return toShapeId().hashCode() * 17 + Node.objectNode().hashCode();
```
sugmanue, line 73: I would just use the hash code of the shape id, the other values are constant and thus won't add value to the final hash code.

sugmanue (approved):

isaiahvita (approved):

PR #424, opened 2023-04-27
Add code-generation for EndpointProvider (EndpointResolver)

isaiahvita:
- Renamed top level endpoints orchestrator from `SmithyEndpointGenerator.java` to `EndpointResolutionV2Generator.java`. I think this new name is more reflective of what the top-level endpoints2.0 generator class should be called
- Added an EndpointResolverGenerator.java class. This is the endpoints2.0 "Endpoint Provider". I named it "Resolver" instead of "Provider" because ultimately the action that it will be doing is `ResolveEndpoint` and that name cannot change since its legacy. And having both provider and resolve/resolver naming in the code seemed too confusing, so i aligned on one name.
- Moved the endpoints2.0 code into the ProtocolGenerator which allowed loading of library functions from downstream generators. This can make sense because it guarantees that only one endpoint ruleset engine is chosen and an endpoint can be thought of as coupled to a protocol.


transport/fields.go
```diff
@@ -0,0 +1,140 @@
...
+	return f.name
+}
+
+// Add appends the given values to the field.
+func (f Field) Add(values ...string) Field {
+	if len(values) == 0 {
+		return f
+	}
+	orig := f.values
+	f.values = make([]string, len(orig), len(orig)+len(values))
```
aajtodd, line 105: **fix**: This copy and the one in `Set` is unnecessary, just append to the original list

transport/fields.go
```diff
@@ -0,0 +1,140 @@
...
+// NewField constructs a new Field with the given name and values.
+func NewField(name string, values ...string) Field {
+	return Field{
+		name:   name,
+		values: values,
+	}
+}
+
+// Name returns the field name
+func (f Field) Name() string {
```
aajtodd, line 95: **fix**: All of these should be pointer receivers

transport/endpoint.go
```diff
@@ -0,0 +1,12 @@
+package transport
+
+import "github.com/aws/smithy-go"
+
+// Endpoint is a Smithy endpoint.
```
aajtodd, line 5: I think we can be a little more descriptive here

properties.go
```diff
@@ -0,0 +1,65 @@
...
+// key was not found.
+//
+// Panics if key type is not comparable.
+func (m Properties) Get(key interface{}) interface{} {
+	return m.values[key]
+}
+
+// Clone creates a shallow copy of Properties entries, returning a new Properties
+// value with the original entries copied into it.
+func (m Properties) Clone() Properties {
```
aajtodd, line 30: **fix**: Don't mix receiver types. Just use a pointer receiver for all these methods since we have methods that mutate.

transport/endpoint.go
```diff
@@ -0,0 +1,12 @@
...
+type Endpoint struct {
+	URI string
+
+	Fields *FieldSet
```
aajtodd, line 9: **Question**: Why is `Fields` as a type necessary? It seems to be representing headers, why should we not just use [net/http#Header](https://pkg.go.dev/net/http#Header)?

transport/endpoint.go
```diff
@@ -0,0 +1,12 @@
+package transport
+
+import "github.com/aws/smithy-go"
+
+// Endpoint is a Smithy endpoint.
+type Endpoint struct {
+	URI string
```
aajtodd, line 7: **Question**: Same question, why not just use [net/url#URL](https://pkg.go.dev/net/url#URL)?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/endpoints/EndpointResolverGenerator.java
```diff
@@ -0,0 +1,514 @@
...
+            scope = scope.withIdent(new Reference(ident, SourceLocation.NONE), conditionName);
+        } else {
+            conditionName = nameForExpression(fn);
+        }
+
+        if (fn.type() instanceof Type.Option || isConditionalFnResultOptional(condition, fn)) {
+            return goTemplate("""
+                    if exprVal := $target:W; exprVal != nil {
+                        $condName:L := *exprVal
+                        _ = $condName:L
```
aajtodd, line 251: **Question**: Why is this unused assignment necessary? Presumably the variable represented by `condName` is used somewhere so I'm curious why this is necessary?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/endpoints/EndpointResolverGenerator.java
```diff
@@ -0,0 +1,514 @@
...
+        var condition = conditions.get(0);
+        var remainingConditions = conditions.subList(1, conditions.size());
+
+        var generator = new ExpressionGenerator(scope, this.fnProvider);
+        var fn = conditionalFunc(condition);
+
+        String conditionName;
+        if (condition.getResult().isPresent()) {
+            var ident = condition.getResult().get();
+            conditionName = "_" + ident.asString();
```
aajtodd, line 238: **nit**: `conditionName` is slightly confusing here, maybe `conditionIdentifier` since we are creating a variable/identifier?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/endpoints/EndpointResolverGenerator.java
```diff
@@ -0,0 +1,514 @@
...
+     * @return name
+     */
+    private static String nameForRef(Reference ref) {
+        return "_" + ref.getName();
+    }
+
+    private GoWriter.Writable generateEndpoint(Endpoint endpoint, Scope scope) {
+        return goTemplate("""
+                $endpointType:T{
+                    URI: $url:W,
```
aajtodd, line 312: **Question**: 

This generates code like this:

```go
URI: func() string {
	var out strings.Builder
	out.WriteString("https://")
	out.WriteString(_Bucket)
	out.WriteString(".op-")
	out.WriteString(_outpostId)
	out.WriteString(".s3-outposts.")
	out.WriteString(_Region)
	out.WriteString(".")
	out.WriteString(_regionPartition.DnsSuffix)
	return out.String()
}(),
```

Is there not any way to generate this cleaner like:

```go
URI: strings.Join(
    []string {
        "https://",
        _Bucket,
        ".op-",
        _outpostId,
        ".s3-outposts.",
        _Region,
        ".",
        _regionPartition.DnsSuffix
    }
)
```

Overall minor point/question.

aajtodd (changes requested):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/endpoints/EndpointRulesEngineGenerator.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Generates all components required for Smithy Ruleset Endpoint Resolution.
+ * These components include a Provider, Parameters, and Tests.
+ */
+public class EndpointRulesEngineGenerator {
+
+    public static final String FEATURE_NAME = "RulesEngine";
+    public static final String PARAMETERS_TYPE_NAME = "EndpointParameters";
+    public static final String RESOLVER_INTERFACE_NAME = "Endpoint" + FEATURE_NAME + "Resolver";
```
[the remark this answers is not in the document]
isaiahvita: yah so the rationale behind this is the following:
-  `EndpointResolver` is already taken (it was the initial public type that exists currently)
- `EndpointProvider` i didnt use the term "provider" anywhere rather i used "resolver". This is because ultimately the existing Go interface we want to comply with has its function named `ResolveEndpoint` and my understanding is that "resolving an endpoint" and "providing an endpoint" is the same thing (correct me if im wrong). so it felt confusing to see both of these terms as unintentionally different ways of describing the same thing.
- i didnt see "rules engine" as a description of an implementation but rather the public facing name of "endpoints 2.0", this is because this name is used in all base smithy packages for endpoints 2.0 stuff.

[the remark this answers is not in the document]
isaiahvita: discussed offline, going to go with `V2` instead of `RulesEngine` because a customer may implement this themselves without rules and V2 signifies that this is an iteration and should be used over the previous.

transport/endpoint.go
isaiahvita: discussed offline, replaced URI string type with Go URL type. the rationale being its simpler to represent something that more closely resembles reality.
replaced `Fields` with Go net Headers. the rationale being the Go net Headers do the same thing without us having to maintain our own version.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/endpoints/EndpointResolverGenerator.java
isaiahvita: added a todo to reconsider this

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ProtocolGenerator.java
```diff
@@ -434,6 +434,15 @@ default String getErrorCode(ServiceShape service, StructureShape errorShape) {
     default void generateEventStreamComponents(GenerationContext context) {
     }
 
+
+    /**
+     * Generate all components necessary for Endpoint Rules Engine endpoint resolution.
+     *
+     * @param context the generation context.
+     */
+    default void generateEndpointRulesEngine(GenerationContext context) {
```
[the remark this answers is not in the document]
isaiahvita: removed "rules engine" from the implementation

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/endpoints/EndpointResolverGenerator.java
```diff
@@ -0,0 +1,529 @@
...
+
+                func $newResolverFn:T() *$resolverImplementationType:T {
+                    return &$resolverImplementationType:T{}
+                }
+
+                $resolveEndpointMethodDocs:W
+                func (r *$resolverImplementationType:T) $resolveEndpointMethodName:L(
+                    ctx $context:T, $paramArgName:L $parametersType:T,
+                ) (
+                    endpoint $endpointType:T, err error,
```
lucix-aws, line 123: Do we really get anything out of using the predeclared return style here?

endpoints/private/rulesfn/uri.go
```diff
@@ -65,17 +65,17 @@ func ParseURL(input string) *URL {
 	// be reverted since the returned URL will be used in string builders.
 	authority := strings.ReplaceAll(u.Host, "%", "%25")
 
-	var isIP bool
+	var isIp bool
```
lucix-aws, line 68: Why did we change this? I'm pretty sure go styling prefers the original state.

properties.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Properties provides storing and reading metadata values. Keys may be any
+// comparable value type. Get and set will panic if key is not a comparable
+// value type.
+//
+// Properties uses lazy initialization, and Set method must be called as an
+// addressable value, or pointer. Not doing so may cause key/value pair to not
+// be set.
+type Properties struct {
+	values map[interface{}]interface{}
```
lucix-aws, line 17: We may want to consider strongly typing these (I know the SEP says not to, but there's nothing wrong with doing so if we explicitly mark the types as not stable IMO), this would simplify consuming it in runtime down the road (you'd just pull a concrete instance of something out of the property bag rather than having to construct it from what is essentially a string -> void* map on the fly).

That's what we did in kotlin, and we actually also have an extensibility mechanism for a white-label SDK implementor to generate their own typed values.

I don't know that we necessarily _must_ to go this way in go, you could still do well with compartmentalized abstractions to deal with pulling out the raw map at runtime, but we should at least be aware of the option.

endpoints/private/rulesfn/uri.go
isaiahvita: its because the models only case the first letter of acronyms. the generator takes that casing by default, and there would have to be special logic that capitalizes all letters of acronyms. i couldnt readily find a utility in the Smithy utils that does this. and this seems like something we should use a library for and i wasnt sure if our org had any open-source policy on what 3P tools we can include in our repositories, do you know of any restrictions? all that said, i can make this work though if we need to.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/endpoints/EndpointResolverGenerator.java
isaiahvita: none that i am aware of other than not including the declaration of `endpoint` in the body, if thats even a benefit. tbh, i didnt think too hard about this and let it be since i wasnt aware of any actionable downside or benefit

properties.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Set stores the value pointed to by the key. If a value already exists at
+// that key it will be replaced with the new value.
+//
+// Set method must be called as an addressable value, or pointer. If Set is not
+// called as an addressable value or pointer, the key value pair being set may
+// be lost.
+//
+// Panics if the key type is not comparable.
+func (m *Properties) Set(key, value interface{}) {
```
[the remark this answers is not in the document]
isaiahvita: for my own understanding, can you explain why this is not preferred?

properties.go
isaiahvita: i see, did you have a particular type in mind? say, `map[string]string`?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -244,11 +244,11 @@ void execute() {
 
             writers.useFileWriter("endpoints.go", settings.getModuleName(), writer -> {
                 ProtocolGenerator.GenerationContext context = contextBuilder.writer(writer).build();
-                protocolGenerator.generateEndpointRulesEngine(context);
+                protocolGenerator.generateEndpointResolutionV2(context);
```
aajtodd, line 247: nit: You probably don't need to include the `v2` in the codegen component naming, it could simply be `generateEndpointResolver`

transport/endpoint.go
```diff
@@ -1,12 +1,17 @@
...
+// Endpoint is the endpoint object returned
+// by Endpoint resolution V2
 type Endpoint struct {
-	URI string
+	URI url.URL
 
-	Fields *FieldSet
+	Headers *http.Header
 
 	Properties smithy.Properties
```
aajtodd, line 16: fix: add documentation to each member and how it's used, I know it feels self evident but we should document all public types/members.

transport/endpoint.go
```diff
@@ -1,12 +1,17 @@
...
 
-// Endpoint is a Smithy endpoint.
+// Endpoint is the endpoint object returned
+// by Endpoint resolution V2
 type Endpoint struct {
-	URI string
+	URI url.URL
 
-	Fields *FieldSet
+	Headers *http.Header
```
aajtodd, line 14: **fix**: This can simply be `http.Header` (no pointer) since it's a map type

transport/endpoint.go
```diff
@@ -1,12 +1,17 @@
...
+import (
+	"net/http"
+	"net/url"
+	"github.com/aws/smithy-go"
+)
 
-// Endpoint is a Smithy endpoint.
+// Endpoint is the endpoint object returned
+// by Endpoint resolution V2
 type Endpoint struct {
```
aajtodd, line 11: **question**: We have the `endpoints` package should this not go in there? Or does that create issues with the package layout w.r.t `rulesfn`?

properties.go
aajtodd: I think this is fine it mirrors context API and we can't introduce generics anywhere so this looks ok to me for Go.

aajtodd (approved):

lucix-aws (approved):

properties.go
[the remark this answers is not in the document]
isaiahvita: discussed offline, the worry is just the asymmetrical api from the customers perspective. im going to change this to a value receiver, and have the `Set` method return the `Properties` object. so the customer will have to reassign the `Properties`. i think this feels more Go idiomatic

properties.go
isaiahvita: discussed offline with @lucix-aws @aajtodd . leaving this for now, and will prob add in strong typing at the accessor method level at the end

PR #428, opened 2023-06-07
Add deprecated option to config field resolvers

isaiahvita:
This utility may be necessary for consumers if they want to deprecated a With helper function that corresponds to a config field

### DIFF
--- a/middleware/metadata.go
+++ b/middleware/metadata.go
@@ -1,10 +1,10 @@
 package middleware
 
-import "github.com/aws/smithy-go"
-
 // MetadataReader provides an interface for reading metadata from the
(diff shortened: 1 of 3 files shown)


middleware/metadata.go
```diff
@@ -1,10 +1,10 @@
 package middleware
```
isaiahvita, line 1: for some reason, in previous ep20 commits, i modified this `middleware/metadata.go` file which shouldnt have been modified at all. these modifications resulted in breaking changes in middleware signatures. so in this commit im reverting it to its original state (prior to any ep20 commits)

aajtodd (approved):

PR #429, opened 2023-06-20
Modify user agent syntax

wty-Bryant:
Update user agent builder syntax while building user agent header string

### DIFF
--- a/.changelog/916abfa459a34a0a990762df94da11cf.json
+++ b/.changelog/916abfa459a34a0a990762df94da11cf.json
@@ -1,0 +1,8 @@
+{
+    "id": "916abfa4-59a3-4a0a-9907-62df94da11cf",
+    "type": "bugfix",
+    "description": "modify user agent builder syntax",
+    "modules": [
+        "."
+    ]
+}
--- a/transport/http/user_agent.go
+++ b/transport/http/user_agent.go
@@ -4,6 +4,11 @@ import (
 	"strings"
 )
 
+var validChars = map[rune]bool{
+	'!': true, '#': true, '$': true, '%': true, '&': true, '\'': true, '*': true, '+': true,
+	'-': true, '.': true, '^': true, '_': true, '`': true, '|': true, '~': true,
+}
+
 // UserAgentBuilder is a builder for a HTTP User-Agent string.
 type UserAgentBuilder struct {
 	sb strings.Builder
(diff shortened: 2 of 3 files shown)


aajtodd (changes requested):
This isn't correct. I think we'll need to handle the `#` on the AWS side [here](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/middleware/user_agent.go#L1). 

For example the [addProductName](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/middleware/user_agent.go#L103) function calls

```go
builder.AddKeyValue(aws.SDKName, aws.SDKVersion)
```

This would now be formatted as `aws-sdk-go-v2#1.2.3` but it should still be `aws-sdk-go/1.2.3`. I think we'll have to handle the logic [here](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/middleware/user_agent.go#L198) instead. I suggest we update the tests on the AWS side and make the appropriate changes there instead.

I think we can also probably handle the valid chars on the AWS side and just leave `smithy-go` UA abstraction alone. That will be the least invasive change and keep the changes isolated to AWS.


------

The other question I have is do we implement support for the app ID related settings anywhere?

wty-Bryant:
> This isn't correct. I think we'll need to handle the `#` on the AWS side [here](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/middleware/user_agent.go#L1).
> 
> For example the [addProductName](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/middleware/user_agent.go#L103) function calls
> 
> ```go
> builder.AddKeyValue(aws.SDKName, aws.SDKVersion)
> ```
> 
> This would now be formatted as `aws-sdk-go-v2#1.2.3` but it should still be `aws-sdk-go/1.2.3`. I think we'll have to handle the logic [here](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/middleware/user_agent.go#L198) instead. I suggest we update the tests on the AWS side and make the appropriate changes there instead.
> 
> I think we can also probably handle the valid chars on the AWS side and just leave `smithy-go` UA abstraction alone. That will be the least invasive change and keep the changes isolated to AWS.
> 
> The other question I have is do we implement support for the app ID related settings anywhere?

Nope, I plan to put app ID setting support in another PR cause we need to correct the UA syntax first.

aajtodd:
Can we close this if we aren't going to use it?

PR #430, opened 2023-06-22
Add endpoints2.0 middleware and param binding generation.

isaiahvita:
All endpoints2.0 middleware and parameter binding generation lives here, while hooks were added for builtin generation on consumer SDKs


lucix-aws (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/GoIntegration.java
```diff
@@ -186,4 +187,34 @@ default List<RuntimeClientPlugin> getClientPlugins() {
...
         return serviceId;
     }
+
+    /**
+     * Used by integrations to set a member on the endpoints resolution
+     * middleware object.
+     *
+     * @param writer Settings used to generate.
+     */
+    default void renderEndpointBuiltInField(GoWriter writer) {
```
aajtodd, line 197: **suggestion**: It might be better to collapse these into one method that returns a type responsible for these different things. 

e.g.

```
default void endpointBuiltinHandler(): EndpointBuiltInHandler


interface EndpointBuiltInHandler {
    default void renderEndpointBuiltInField(GoWriter writer) { }

    default void renderEndpointBuiltInInvocation(GoWriter writer) {}

    default void renderEndpointBuiltInInitialization(GoWriter writer, Parameters parameters) { }
}
```

Grouping these things together reduces the API surface of `GoIntegration`, it also helps in finding all the various methods that may need overridden for related functionality.

aajtodd (approved):

PR #431, opened 2023-06-25
Add hooks for consumer written auth scheme resolution

isaiahvita:
This adds hooks into GoIntegration for auth scheme resolution to be used by endpoint resolution middleware generation.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/AuthenticationSchemeGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+package software.amazon.smithy.go.codegen;
+
+import software.amazon.smithy.model.shapes.ServiceShape;
+
+/**
+ * Used by integrations to generate authentication
+ * scheme resolution.
+ *
+ */
+public interface AuthenticationSchemeGenerator {
```
isaiahvita, line 25: wanted to get folks' thoughts on this. creating an interface and making a getter on `GoIntegration` rather than embedding `renderEndpointBasedAuthSchemeResolution` directly into `GoIntegration`. im figuring this class can contain future auth scheme resolution changes?

maybe this is likely overkill for now since we dont know how SRA auth stuff will look in Go yet, so might as well just embed `renderEndpointBasedAuthSchemeResolution` into `GoIntegration`?

lucix-aws (changes requested):
See above comment.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/AuthenticationSchemeGenerator.java
isaiahvita: im guess that the intention is that the `EndpointPropertyRenderer` look something like this:

```java
public interface EndpointPropertyRenderer {
    default void renderEndpointProperty(GoWriter writer, Model model) {
        // pass
    }
}
```
i think this could potentially be a problem with forcing every future property into a specific signature. like what if a new property needs an additional parameter? i suppose this is fine while smithy-go isnt GA. but if we were, we would have to add a new overloaded method to the `EndpointPropertyRenderer`. would this be cleaner than adding a new interface for the property with its own GoIntegration overridable method? thoughts?

isaiahvita: discussed offline. there was a misunderstanding on how this integration was used. we decided to go with `renderPostEndpointResolutionHook` on the GoIntegration. its what kotlin did and its more amenable to us removing the endpoint-based auth scheme resolution stuff later with planned work for this year.

aajtodd (approved):

lucix-aws (approved):

PR #439, opened 2023-07-20
Disable Host Prefix in certain case

isaiahvita:
Instead of using a model filter to completely remove the endpoint host prefix middleware from every operation. Adding a runtime disabler allows for custom v1 resolvers to still use the endpoint host prefix middleware.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
@@ -77,7 +77,8 @@ public class EndpointHostPrefixMiddleware implements GoIntegration {
                         """
                         ctx = $T(ctx, true)
                         """,
-                        SymbolUtils.createPointableSymbolBuilder("DisableEndpointHostPrefix", SmithyGoDependency.SMITHY_HTTP_TRANSPORT).build()
+                        SymbolUtils.createPointableSymbolBuilder(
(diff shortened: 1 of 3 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
```diff
@@ -64,6 +65,29 @@ public void processFinalizedModel(GoSettings settings, Model model) {
...
+
+            for (Segment segment : endpointTrait.getHostPrefix().getLabels()) {
+                if (segment.isLabel() && segment.getContent().equals("AccountId") && !written) {
+                    writer.write(
+                        """
+                        ctx = $T(ctx, true)
+                        """,
+                        SymbolUtils.createPointableSymbolBuilder("DisableEndpointHostPrefix", SmithyGoDependency.SMITHY_HTTP_TRANSPORT).build()
+                    );
+                    written = true;
```
aajtodd, line 83: purpose?

aajtodd (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EndpointHostPrefixMiddleware.java
isaiahvita: i thought just in case there was a certain operation that satisfied:
```
                if (segment.isLabel() && segment.getContent().equals("AccountId") && !written) {
```
multiple times. i didnt want it to write out that context modification multiple times in the EP20 middleware

lucix-aws (approved):

PR #445, opened 2023-08-11
Modify and Merge protocol test request unit tests codegen logic

wty-Bryant:
Modify and Merge protocol test request unit tests codegen logic to isolate cases from local server and instead capture and check request directly via middleware

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
@@ -19,6 +19,7 @@ package software.amazon.smithy.go.codegen.integration;
 
 import java.util.function.Consumer;
 import java.util.logging.Logger;
+import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.go.codegen.GoWriter;
 import software.amazon.smithy.go.codegen.SmithyGoDependency;
 import software.amazon.smithy.go.codegen.SymbolUtils;
@@ -195,9 +196,7 @@ public class HttpProtocolUnitTestRequestGenerator extends HttpProtocolUnitTestGe
      *
      * @param writer writer to write generated code with.
      */
-    protected void generateTestBodySetup(GoWriter writer) {
-        writer.write("var actualReq *http.Request");
-    }
+    protected void generateTestBodySetup(GoWriter writer) {}
 
     /**
      * Hook to generate the HTTP response body of the protocol test.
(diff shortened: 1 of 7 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
```diff
@@ -282,7 +274,8 @@ protected void generateTestServer(
             String name,
             Consumer<GoWriter> handler
     ) {
-        super.generateTestServer(writer, name, handler);
+        // We aren't using a test server, but we do need a URL to set.
+        writer.write("serverURL := \"http://localhost:8888/\"");
```
lucix-aws, line 278: Can't we just remove this entire `generateTestServer` method? Is its code needed at all anymore?

transport/http/middleware_capture_request.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"github.com/aws/smithy-go/middleware"
+	"net/http"
+	"strconv"
+)
+
+const captureRequestID = "CaptureProtocolTestRequest"
+
+// AddCaptureRequestMiddleware captures serialized http request during protocol test for check
+func AddCaptureRequestMiddleware(stack *middleware.Stack, req *http.Request) error {
```
lucix-aws, line 14: **Question:** Do we _have_ to expose this in the public API?

If I understand correctly, the generated protocol tests in the SDK repo will have to see it to use it, but I think I'd rather we just generate it as unexported code in the protocol test instead of having it be exported in smithy-go just to be used for testing.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
```diff
@@ -250,20 +241,21 @@ protected void generateTestAssertions(GoWriter writer) {
         writeAssertNil(writer, "err");
         writeAssertNotNil(writer, "result");
 
-        writeAssertScalarEqual(writer, "c.ExpectMethod", "actualReq.Method", "method");
-        writeAssertScalarEqual(writer, "c.ExpectURIPath", "actualReq.URL.RawPath", "path");
+        writeAssertScalarEqual(writer, "c.ExpectMethod", "capturedReq.Method", "method");
```
lucix-aws, line 244: This is totally fine but you might have been able to get away with less changes by just connecting your new middleware to the originally declared `actualReq` request.

transport/http/middleware_capture_request.go
lucix-aws: Alternatively we can put it somewhere under `private/` like we've done in a few other places - to have it usable by the SDK, but to signal that we don't want other people doing so.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
wty-Bryant: If we do that we have to remove it from super class `HttpProtocolUnitTestGenerator` and simply provide this fake url to request and response generator separately.

transport/http/middleware_capture_request.go
wty-Bryant: I prefer alternative method cause the it prevent smithy go codegen using some sdk dependency of middleware. I will create a new `private/` directory under smithy-go repo and move the middleware there.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
wty-Bryant: Also we need the test cases' host info in some host test cases, I prefer to keep it now in case of breaking changes.

lucix-aws (approved):

PR #446, opened 2023-08-16
fix: properly prevent unreachable error codegen in endpoint rules

lucix-aws:
Prevents generation of unreachable code by not considering the static `uriEncode` function as a condition when deciding whether to generate fallback.

A source comment next to the change contains a more detailed explanation.


aajtodd (approved):
This may be something to talk with Smithy about...it feels like we shouldn't have to handle this and that they should maybe provide a utility. Never hurts to ask anyway.

isaiahvita (approved):

PR #447, opened 2023-08-16
Modify and Merge protocol test request unit tests codegen logic

wty-Bryant:
Modify and Merge protocol test request unit tests codegen logic to isolate cases from local server and instead capture and check request directly via middleware

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
@@ -19,6 +19,7 @@ package software.amazon.smithy.go.codegen.integration;
 
 import java.util.function.Consumer;
 import java.util.logging.Logger;
+import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.go.codegen.GoWriter;
 import software.amazon.smithy.go.codegen.SmithyGoDependency;
 import software.amazon.smithy.go.codegen.SymbolUtils;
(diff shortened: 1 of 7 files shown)


lucix-aws (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
```diff
@@ -236,8 +217,18 @@ protected void generateTestServerHandler(GoWriter writer) {
      */
     @Override
     protected void generateTestInvokeClientOperation(GoWriter writer, String clientName) {
+        Symbol stackSymbol = SymbolUtils.createPointableSymbolBuilder("Stack",
+                SmithyGoDependency.SMITHY_MIDDLEWARE).build();
         writer.addUseImports(SmithyGoDependency.CONTEXT);
-        writer.write("result, err := $L.$T(context.Background(), c.Params)", clientName, opSymbol);
+        writer.openBlock("result, err := $L.$T(context.Background(), c.Params, func(options *Options) {", "})",
```
isaiahvita, line 223: nit: in the future, we should avoid using a series of `openBlock`s like this, and just rely on `gofmt` to do all the formatting. see something like this
https://github.com/aws/smithy-go/blob/main/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/endpoints/EndpointMiddlewareGenerator.java#L266

isaiahvita (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpProtocolUnitTestRequestGenerator.java
```diff
@@ -282,7 +274,8 @@ protected void generateTestServer(
             String name,
             Consumer<GoWriter> handler
     ) {
-        super.generateTestServer(writer, name, handler);
+        // We aren't using a test server, but we do need a URL to set.
```
isaiahvita, line 277: nit: actually this comment doesnt do a lot of good when its generated. because the generated code wont have this comment. if we can, we should make the URL value even more obvious that its a placeholder (while still being parse-able by whatever needs it). for example, would `http://thisisa.placeholder:4242` work?

PR #449, opened 2023-09-20
add auth and http/auth types for SRA auth

lucix-aws:
Adds runtime types for SRA auth.


lucix-aws:
moved to https://github.com/aws/smithy-go/pull/456

PR #457, opened 2023-10-03
implement `Request.GetBody` to support retry in HTTP Client

smaeda-ks:
related: https://github.com/aws/aws-sdk-go-v2/discussions/2008

net/http:
https://github.com/golang/go/blob/638e0d36d2a6b2efce0550978bf48e7df1166a0b/src/net/http/client.go#L508-L538

net/http2:
https://github.com/golang/net/blob/a600b3518eed7a9a4e24380b4b249cb986d9b64d/http2/transport.go#L610-L644

> By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Yes.

### DIFF
--- a/transport/http/request.go
+++ b/transport/http/request.go
@@ -1,6 +1,7 @@
 package http
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


transport/http/request.go
```diff
@@ -176,6 +177,13 @@ func (r *Request) Build(ctx context.Context) *http.Request {
 		// "unknown". This is unwanted behavior.
 		if req.ContentLength != 0 && r.stream != nil {
 			req.Body = iointernal.NewSafeReadCloser(ioutil.NopCloser(stream))
+			// Implement GetBody so the HTTP Client can retry with body.
+			req.GetBody = func() (io.ReadCloser, error) {
+				buf := new(bytes.Buffer)
```
lucix-aws, line 182: I think this is something we _want_ to do at a surface level but I don't want to just blindly switch to in-memory on a redirect.

At the very least I would think we can assert whether the body is a seeker such that we could rewind instead.

Beyond that I'm questioning whether we should do the memory fallback at all, or at least have it be opt-in.

PR #462, opened 2023-10-16
feat: sra i&a: implement auth resolution workflow in middleware

lucix-aws:
main changes:

* Add authentication middlewares:
  * `resolveAuthScheme`
  * `resolveIdentity`
  * `signRequestMiddleware`
  * **NOTE:** I haven't reordered the phasing for these at this time, auth resolution currently happens before endpoint resolution (so it's still in serialize unlike how we spec'd to move it to finalize). The operation input isn't available in `FinalizeInput`, we _can_ add it but I'm leaving it as something to explore externally. For the most part I think what's important is that identity resolution is part of finalize such that retries handle it.
* Persist resolved auth scheme and identity on operation context.
* Merge resolved endpoint signer properties into resolved scheme signer properties.
* Add auth schemes and auth scheme resolver to config
* Extend `AuthSchemeDefinition` to expose codegen to plug in `IdentityResolverOptions` expressions and `AuthScheme` expressions.

misc:

* Remove `operationName` interface on input structs. Operation name is now bound directly on middleware add.
* Correct typing for property getter keys. They have to be structs instead of `var`s of anonymous empty structs.
* Remove `HttpBearerAuth` integration (moved to SDK), the generated code is not compatible with I&A as it bakes the signer into the client config.
* Remove "default" auth integrations for now, overriding the integration SDK-side isn't something I want to handle at this time.
* Refactor endpoint resolution property codegen, returned endpoint properties are now strongly typed and set accordingly.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
```diff
@@ -998,6 +999,18 @@ public ChainWritable() {
             writables = new ArrayList<>();
         }
 
+        public static ChainWritable of(GoWriter.Writable... writables) {
```
isaiahvita, line 1002: nice

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -294,6 +316,71 @@ private void generateApplicationProtocolTypes() {
         }).write("");
     }
 
+    private void writeProtocolResolvers() {
```
isaiahvita, line 319: nit: `writeProtocolResolvers` and `generateProtocolResolvers` sound like the same thing. it looks like the former is more writing/generating the invocations and the latter is doing the definitions, so maybe work those intentions into the name of the functions

transport/http/auth.go
```diff
@@ -2,7 +2,6 @@ package http
 
 import (
 	"context"
-	"net/http"
```
isaiahvita, line 5: nice, i wonder if it was just an oversight that we werent using the smithy request object already

isaiahvita (approved):

transport/http/auth.go
lucix-aws: Well this is the new signer type from a previous PR, but yes I felt that it should use the smithy transport type instead of the underlying one.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/auth/GetIdentityMiddlewareGenerator.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private final ProtocolGenerator.GenerationContext context;
+
+    public GetIdentityMiddlewareGenerator(ProtocolGenerator.GenerationContext context) {
+        this.context = context;
+    }
+
+    public static GoWriter.Writable generateAddMiddleware() {
+        return goTemplate("""
+                err = stack.Finalize.Add(&$L{
```
syall, line 40: I'm not 100% sure about this, but I don't think this middleware is needed.

In `generateSelectScheme()` of the auth scheme middleware, identity resolvers are already invoked to see if it's valid or not for choosing an auth scheme, so I think you can just attach that resolved identity to the `resolvedAuthScheme`.

lucix-aws: We check that the resolver _exists_ there at that time, but we don't actually get the identity until this middleware (hence `getIdentityMiddleware`).

We could definitely save the resolver on the first call as you've said though it's really only a string lookup.

syall (approved):

PR #465, opened 2023-10-27
feat: sra i&a: add sigv4a and reorder operation phases

lucix-aws:
* Add sigv4a auth definition
  * There's a temporary sigv4a "trait" which is used downstream, to be replaced by the real thing at the time it's added
* Reorder operation phasing: auth resolution, identity retrieval, endpoint resolution, and signing now occur in that order at the beginning of the finalize phase
  * The operation input is now stored on the context at the beginning of serialize to be accessed as necessary by later phases
  * "resolve endpoint" middleware is now always generated and simply empty without a ruleset. A large number of components index off of it (both here and in the SDK downstream), better to have it be an "explicit" phase to not have to worry about it being there for stack operations
  * Path merging behavior now occurs as part of endpoint resolution, since that happens after serialization
* Add anonymous identity and resolver
* Give `ConfigFieldResolver` a new target for finalizing post-client creation (reverting the behavior of the old "finalize" to happen before client creation such that option modifications actually persist)

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/auth/SigV4aTrait.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/auth/SigV4aTrait.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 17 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/auth/SigV4aTrait.java
```diff
@@ -0,0 +1,41 @@
...
+package software.amazon.smithy.go.codegen.auth;
+
+import software.amazon.smithy.model.node.Node;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.AbstractTrait;
+
+/**
+ * Fake trait for aws.auth#sigv4a until smithy adds it.
+ */
+public final class SigV4aTrait extends AbstractTrait {
```
syall, line 25: nit: mark as `@SmithyInternalApi`

syall (approved):

PR #467, opened 2023-11-02
Add Gzip request compression feature

wty-Bryant:
Add Gzip request compression middleware and codegen logic in smithy go


lucix-aws:
This needs a changelog entry, the tool we use works exactly as it does in the SDK.

private/requestcompression/request_compression.go
```diff
@@ -0,0 +1,105 @@
...
+	"context"
+	"fmt"
+	"github.com/aws/smithy-go/middleware"
+	"github.com/aws/smithy-go/transport/http"
+	"io"
+)
+
+// AddRequestCompression add requestCompression middleware to op stack
+func AddRequestCompression(stack *middleware.Stack, DisableRequestCompression bool, RequestMinCompressSizeBytes int64) error {
+	return stack.Build.Add(&requestCompression{
```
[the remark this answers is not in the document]
wty-Bryant: "Clients MUST encode the request content prior to request signing, otherwise the request signature will mismatch."

According to SEP requirement, I agree with you that this middleware should be injected at an earlier stage

private/requestcompression/request_compression.go
```diff
@@ -0,0 +1,105 @@
...
+		disableRequestCompression:   DisableRequestCompression,
+		requestMinCompressSizeBytes: RequestMinCompressSizeBytes,
+	}, middleware.After)
+}
+
+type requestCompression struct {
+	disableRequestCompression   bool
```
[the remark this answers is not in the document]
wty-Bryant: I agree that the supported algorithm list has been checked at smithy model side. Considering we don't config algorithm list now, I prefer to just set the list here as some constant

private/requestcompression/request_compression.go
```diff
@@ -0,0 +1,137 @@
...
+
+	req, ok := in.Request.(*http.Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown request type %T", req)
+	}
+
+	for _, algorithm := range m.compressAlgorithms {
+		compressFunc := allowedAlgorithms[algorithm]
+		if compressFunc != nil {
+			if stream := req.GetStream(); stream != nil {
```
lucix-aws, line 77: A few questions--

1. is the `else` block necessary? AFAIK we _always_ use the `Get/SetStream` wrapped from smithy go to modify the request body.
2. I suspect this doesn't handle event streams. each individual message payload would have to be zipped, much like how we have to sign each payload

private/requestcompression/request_compression.go
```diff
@@ -0,0 +1,137 @@
...
+				}
+				compressedBytes, err := compressFunc(bytes.NewReader(body))
+				if err != nil {
+					return out, metadata, fmt.Errorf("failed to compress request body, %v", err)
+				}
+
+				req.Body = io.NopCloser(bytes.NewReader(compressedBytes))
+			}
+
+			req.Header.Add("Content-Encoding", "gzip")
```
lucix-aws, line 108: Shouldn't you take the name of the selected algorithm instead of hardcoding gzip?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/requestcompression/RequestCompression.java
```diff
@@ -0,0 +1,128 @@
...
+                func $add:L(stack $stack:P, options Options) error {
+                    return $addInternal:T(stack, options.DisableRequestCompression, options.RequestMinCompressSizeBytes,
+                    $algorithms:L)
+                }
+                """,
+                MapUtils.of(
+                "add", ADD_REQUEST_COMPRESSION,
+                "stack", stackSymbol,
+                "addInternal", addInternalSymbol,
+                "algorithms", String.format("\"%s\"", String.join(",", algorithms))
```
lucix-aws, line 90: There's no reason to microformat the list of algorithms like this, just generate it directly, e.g.

```
@requestCompression(
    encodings: ["gzip", "br"]
)
```

translates to list

```
[]string{"gzip", "br"}
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/requestcompression/RequestCompression.java
```diff
@@ -0,0 +1,128 @@
...
+            SymbolProvider symbolProvider,
+            GoDelegator goDelegator
+    ) {
+        ServiceShape service = settings.getService(model);
+        if (!isRequestCompressionService(model, service)) {
+            return;
+        }
+
+        Set<String> algorithms = RequestCompressionTrait.SUPPORTED_COMPRESSION_ALGORITHMS;
+        goDelegator.useShapeWriter(service, writeMiddlewareHelper(algorithms));
```
lucix-aws, line 62: We have to generate this *per-operation* since the compression trait is applied there.

e.g. consider the following three modeled operations and their expected codegen:

```
@requestCompression{
    encodings: ["gzip"]
}
operation Foo {
  // ...
}
// should generate:
func addOperationFooRequestCompressionMiddleware(...) {
	return stack.Serialize.Add(&requestCompression{
                // ...
		compressAlgorithms: []string{"gzip"},
	}, middleware.After)
}
// and the operation stack setup should now include:
func (c *Client) addOperationFooMiddlewares(...) {
    // ...
    if err = addOperationFooRequestCompressionMiddleware(...); err != nil {
        return err
    }
    // ...
}

@requestCompression{
    encodings: ["br", "gzip"]
}
operation Bar {
  // ...
}
// should generate:
func addOperationBarRequestCompressionMiddleware(...) {
	return stack.Serialize.Add(&requestCompression{
                // ...
		compressAlgorithms: []string{"br", "gzip"},
	}, middleware.After)
}
// and the operation stack setup should now include:
func (c *Client) addOperationBarMiddlewares(...) {
    // ...
    if err = addOperationBarRequestCompressionMiddleware(...); err != nil {
        return err
    }
    // ...
}

// no request compression
operation Bar {
  // ...
}
// should not generate a middleware, and the operation stack setup shouldn't have anything new
```

See `AwsHttpChecksumGenerator.java` for an example of how this is done.

private/requestcompression/request_compression.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+type compressFunc func(io.Reader) ([]byte, error)
+
+var allowedAlgorithms = map[string]compressFunc{
+	GZIP: gzipCompress,
+}
+
+// AddRequestCompression add requestCompression middleware to op stack
+func AddRequestCompression(stack *middleware.Stack, DisableRequestCompression bool, RequestMinCompressSizeBytes int64, algorithms string) error {
```
lucix-aws, line 34: lowercase the param names

lucix-aws: also it's in-context so e.g. `DisableRequestCompression` can become simply `disabled` etc

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/requestcompression/RequestCompression.java
wty-Bryant: Okay I previously thought the `SUPPORTED_COMPRESSION_ALGORITHMS` of that trait is static so service level middleware is enough, but seems like we have multi-encodings for diff ops, I will change that

private/requestcompression/request_compression.go
wty-Bryant: I agree, but I prefer to keep the naming in middleware fields in case of extra comment

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/requestcompression/RequestCompression.java
```diff
@@ -0,0 +1,166 @@
...
+
+    public static boolean isRequestCompressionService(Model model, ServiceShape service) {
+        return TopDownIndex.of(model)
+                .getContainedOperations(service).stream()
+                .anyMatch(it -> it.hasTrait(RequestCompressionTrait.class));
+    }
+
+    private GoWriter.Writable writeMiddlewareHelper(SymbolProvider symbolProvider, OperationShape operation) {
+        var stackSymbol = SymbolUtils
+                .createPointableSymbolBuilder("Stack", SmithyGoDependency.SMITHY_MIDDLEWARE)
```
lucix-aws, line 107: `SmithyGoTypes` was merged with the SRA auth refactor - you can convert wrote non-generated imports like this to `SmithyGoTypes.Middleware.Stack` etc.

(add static symbols to `SmithyGoTypes` as needed)

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/requestcompression/RequestCompression.java
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                    SymbolProvider symbolProvider = GoCodegenPlugin.createSymbolProvider(model, settings);
+                    String funcName = getAddRequestCompressionMiddlewareFuncName(
+                            symbolProvider.toSymbol(operation).getName()
+                    );
+                    runtimeClientPlugins.add(RuntimeClientPlugin.builder().operationPredicate((m, s, o) -> {
+                        if (!o.hasTrait(RequestCompressionTrait.class)) {
+                            return false;
+                        }
+                        return o.equals(operation);
```
syall, line 72: Is there a reason why the runtime client plugin is not defined in `getClientPlugins()`?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/requestcompression/RequestCompression.java
```diff
@@ -0,0 +1,160 @@
...
+import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.traits.RequestCompressionTrait;
+import software.amazon.smithy.utils.ListUtils;
+import software.amazon.smithy.utils.MapUtils;
+
+
+public final class RequestCompression implements GoIntegration {
```
syall, line 44: nit overall: private fields should go below public fields, and private methods should go towards the bottom of the class below all the public methods.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/requestcompression/RequestCompression.java
```diff
@@ -0,0 +1,160 @@
...
+                "add", getAddRequestCompressionMiddlewareFuncName(operationName),
+                "stack", SmithyGoTypes.Middleware.Stack,
+                "addInternal", SmithyGoTypes.Private.RequestCompression.AddRequestCompression,
+                "algorithms", generateAlgorithmList(trait.getEncodings())
+                ));
+    }
+
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins() {
+       runtimeClientPlugins.add(
```
syall, line 124: Use `List.of()` instead, and register all client plugins here immutably instead of using state.

syall (changes requested):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SmithyGoDependency.java
```diff
@@ -53,6 +53,8 @@ public final class SmithyGoDependency {
     public static final GoDependency SMITHY_HTTP_TRANSPORT = smithy("transport/http", "smithyhttp");
     public static final GoDependency SMITHY_MIDDLEWARE = smithy("middleware");
     public static final GoDependency SMITHY_PRIVATE_PROTOCOL = smithy("private/protocol", "smithyprivateprotocol");
+    public static final GoDependency SMITHY_REQUEST_COMPRESSION =
+    smithy("private/requestcompression", "smithyrequestcompression");
```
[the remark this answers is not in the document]
wty-Bryant: smithy go doesn't allow single line exceeding 120 chars, that's why there's a separate line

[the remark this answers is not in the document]
syall:
```suggestion
    public static final GoDependency SMITHY_REQUEST_COMPRESSION =
        smithy("private/requestcompression", "smithyrequestcompression");
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/requestcompression/RequestCompression.java
wty-Bryant: One possible reason is we need  `GoSettings` to create a `SymbolProvider` for resolving middleware fn name for each op, which is not provided in `getClientPlugins` params

syall: https://github.com/aws/smithy-go/issues/485

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/requestcompression/RequestCompression.java
lucix-aws: `getClientPlugins` should probably accept the `Model` and service shape ID much like these other functions do long-term.

lucix-aws (approved):

codegen/smithy-go-codegen/src/main/resources/META-INF/services/software.amazon.smithy.go.codegen.integration.GoIntegration
```diff
@@ -12,3 +12,5 @@ software.amazon.smithy.go.codegen.endpoints.EndpointClientPluginsGenerator
 # modeled auth schemes
 software.amazon.smithy.go.codegen.integration.auth.SigV4AuthScheme
 software.amazon.smithy.go.codegen.integration.auth.AnonymousAuthScheme
+
```
syall, line 15: nit: remove extra line

syall (approved):

PR #469, opened 2023-11-07
feat: implement smithy reference arch for identity and auth resolution

lucix-aws:
This is the final merge candidate for SRA identity and auth.

It is more or less the sum of all the previous PRs against the `feat-sraauth` branch, except I removed the sugar around constructing auth options. I wasn't particularly happy with that (had concerns about differentiating between zero values and unset), they weren't at all necessary either since you can create auth options directly.

The most important thing here is to review the API surface and make sure we're ready to commit.


isaiahvita (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SetOperationInputContextMiddleware.java
```diff
@@ -0,0 +1,41 @@
...
+
+import static software.amazon.smithy.go.codegen.GoStackStepMiddlewareGenerator.createSerializeStepMiddleware;
+import static software.amazon.smithy.go.codegen.GoWriter.emptyGoTemplate;
+import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
+
+/**
+ * Middleware to set the final operation input on the context at the start of the serialize step such that protocol
+ * middlewares in later phases can use it.
+ */
+public class SetOperationInputContextMiddleware {
```
syall, line 26: more of a nit, but does this need its own middleware if it's just a 1-line op? why not inline?

syall (approved):

PR #472, opened 2023-11-15
pass through unrecognized signer names as-is

lucix-aws:
Endpoint resolution spec explicitly says we're not supposed to error on unrecognized signer names, this is a minor regression from the merge of #469. This change restores the intended behavior.


isaiahvita (approved):

syall (approved):

isaiahvita:
approved, but can you provide a brief description on the motivation in the PR details

PR #496, opened 2024-02-02
Add accountID into endpoint2.0 param binding

wty-Bryant:
During the ep2.0 resolution, the accountID param is retrieved from ctx's credentials, thus in this PR, the ctx is added as a new param into endpoint binding func codegen.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/auth/AuthParametersResolverGenerator.java
```diff
@@ -42,7 +43,7 @@ public GoWriter.Writable generate() {
         loadResolvers();
 
         return goTemplate("""
-                func $name:L(operation string, input interface{}, options Options) $params:P {
+                func $name:L(operation string, input interface{}, options Options, ctx $context:T) $params:P {
```
lucix-aws, line 46: `Context`, as a rule, goes first in an API's argument list.

lucix-aws (changes requested):

lucix-aws (approved):

PR #515, opened 2024-06-04
Add time.Duration to the list of Go Stdlib Types

Madrigal:
*Issue #, if available:*
N/A

*Description of changes:*
Adds `time.Duration` to the list of Go standard types. This will be used by later PRs 

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStdlibTypes.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStdlibTypes.java
@@ -35,6 +35,7 @@ public final class GoStdlibTypes {
 
     public static final class Time {
         public static final Symbol Time = SmithyGoDependency.TIME.pointableSymbol("Time");
+        public static final Symbol Duration = SmithyGoDependency.TIME.valueSymbol("Duration");
     }
 
     public static final class Encoding {


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStdlibTypes.java
```diff
@@ -35,6 +35,7 @@ public static final class Context {
 
     public static final class Time {
         public static final Symbol Time = SmithyGoDependency.TIME.pointableSymbol("Time");
+        public static final Symbol Duration = SmithyGoDependency.TIME.valueSymbol("Duration");
```
lucix-aws, line 38: FWIW we've kind of stopped doing this because there are just too many -- instead we just started calling these directly, and we added some helpers so you can do stuff like

`SmithyGoDependency.TIME.struct("Time")`

This is fine though

lucix-aws (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoStdlibTypes.java
Madrigal: Got it, nah I rather do it The Right Way ™

Will close the PR

Issue #238, opened 2020-12-03
refactor smithy-waiters to remove jmespath dependency at runtime


skotambkar:
* smithy-waiters should use smithy-generated ASTs, instead of using jmespath at runtime

skotambkar:
This refactor should address SDK not correctly supporting jmespath expressions consisting of `*` (a wildcard operator). 

Relates to #265

PR #519, opened 2024-06-25
add codegen support for string array and operation context params in endpoint bindings

lucix-aws:
* Add a `GoCodegenContext` field to `CodegenVisitor` and pass it down. This context object contains a handful of principal codegen resources which everything generally needs, and passing it simplifies a lot of constructor calls (observe the difference in `OperationGenerator`. This matches the new `DirectedCodegen` pattern.
* Adds a basic JMESpath expression generator. This is used for the new "operation context params" binding trait but lays the groundwork for getting rid of runtime jmespath in the future (#238).
* Add support for STRING_ARRAY type of endpoint parameter
* Add support for jmespath-based operation context params

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.go.codegen;
+
+import static software.amazon.smithy.go.codegen.util.ShapeUtil.STRING_SHAPE;
+import static software.amazon.smithy.go.codegen.util.ShapeUtil.expectMember;
+import static software.amazon.smithy.go.codegen.util.ShapeUtil.listOf;
+import static software.amazon.smithy.utils.StringUtils.capitalize;
+
+import java.util.List;
+import software.amazon.smithy.codegen.core.CodegenException;
+import software.amazon.smithy.jmespath.JmespathExpression;
+import software.amazon.smithy.jmespath.ast.FieldExpression;
+import software.amazon.smithy.jmespath.ast.FunctionExpression;
+import software.amazon.smithy.jmespath.ast.ProjectionExpression;
+import software.amazon.smithy.jmespath.ast.Subexpression;
+import software.amazon.smithy.model.shapes.ListShape;
+import software.amazon.smithy.model.shapes.Shape;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+/**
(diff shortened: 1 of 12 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+        this.root = expr;
+    }
+
+    public Result generate(String ident) {
+        writer.write("v1 := $L", ident);
+        return visit(root, input);
+    }
+
+    private Result visit(JmespathExpression expr, Shape current) {
+        if (expr instanceof FunctionExpression) {
```
Madrigal, line 63: [nit] Since Java 17 you can do 
```
if (expr instanceof FunctionExpression fexpr) {
    // look ma, typecasting is already done
    return visitFunction(fexpr, current)
}
```

[also nit, but tangential] if we could switch to Java 21 we could even do Pattern Matching and do something like
```
return switch (expr) {
    case FunctionExpression fexpr -> return visitFunction(fexpr, current);
    ....
}
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Traverses a JMESPath expression, producing a series of statements that evaluate the entire expression. The generator
+ * is shape-aware and the return indicates the underlying shape being referenced in the final result.
+ * <br/>
+ * Note that the use of writer.write() here is deliberate, it's easier to structure the code in that way instead of
+ * trying to recursively compose/organize Writable templates.
+ */
+@SmithyInternalApi
+public class GoJmespathExpressionGenerator {
```
Madrigal, line 42: We should really have tests for this class

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        var leftMember = expectMember(ctx.model(), (ListShape) left.shape);
+
+        // We have to know the element type for the list that we're generating, use a dummy writer to "peek" ahead and
+        // get the traversal result
+        var lookahead = new GoJmespathExpressionGenerator(ctx, new GoWriter(""), leftMember, expr.getRight())
+                .generate("v");
+
+        ++idIndex;
```
Madrigal, line 91: not quite sure I follow why we increase the index when visiting the projection. Each field increases the index because we're visiting a concrete field, but why do we increase it here?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+        writer.write("""
+                var v$1L []string
+                for k := range $2L {
+                    v$1L = append(v$1L, k)
+                }""", idIndex, arg.ident);
+
+        return new Result(listOf(STRING_SHAPE), "v" + idIndex);
+    }
+
+    public record Result(Shape shape, String ident) {}
```
Madrigal, line 141: <3

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
lucix-aws: We need a new identifier for the slice variable that holds the result of the overall projection (it's declared in the following write()).

wty-Bryant (approved):

Madrigal (approved):

PR #520, opened 2024-06-27
true up GoDelegator and GoCodegenContext integrations

lucix-aws:
* makes `GoDelegator` extend `WriterDelegator<GoWriter`, making it directly compatible with the corresponding field in the codegen context
* adds new `writeAdditionalFiles(ctx)` hook, which completely obsoletes the other two versions with big arg lists
* partial refactor of waiters integration (relevant for upcoming jmespath codegen) to demonstrate this

This is a ***net-zero*** codegen change - tested by regenerating SDK v2.


codegen/smithy-go-codegen/src/test/java/software/amazon/smithy/go/codegen/GenerateStandaloneGoModuleTest.java
```diff
@@ -25,8 +25,9 @@ public void testGenerateGoModule() throws Exception {
         var testPath = Files.createTempDirectory(getClass().getName());
         LOGGER.warning("generating test suites into " + testPath);
 
+        // symbol provider not needed here since only useFileWriter is called
         var fileManifest = FileManifest.create(testPath);
-        var writers = new GoWriterDelegator(fileManifest);
+        var writers = new GoDelegator(fileManifest, null);
```
Madrigal, line 30: Can we create another constructor without a Symbol provider instead?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -60,58 +59,36 @@ public Set<Symbol> getAdditionalClientOptions() {
...
-            Model model,
-            SymbolProvider symbolProvider,
-            GoWriter writer,
-            ServiceShape service,
-            OperationShape operation,
-            Map<String, Waiter> waiters
-    ) {
-        // generate waiter function
-        waiters.forEach((name, waiter) -> {
-            // write waiter options
```
Madrigal, line 100: nice

Madrigal (approved):

wty-Bryant (approved):

codegen/smithy-go-codegen/src/test/java/software/amazon/smithy/go/codegen/GenerateStandaloneGoModuleTest.java
lucix-aws: I would err on the side of no since it's not really supposed to be used that way (useShapeWriter calls would explode). This is an ancient test and it was easier to just skip it instead of building up the GoSettings component required for the symbol provider.

PR #527, opened 2024-07-17
add rollout switches for waiters

lucix-aws:


wty-Bryant (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
```diff
@@ -58,8 +58,16 @@ public Set<Symbol> getAdditionalClientOptions() {
         return emptySet();
     }
 
+    public boolean enabledForService(Model model, ShapeId service) {
```
Madrigal, line 61: Why do we return true on both waiters? I see that on aws/aws-sdk-go-v2/#2717 we use a public member on waiters2 instead of relying on this interface

lucix-aws: The default is true, SDK v2 is extending both of these hooks to be conditional.

Madrigal: ah, I see, `AwsWaiters` extends from this class and it defines whether it should use this or not.

PR #529, opened 2024-08-06
Add serializers middleware logic for protocol generator awsjson10.

nathanhit:
*Issue #, if available:*

*Description of changes:*
* Update Protocol Generator request serializer generation to a cleaner and complete build.
* Add a new class SerializeMiddleware that handles all generation for request serializers.
* Add support for shared serializers to handle complex input shapes, nested within serializer request generaiton.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
@@ -32,10 +32,6 @@ import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.utils.SmithyInternalApi;
 
-
-
-
-
 @SmithyInternalApi
 public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
 
@@ -52,13 +48,18 @@ public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
     @Override
     public void generateRequestSerializers(GenerationContext context) {
         var writer = context.getWriter().get();
-        var model = context.getModel();
-        var ops = model.getOperationShapes();
+        var ops = context.getModel().getOperationShapes();
+        for (var op : ops) {
+            var middle = new SerializeMiddleware(context, op, writer);
+            writer.write(middle.generate());
+            writer.write("\n");
+        }
+        generateSharedSerializers(context, writer, ops);
+    }
+
+    public void generateSharedSerializers(GenerationContext context, GoWriter writer, Set<OperationShape> ops) {
         Set<Shape> shared = new java.util.HashSet<>(Set.of());
         for (var op : ops) {
-            var middle = new SerializeMiddleware(this, context, op, writer);
-            writer.write(middle.generateOperationStubs());
-            writer.write("\n");
             Set<Shape> shapes = getShapesToSerde(context.getModel(), context.getModel().expectShape(
                     op.getInputShape()));
             shared.addAll(shapes);
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
@@ -71,7 +71,7 @@ public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
 
     @Override
     public void generateResponseDeserializers(GenerationContext context) {
-        GoWriter writer = context.getWriter().get();
+        var writer = context.getWriter().get();
         var model = context.getModel();
         var ops = model.getOperationShapes();
         for (var op : ops) {
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
@@ -17,23 +17,28 @@ package software.amazon.smithy.go.codegen.protocol.aws;
 
 import static software.amazon.smithy.go.codegen.ApplicationProtocol.createDefaultHttpApplicationProtocol;
 import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
+import static software.amazon.smithy.go.codegen.serde.SerdeUtil.getShapesToSerde;
 
 import java.util.Map;
+import java.util.Set;
 import software.amazon.smithy.aws.traits.protocols.AwsJson1_0Trait;
 import software.amazon.smithy.go.codegen.ApplicationProtocol;
 import software.amazon.smithy.go.codegen.GoWriter;
 import software.amazon.smithy.go.codegen.SmithyGoDependency;
-//import software.amazon.smithy.go.codegen.SmithyGoTypes;
 import software.amazon.smithy.go.codegen.integration.ProtocolGenerator;
+import software.amazon.smithy.go.codegen.server.protocol.JsonSerializerGenerator;
 import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.Shape;
 import software.amazon.smithy.model.shapes.ShapeId;
-//import software.amazon.smithy.model.shapes.StructureShape;
-//import software.amazon.smithy.utils.MapUtils;
 import software.amazon.smithy.utils.SmithyInternalApi;
+
+
+
 
 
 @SmithyInternalApi
 public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
+
     @Override
     public ShapeId getProtocol() {
         return AwsJson1_0Trait.ID;
@@ -49,14 +54,22 @@ public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
         var writer = context.getWriter().get();
         var model = context.getModel();
         var ops = model.getOperationShapes();
+        Set<Shape> shared = new java.util.HashSet<>(Set.of());
         for (var op : ops) {
-            requestSerializerCode(op, writer);
+            var middle = new SerializeMiddleware(this, context, op, writer);
+            writer.write(middle.generateOperationStubs());
+            writer.write("\n");
+            Set<Shape> shapes = getShapesToSerde(context.getModel(), context.getModel().expectShape(
+                    op.getInputShape()));
+            shared.addAll(shapes);
         }
+        var generator = new JsonSerializerGenerator(context.getModel(), context.getSymbolProvider());
+        writer.write(generator.generate(shared));
     }
 
     @Override
     public void generateResponseDeserializers(GenerationContext context) {
-        var writer = context.getWriter().get();
+        GoWriter writer = context.getWriter().get();
         var model = context.getModel();
         var ops = model.getOperationShapes();
         for (var op : ops) {
@@ -84,48 +97,6 @@ public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
         // TODO
     }
 
-    private void requestSerializerCode(OperationShape op, GoWriter writer) {
-        var opName = "awsAwsjson10_serializeOp" + op.toShapeId().getName();;
-
-        /*Struct Definition*/
-        var struct = goTemplate("""
-                type $L struct{
-                }
-                """, opName
-        );
-        writer.write(struct);
-
-
-        /* ID Function */
-        var idFunction = goTemplate("""
-                func (op *$L) ID() string {
-                    return "OperationSerializer"
-                }
-                """, opName
-        );
-        writer.write(idFunction);
-
-        /* Handle Serialize Function */
-        var handleFunction = goTemplate(
-                    """
-                    func (op *$structName:L) HandleSerialize (ctx $context:T, in $input:T, next $handler:T) (
-                    out $output:T, metadata $metadata:T, err error) {
-                        return next.HandleSerialize(ctx, in)
-                    }
-                    """, Map.of(
-                            "context", SmithyGoDependency.CONTEXT.interfaceSymbol("Context"),
-                        "input", SmithyGoDependency.SMITHY_MIDDLEWARE.struct("SerializeInput"),
-                        "handler", SmithyGoDependency.SMITHY_MIDDLEWARE.struct("SerializeHandler"),
-                        "output", SmithyGoDependency.SMITHY_MIDDLEWARE.struct("SerializeOutput"),
-                        "metadata", SmithyGoDependency.SMITHY_MIDDLEWARE.struct("Metadata"),
-                        "structName", opName
-                ));
-        writer.write(handleFunction);
-
-        /* Operation End */
-        writer.write("\n");
-    }
-
     private void responseDeserializerCode(OperationShape op, GoWriter writer) {
         var opName = "awsAwsjson10_deserializeOp" + op.toShapeId().getName();
 
@@ -150,11 +121,11 @@ public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
         //Handle Serialize Function
         var handleFunction = goTemplate(
                 """
-                func (op *$structName:L) HandleDeserialize (ctx $context:T, in $input:T, next $handler:T) (
-                out $output:T, metadata $metadata:T, err error) {
-                    return out, metadata, err
-                }
-                """, Map.of(
+                        func (op *$structName:L) HandleDeserialize (ctx $context:T, in $input:T, next $handler:T) (
+                        out $output:T, metadata $metadata:T, err error) {
+                            return out, metadata, err
+                        }
+                        """, Map.of(
                         "context", SmithyGoDependency.CONTEXT.interfaceSymbol("Context"),
                         "input", SmithyGoDependency.SMITHY_MIDDLEWARE.struct("DeserializeInput"),
                         "handler", SmithyGoDependency.SMITHY_MIDDLEWARE.struct("DeserializeHandler"),
@@ -167,3 +138,4 @@ public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
         writer.write("\n");
     }
 }
+
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
@@ -19,6 +19,7 @@ import static software.amazon.smithy.go.codegen.ApplicationProtocol.createDefaul
 import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
 import static software.amazon.smithy.go.codegen.serde.SerdeUtil.getShapesToSerde;
 
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import software.amazon.smithy.aws.traits.protocols.AwsJson1_0Trait;
@@ -58,7 +59,7 @@ public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
     }
 
     public void generateSharedSerializers(GenerationContext context, GoWriter writer, Set<OperationShape> ops) {
-        Set<Shape> shared = new java.util.HashSet<>(Set.of());
+        Set<Shape> shared = new HashSet<>();
         for (var op : ops) {
             Set<Shape> shapes = getShapesToSerde(context.getModel(), context.getModel().expectShape(
                     op.getInputShape()));
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
@@ -1,0 +1,223 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.go.codegen.protocol.aws;
+
+import static software.amazon.smithy.go.codegen.GoStackStepMiddlewareGenerator.createSerializeStepMiddleware;
+import static software.amazon.smithy.go.codegen.GoWriter.emptyGoTemplate;
+import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
+import static software.amazon.smithy.go.codegen.server.protocol.JsonSerializerGenerator.getSerializerName;
+
+import java.util.Map;
+import software.amazon.smithy.go.codegen.EventStreamGenerator;
+import software.amazon.smithy.go.codegen.GoStdlibTypes;
+import software.amazon.smithy.go.codegen.GoWriter;
+import software.amazon.smithy.go.codegen.SmithyGoDependency;
+import software.amazon.smithy.go.codegen.SmithyGoTypes;
+import software.amazon.smithy.go.codegen.integration.ProtocolGenerator;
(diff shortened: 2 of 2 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
```diff
@@ -49,14 +54,22 @@ public void generateRequestSerializers(GenerationContext context) {
         var writer = context.getWriter().get();
         var model = context.getModel();
         var ops = model.getOperationShapes();
+        Set<Shape> shared = new java.util.HashSet<>(Set.of());
         for (var op : ops) {
-            requestSerializerCode(op, writer);
```
lucix-aws, line 53: You can use two loops here - keep these logically separate, the cost of double iteration is insignificant in this context.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
```diff
@@ -49,14 +54,22 @@ public void generateRequestSerializers(GenerationContext context) {
...
+            shared.addAll(shapes);
         }
+        var generator = new JsonSerializerGenerator(context.getModel(), context.getSymbolProvider());
+        writer.write(generator.generate(shared));
     }
 
     @Override
     public void generateResponseDeserializers(GenerationContext context) {
-        var writer = context.getWriter().get();
+        GoWriter writer = context.getWriter().get();
```
lucix-aws, line 72: No reason to change, things like this clog the diff in the PR, making the change harder to review.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+import software.amazon.smithy.go.codegen.SmithyGoDependency;
+import software.amazon.smithy.go.codegen.SmithyGoTypes;
+import software.amazon.smithy.go.codegen.integration.ProtocolGenerator;
+import software.amazon.smithy.go.codegen.integration.ProtocolUtils;
+import software.amazon.smithy.go.codegen.trait.BackfilledInputOutputTrait;
+import software.amazon.smithy.model.knowledge.EventStreamIndex;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.utils.MapUtils;
+
```
lucix-aws, line 36: Avoid excessive newlines.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+
+
+
+
+@SuppressWarnings("checkstyle:RegexpSingleline")
```
lucix-aws, line 41: We need justification to suppress something like this - in this case we have none. Remove and address the underlying issue.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+                    }
+                    _ = input
+                
+                    req, ok := in.Request.($request:P)
+                    if !ok {
+                        return out, metadata, $errorf:T("unexpected transport type %T", in.Request)
+                    }
+                
+                    req.Method = $methodPost:T
+                    req.URL.Path = "/service/$service:L/operation/$operation:L"
```
lucix-aws, line 151: This is not correct -- per awsjson1.0 spec, all HTTP paths are `/`.

lucix-aws: spec for reference: https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html#protocol-behaviors

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+                    req.Header.Set("smithy-protocol", $protocol:S)
```
lucix-aws, line 152: Not used in awsjson10, remove

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+    protected final ProtocolGenerator.GenerationContext ctx;
+    protected final OperationShape operation;
+    protected final GoWriter writer;
+
+    protected final StructureShape input;
+    protected final StructureShape output;
+    private final EventStreamIndex eventStreamIndex;
+
+//    private final String SMITHY_PROTOCOL_NAME = "awsjson10";
+    private final String contentType = "application/awsjson10";
```
lucix-aws, line 53: Incorrect content type for awsjson10

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+                    if !ok {
+                        return out, metadata, $errorf:T("unexpected transport type %T", in.Request)
+                    }
+                
+                    req.Method = $methodPost:T
+                    req.URL.Path = "/service/$service:L/operation/$operation:L"
+                    req.Header.Set("smithy-protocol", $protocol:S)
+                
+                    $contentTypeHeader:W
+                    $acceptHeader:W
```
lucix-aws, line 155: Not used in awsjson10, remove

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+                if err != nil {
+                    return out, metadata, &$error:T{Err: err}
+                }
+
+            
+                payload := $reader:T(jsonEncoder.Bytes())
+                if req, err = req.SetStream(payload); err != nil {
+                    return out, metadata, &$error:T{Err: err}
+                }
+
```
lucix-aws, line 187: Excess spacing, remove

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+                    req.URL.Path = "/service/$service:L/operation/$operation:L"
+                    req.Header.Set("smithy-protocol", $protocol:S)
+                
+                    $contentTypeHeader:W
+                    $acceptHeader:W
+                
+                """,
+        MapUtils.of(
+                "input", ctx.getSymbolProvider().toSymbol(input),
+                "request", generator.getApplicationProtocol().getRequestType(),
```
lucix-aws, line 160: We know this is a smithyhttp.Request since it's awsjson10 - so you can just embed that symbol here. That will also enable you to get rid of the protocol generator being passed into the middleware.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+
+    @Override
+    public void accept(GoWriter writer) {
+        var name = ProtocolGenerator.getSerializeMiddlewareName(operation.getId(), ctx.getService(),
+                generator.getProtocolName());
+        var middleware = createSerializeStepMiddleware(name, ProtocolUtils.OPERATION_SERIALIZER_MIDDLEWARE_ID);
+//        writer.write(middleware.asWritable(generateOperationStubs(), emptyGoTemplate()));
+    }
+
+    public GoWriter.Writable generateOperationStubs() {
```
lucix-aws, line 78: This is doing more than what the name implies (generating "stubs") - you're actually generating the whole middleware, so name it accordingly.

lucix-aws: e.g. "generate", "generateMiddleware", or writer.write() it directly as a Writable.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+                        "output", SmithyGoDependency.SMITHY_MIDDLEWARE.struct("SerializeOutput"),
+                        "metadata", SmithyGoDependency.SMITHY_MIDDLEWARE.struct("Metadata"),
+                        "opName", this.serialName,
+                        "body", generateHandleSerializeBody())
+        );
+    }
+
+    private GoWriter.Writable generateHandleSerializeBody() {
+        return goTemplate("""
+                    $route:W
```
lucix-aws, line 123: [don't have to fix]

Investigate re-writing this middleware as one that extends the base `SerializeRequestMiddleware` - that will enable you to just focus on filling out the routing/payload serialize components of the protocol without having to worry about generating the middleware shell, ID method, etc.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+        MapUtils.of(
+                "input", ctx.getSymbolProvider().toSymbol(input),
+                "request", generator.getApplicationProtocol().getRequestType(),
+                "methodPost", GoStdlibTypes.Net.Http.MethodPost,
+                "service", ctx.getService().getId().getName(),
+                "operation", operation.getId().getName(),
+                "protocol", generator.getProtocol(),
+                "contentTypeHeader", setContentTypeHeader(),
+                "acceptHeader", acceptHeader(),
+                "serialize", handlePayload(),
```
lucix-aws, line 167: Not used in this template, so reading this is somewhat confusing.

lucix-aws: Make sure that you clean up unused arguments in templates.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+                jsonEncoder := $encoder:T()
+            
+                err = $serialize:L(input, jsonEncoder.Value)
+                if err != nil {
+                    return out, metadata, &$error:T{Err: err}
+                }
+
+            
+                payload := $reader:T(jsonEncoder.Bytes())
+                if req, err = req.SetStream(payload); err != nil {
```
lucix-aws, line 184: This is a little confusing to read because of the compound assign-if clause - you have to remember that `req` was created earlier above in the serialize middleware (and in another function in the generator)

Prefer something like

```go
req, err = req.SetStream(payload)
if err != nil {
  // ...
}
```

We generally use assign-ifs to create temporaries that only exist in the scope of that if statement.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+    MapUtils.of(
+            "serialize", getSerializerName(input),
+            "encoder", SmithyGoTypes.Encoding.Json.NewEncoder,
+            "reader", GoStdlibTypes.Bytes.NewReader,
+            "error", SmithyGoTypes.Smithy.SerializationError
+    ));
+    }
+
+    private GoWriter.Writable setContentTypeHeader() {
+        if (input.hasTrait(BackfilledInputOutputTrait.class)) {
```
lucix-aws, line 201: This isn't necessary for awsjson10 - since we put a synthetic empty input struct on each operation that didn't have one, and you're generating the serializers for them and calling them, you're guaranteed to get an empty object `{}` as dictated by the protocol. https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html#empty-body-serialization

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+            return emptyGoTemplate();
+        }
+
+        return goTemplate("""
+                req.Header.Set("Content-Type", $S)
+                """, isInputEventStream() ? EventStreamGenerator.AMZ_CONTENT_TYPE : contentType);
+    }
+
+    private GoWriter.Writable acceptHeader() {
```
lucix-aws, line 210: Remove, since it will be unused.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,223 @@
...
+                        return out, metadata, $errorf:T("unexpected transport type %T", in.Request)
+                    }
+                
+                    req.Method = $methodPost:T
+                    req.URL.Path = "/service/$service:L/operation/$operation:L"
+                    req.Header.Set("smithy-protocol", $protocol:S)
+                
+                    $contentTypeHeader:W
+                    $acceptHeader:W
+
```
lucix-aws, line 156: You're missing `X-Amz-Target` - set that according to https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html#protocol-behaviors

lucix-aws: Note that contentLength shouldn't be necessary, the underlying HTTP client should figure it out for bytes.Readers like we're doing here.

lucix-aws (changes requested):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
```diff
@@ -47,16 +48,29 @@ public ApplicationProtocol getApplicationProtocol() {
...
+        for (var op : ops) {
+            var middle = new SerializeMiddleware(context, op, writer);
+            writer.write(middle.generate());
+            writer.write("\n");
+        }
+        generateSharedSerializers(context, writer, ops);
+    }
+
+    public void generateSharedSerializers(GenerationContext context, GoWriter writer, Set<OperationShape> ops) {
+        Set<Shape> shared = new java.util.HashSet<>(Set.of());
```
Madrigal, line 61: You're creating an unnecessary set.

When you call  `Set.of()` you're creating an empty set. Then, when you call `new HashSet<>()`, you're creating another new set with an empty set as an argument. This is usually done when you want to copy all the elements of set A into set B 

```
Set<String> A = (some values) 
Set<String> B = new HashSet<>(A); // copy all the values already present in A to B
```

But you don't need that here, you just want an empty set. Just call `new HashSet<>()`.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
```diff
@@ -17,23 +17,24 @@
...
 import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
+import static software.amazon.smithy.go.codegen.serde.SerdeUtil.getShapesToSerde;
 
 import java.util.Map;
+import java.util.Set;
 import software.amazon.smithy.aws.traits.protocols.AwsJson1_0Trait;
 import software.amazon.smithy.go.codegen.ApplicationProtocol;
 import software.amazon.smithy.go.codegen.GoWriter;
 import software.amazon.smithy.go.codegen.SmithyGoDependency;
-//import software.amazon.smithy.go.codegen.SmithyGoTypes;
```
Madrigal, line 26: Don't commit commented out code. It's fine when testing locally, but it makes the code harder to follow. Remove the commented imports

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
Madrigal: This is marked as resolved, but the type was still changed. What Luc means is "move it back to how it originally was" (i.e. `var writer`) so it doesn't show up in the diff and we can focus only on the things that have actually changed

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,165 @@
...
+                    req.Header.Set("Content-Type", $contentType:S)
+                    req.Header.Set("X-Amz-Target", $target:S)
+
+                """,
+        MapUtils.of(
+                "input", ctx.getSymbolProvider().toSymbol(input),
+                "request", SMITHY_HTTP_TRANSPORT.pointableSymbol("Request"),
+                "methodPost", GoStdlibTypes.Net.Http.MethodPost,
+                "contentType", contentType,
+                "target", ctx.getService().getId().getName() + operation.getId().getName(),
```
Madrigal, line 135: You may be missing a dot

> The value of this header is the [shape name](https://smithy.io/2.0/spec/model.html#grammar-token-smithy-Identifier) of the service's [Shape ID](https://smithy.io/2.0/spec/model.html#shape-id) joined to the [shape name](https://smithy.io/2.0/spec/model.html#grammar-token-smithy-Identifier) of the operation's [Shape ID](https://smithy.io/2.0/spec/model.html#shape-id), separated by a single period (.) character.
For example, the value for the operation ns.example#MyOp of the service ns.example#MyService is MyService.MyOp.

https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
nathanhit: I can't find this import in my files? Maybe I made a change after committing. Sorry about that.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
nathanhit: Ah got it, sorry I thought that changing it back now would cause a new diff deletion/insert highlight. Changed it back to the original var.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
lucix-aws: This remains unaddressed.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,165 @@
...
+        this.ctx = ctx;
+        this.operation = operation;
+        this.writer = writer;
+
+        this.input = ctx.getModel().expectShape(operation.getInputShape(), StructureShape.class);
+        this.output = ctx.getModel().expectShape(operation.getOutputShape(), StructureShape.class);
+    }
+
+    public GoWriter.Writable generate() {
+        serialName = "awsAwsjson10_serializeOp" + operation.toShapeId().getName();
```
lucix-aws, line 53: Idiomatically speaking we wouldn't do this the way you've done (as an instance variable).

This would generally be calculated on-the-fly through a static method e.g.

```java
public static String getSerializerName(OperationShape op) { /* ... */ }
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,165 @@
...
+
+                """,
+                MapUtils.of(
+                        "route", handleHttp(),
+                        "serialize", handlePayload()
+                ));
+    }
+
+    private GoWriter.Writable handleHttp() {
+        String contentType = "application/x-amz-json-1.0";
```
lucix-aws, line 110: This is a constant, it should be declared as such at the class level e.g.

```java
public static final String CONTENT_TYPE = "...";
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,165 @@
...
+                    $serialize:W
+
+                """,
+                MapUtils.of(
+                        "route", handleHttp(),
+                        "serialize", handlePayload()
+                ));
+    }
+
+    private GoWriter.Writable handleHttp() {
```
lucix-aws, line 109: This name doesn't really make sense or tell us anything.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
```diff
@@ -0,0 +1,165 @@
...
+
+                payload := $reader:T(jsonEncoder.Bytes())
+                req, err = req.SetStream(payload)
+                if err != nil {
+                    return out, metadata, &$error:T{Err: err}
+                }
+
+                in.Request = req
+            """,
+    MapUtils.of(
```
lucix-aws, line 158: JFYI, don't need to address it now, but pay attention to your spacing/indentation as you work. The indentation here is off.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
nathanhit: I seriously can't find this import - am I missing something here? I also checked it in vim and don't see it.

Screenshot 2024-08-08 at 2 56 52 PM

lucix-aws: Ignore me, I was tripped up by the diff. They're 100% not there anymore in the remote.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/SerializeMiddleware.java
nathanhit: Would handleHttpRequest be better? The function error checks the input parameters and request and then sets up the header for the HTTP request for the respective portion of the "Handle Serialize" function - I'm not sure what else to name it.

lucix-aws: That's not really any different, just more verbose.

You've said there what it's doing, just translate that to code. I wouldn't think too hard about it.

"handleProtocolSetup"

"handleRequestPreamble"

"generateRequestPreamble"

"generateProtocolHeaders"

etc.

nathanhit: Ah got it def was overthinking it.

lucix-aws (approved):

Issue #470, opened 2023-11-15
codegen/runtime: SRA observability


lucix-aws:
Implement SRA observability components:
* tracing
* metrics

bwehrle:
Can you specify which metrics API you intend to provide metrics in?  Prometheus, OpenTelemetry?

PR #534, opened 2024-08-29
add tracing APIs and instrumentation

lucix-aws:
Tracing component of #470

* misc
  * adds Java-esque abstract Middleware codegen classes
  * adds the most common stdlib functions we use directly into Writer's context so they can be referenced without having to inject a symbol (e.g. `fmt.Sprintf`)
* Adds `smithy-go/tracing` APIs
* New module: `github.com/aws/smithy-go/tracing/smithy-otel-tracing` adapts an OTEL SDK TracerProvider for use with Smithy clients
* Instrument clients operations with the following span hierarchy:
  * <Service.Operation>
    * Initialize
    * BuildRequest
      * OperationSerializer
    * ResolveAuthScheme
    * GetIdentity
    * ResolveEndpoint
    * ComputePayloadSHA256
    * (note - everything from here down is part of the retry loop, but it is not instrumented that way here because all of the retry logic is SDK-side right now)
      * SignRequest
      * DoHTTPRequest
      * OperationDeserializer
* Adds the following Span attributes
  * `rpc.method` - name of the operation
  * `rpc.service` - sdkID of the service (or shape name if not present)
  * `rpc.system` - hardcoded to `aws-api`
    * Theoretically this would vary by client scope (e.g. non-AWS SDKs would set something different here). I've elected to defer allowing configuration of it for now since it's really just a string, if you disagree let's discuss
  * `error.go.type` - literal go type of an operation's error e.g. `&smithy.GenericAPIError`
  * `error.go.error` - results of `Error()` on an operation's error
  * `error.api.fault` - if error is a smithy API error, the fault type (client/server/unknown)
  * `error.api.code` - if error ", the error code
  * `error.api.message` - if error ", the error message
  * `error` - `true` if operation returned an error
  * `http.proto` - HTTP protocol version (as read from the response)
  * `http.request.content_length` - content length of serialized request body
  * `http.request.method` - HTTP method of serialized request
  * `http.response.status_code` - HTTP status code of raw response
  * `http.response.content_length` - content length of raw response body
  * `operation.resolved_endpoint` - the exact URL returned from endpoint resolution (including the scheme and any base path)
  * `operation.auth.resolved_scheme_id` - the auth scheme ID selected during resolution e.g. aws.auth#sigv4

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import software.amazon.smithy.aws.traits.ServiceTrait;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.go.codegen.auth.AuthSchemeResolverGenerator;
 import software.amazon.smithy.go.codegen.auth.GetIdentityMiddlewareGenerator;
@@ -94,6 +95,7 @@ final class ServiceGenerator implements Runnable {
     private GoWriter.Writable generate() {
         return GoWriter.ChainWritable.of(
                 generateMetadata(),
+                generateObservabilityHelpers(),
                 generateClient(),
                 generateNew(),
                 generateGetOptions(),
@@ -101,6 +103,19 @@ final class ServiceGenerator implements Runnable {
                 generateInputContextFuncs(),
                 generateAddProtocolFinalizerMiddleware()
         ).compose();
+    }
+
+    private GoWriter.Writable generateObservabilityHelpers() {
+        return goTemplate("""
+                func operationTracer(p $tracerProvider:T) $tracer:T {
+                    return p.Tracer($scope:S)
+                }
+                """,
+                Map.of(
+                        "tracerProvider", SmithyGoDependency.SMITHY_TRACING.interfaceSymbol("TracerProvider"),
+                        "tracer", SmithyGoDependency.SMITHY_TRACING.interfaceSymbol("Tracer"),
+                        "scope", settings.getModuleName()
+                ));
     }
 
     private GoWriter.Writable generateMetadata() {
@@ -324,7 +339,7 @@ final class ServiceGenerator implements Runnable {
     @SuppressWarnings("checkstyle:LineLength")
     private GoWriter.Writable generateInvokeOperation() {
         return goTemplate("""
-                func (c *Client) invokeOperation(ctx $context:T, opID string, params interface{}, optFns []func(*Options), stackFns ...func($stack:P, Options) error) (result interface{}, metadata $metadata:T, err error) {
+                func (c *Client) invokeOperation(ctx $context.Context:T, opID string, params interface{}, optFns []func(*Options), stackFns ...func($stack:P, Options) error) (result interface{}, metadata $metadata:T, err error) {
                     ctx = $clearStackValues:T(ctx)
                     $newStack:W
                     options := c.options.Copy()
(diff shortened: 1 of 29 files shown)


wty-Bryant (approved):

tracing/smithy-otel-tracing/adapt.go
```diff
@@ -0,0 +1,186 @@
...
+		return otelattribute.StringSlice(kk, vv), true
+	default:
+		return otelattribute.KeyValue{}, false
+	}
+}
+
+func toOTELKeyValues(props smithy.Properties) []otelattribute.KeyValue {
+	var kvs []otelattribute.KeyValue
+	for k, v := range props.Values() {
+		if kv, ok := toOTELKeyValue(k, v); ok {
```
Madrigal, line 181: I see that you mention that all of these values will be silently ignored, which can make this very hard to debug if things go wrong. Should we at least log this?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -448,4 +497,10 @@ func addProtocolFinalizerMiddlewares(stack $P, options $L, operation string) err
                         SignRequestMiddlewareGenerator.generateAddToProtocolFinalizers()
                 ).compose(false));
     }
+
+    private String getTracingServiceId() {
+        return service.hasTrait(ServiceTrait.class)
+                ? service.expectTrait(ServiceTrait.class).getSdkId().replaceAll("\\s", "")
```
Madrigal, line 503: Interesting, do we really have services with spaces on the sdk id?

Madrigal (approved):
General comment for the spans:

1. Any reason why we picked the middlewares that we picked for when to start a child span? Retry makes a lot of sense, the other ones I suspect are the ones that we know can take longer but curious to hear your thoughts
1. Thanks a lot for actually commenting out the error codes and tracing we emit. We'll likely need to put this on a doc somewhere later, so this would make it a lot easier

tracing/smithy-otel-tracing/adapt.go
lucix-aws: Actually yes, couldn't it just be a span event?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
lucix-aws: Many do -- here's a grep snippet. It's really almost the rule rather than the exception

```
iot-wireless.json                                                                                                                                                [13/1970]
16254:                    "sdkId": "IoT Wireless",                                                                                                                        
                                                                                                                                                                          
connect.json                                                                                                                                                              
1640:                    "sdkId": "Connect",                                                                                                                              
                                                                                                                                                                          
guardduty.json                                                                                                                                                            
6989:                    "sdkId": "GuardDuty",                                                                                                                            
                                                                                                                                                                          
lex-runtime-v2.json                                                                                                                                                       
57:                    "sdkId": "Lex Runtime V2",                                                                                                                         
                                                                                                                                                                          
route-53-domains.json                                                                                                                                                     
4974:                    "sdkId": "Route 53 Domains",                                                                                                                     

mediastore.json
1135:                    "sdkId": "MediaStore",

resource-explorer-2.json
1597:                    "sdkId": "Resource Explorer 2",

elasticache.json
453:                    "sdkId": "ElastiCache",

athena.json
275:                    "sdkId": "Athena",

migration-hub-refactor-spaces.json
3534:                    "sdkId": "Migration Hub Refactor Spaces",

kafkaconnect.json
1856:                    "sdkId": "KafkaConnect",

backup-gateway.json
128:                    "sdkId": "Backup Gateway",
```

lucix-aws:
> General comment for the spans:
> 
> 1. Any reason why we picked the middlewares that we picked for when to start a child span? Retry makes a lot of sense, the other ones I suspect are the ones that we know can take longer but curious to hear your thoughts

Most of them are just the principal "phases" of the operation lifecycle, e.g. resolving endpoint, serializing. There's no real hard list of this anywhere, the closest thing we have to that might be the lifecycle breakdown in the (unpublished) client reference architecture.

The more niche ones were giving `Initialize` a span, and the outer `BuildRequest` (basically everything from serialize to build), I wanted to generally contain those since
1. customizations go there
2. I wanted the top level of root spans to roughly add up to the overall duration. Without the `BuildRequest` one, you'd basically have `OperationSerializer` in an oasis of un-spanned (well root spanned but yes) work.

tracing/smithy-otel-tracing/adapt.go
lucix-aws: Discussed offline, it's probably better to just map to a string representation as default. Pushed.

Madrigal (approved):

wty-Bryant (approved):

PR #535, opened 2024-09-05
Add awsJson10 Error Deserialization support.

nathanhit:
*Issue #, if available:*

*Description of changes:*
Protocol Generator: 

- generateSharedDeserializers contains revisions that attain all error shapes and generate shared deserializers with regular member shapes 
- generateSharedDeserializes generates deserializers specific to each operation and additional error info functions
- generateOperationErrorDeserializers handles all parts of error generation, including cycling through the different operations. There are 4 additional helper functions in protocol generator utilized here. 

Deserialize Middleware:
- Response Checks function calls the error deserializers if a non 200 error is reported

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
@@ -235,7 +235,21 @@ public class AwsJson10ProtocolGenerator implements ProtocolGenerator {
 
                 if jtyp, ok := jv["__type"]; ok {
                     typ = jtyp.(string)
+                } else if jtyp, ok := jv["code"]; ok {
+                    typ = jtyp.(string)
                 }
+                // TODO: Add in Header Check for "x-amzn-errortype"
+
+                hashIndex, colonIndex := 0, len(typ)-1
+                for idx := 0; idx < len(typ); idx++ {
+                    if typ[idx] == '#' && hashIndex == 0 {
+                        hashIndex = idx
+                    }
+                    if typ[idx] == ':' && colonIndex == len(typ)-1 {
+                        colonIndex = idx
+                    }
+                }
+                typ = typ[hashIndex:colonIndex+1]
 
                 if jmsg, ok := jv["message"]; ok {
                     msg = jmsg.(string)
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
@@ -16,19 +16,29 @@
 package software.amazon.smithy.go.codegen.protocol.aws;
 
 import static software.amazon.smithy.go.codegen.ApplicationProtocol.createDefaultHttpApplicationProtocol;
+import static software.amazon.smithy.go.codegen.GoWriter.emptyGoTemplate;
+import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
+import static software.amazon.smithy.go.codegen.protocol.ProtocolUtil.GET_AWS_QUERY_ERROR_CODE;
 import static software.amazon.smithy.go.codegen.serde.SerdeUtil.getShapesToSerde;
+import static software.amazon.smithy.go.codegen.server.protocol.JsonDeserializerGenerator.getDeserializerName;
 
 import java.util.HashSet;
 import java.util.Set;
 import software.amazon.smithy.aws.traits.protocols.AwsJson1_0Trait;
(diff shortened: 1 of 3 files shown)


nathanhit:
Here is the generated code for the deserializers.go file:
[deserializers.txt](https://github.com/user-attachments/files/16900156/deserializers.txt)

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
```diff
@@ -86,10 +97,156 @@ private void generateSharedDeserializers(GenerationContext context, GoWriter wri
...
+    private GoWriter.Writable getProtocolErrorInfo() {
+        return goTemplate("""
+            func getProtocolErrorInfo(payload []byte) (typ, msg string, v $value:T, err error) {
+
+                paid := $reader:T(payload)
+                jsonDecoder := $decoder:T(paid)
+                var val interface{}
+                var jv map[string]interface{}
+
+                jsonDecoder.Decode(&val)
```
Madrigal, line 225: This also returns an error that we should check for

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
```diff
@@ -86,10 +97,156 @@ private void generateSharedDeserializers(GenerationContext context, GoWriter wri
...
+                            "fmtErrorf", GoStdlibTypes.Fmt.Errorf,
+                            "func", ProtocolGenerator.getOperationErrorDeserFunctionName(operation,
+                                    context.getService(), "awsJson10"),
+                            "genericAPIError", SmithyGoDependency.SMITHY.pointableSymbol("GenericAPIError"),
+                            "readAll", SmithyGoDependency.IO.func("ReadAll"),
+                            "smithyhttpResponse", SmithyGoTypes.Transport.Http.Response,
+                            "awsQueryCompatible", context.getService().hasTrait(AwsQueryCompatibleTrait.class)
+                                    ? deserializeAwsQueryError()
+                                    : emptyGoTemplate(),
+                            "errors", GoWriter.ChainWritable.of(
```
Madrigal, line 166: Add this as a variable outside of the map so it's easier to follow, something like this

```java
for(var operation: operations) {
    var errors = context.getService()
        .stream()
        .map(it -> deserializeErrorCase(context, context.getModel().expectShape(it, StructureShape.class)))
        .toList()
}
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
```diff
@@ -86,10 +97,156 @@ private void generateSharedDeserializers(GenerationContext context, GoWriter wri
...
         writer.write(generator.generate(shared));
+
+        generateOperationErrorDeserializers(context, writer, ops);
+
+        writer.write(getProtocolErrorInfo());
+        writer.write(GET_AWS_QUERY_ERROR_CODE);
+    }
+
+    private Set<Shape> generateErrorDeserializers(GenerationContext context, GoWriter writer, Set<OperationShape> ops) {
+        Set<Shape> errorShapes = new HashSet<>();
```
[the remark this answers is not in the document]
nathanhit: Yeah, I found that to be an issue with a lot of the stream operations that are amongst a lot of the different classes, and so I chose to go with the simple format since it's easier to read and understand. If you think the stream might be better and I should replace the code, let me know, but yeah I think it only saves some space in theory.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
nathanhit: good idea

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/aws/AwsJson10ProtocolGenerator.java
nathanhit: I thought that any errors from the "decode" call would have been caught by the jv map error check, and that it didn't matter the type that the decoder would be unloading into, but I suppose that there could also be an error that arises from this that I'm not thinking of. I'll go ahead and add it anyways though.

Madrigal (commented):
Can you add a TODO here mentioning that we don't cover all errors? From [Smithy docs](https://smithy.io/2.0/aws/protocols/aws-json-1_0-protocol.html#operation-error-serialization)

> The component MUST be one of the following: an additional header with the name X-Amzn-Errortype, a body field with the name code, or a body field named __type

We're only covering the 3rd type, body field with `__type`

PR #536, opened 2024-09-13
add metrics APIs and corresponding otel adapter

lucix-aws:
Metrics component of #470 (and downstream https://github.com/aws/aws-sdk-go-v2/issues/1744).

* adds metrics APIs
* adds `smithy-otel-metrics` library which binds a concrete OTEL SDK meter provider to the Smithy version of the API
* adds the following operation metrics as histograms - the units are in seconds per the OTEL spec. all by default have dimensions of (service, method), additional dimensions listed in parentheses where applicable
  * `client.call.duration` - entire elapsed time of call from start to finish
  * `client.call.serialization_duration` - time elapsed to serialize request, including any http bindings
  * `client.call.deserialization_duration`- time elapsed to deserialize request, including http bindings and error deserialization
  * `client.call.resolve_identity_duration` - time elapsed to get credentials (auth scheme id)
  * `client.call.resolve_endpoint_duration` - time elapsed to execute endpoint rules to get an endpoint
  * `client.call.signing_duration` - time elapsed to sign a request (auth scheme id)
* minor tracing adjustment: switches to using package const ServiceID for rpc.service attribute rather than re-deriving in codegen

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
@@ -149,6 +149,8 @@ public abstract class HttpRpcProtocolGenerator implements ProtocolGenerator {
 
             writer.write(goTemplate("""
                     _, span := $T(ctx, "OperationSerializer")
+                    endTimer := startMetricTimer(ctx, "client.call.serialization_duration")
+                    defer endTimer()
                     defer span.End()
                     """, SMITHY_TRACING.func("StartSpan")));
 
@@ -220,6 +222,7 @@ public abstract class HttpRpcProtocolGenerator implements ProtocolGenerator {
             writer.write("in.Request = request");
 
             writer.write("");
+            writer.write("endTimer()");
             writer.write("span.End()");
             writer.write("return next.$L(ctx, in)", generator.getHandleMethodName());
         });
@@ -341,6 +344,7 @@ public abstract class HttpRpcProtocolGenerator implements ProtocolGenerator {
 
             writer.write(goTemplate("""
                     _, span := $T(ctx, "OperationDeserializer")
+                    defer startMetricTimer(ctx, "client.call.deserialization_duration")()
                     defer span.End()
                     """, SMITHY_TRACING.func("StartSpan")));
 
@@ -378,7 +382,6 @@ public abstract class HttpRpcProtocolGenerator implements ProtocolGenerator {
             }
             writer.write("");
 
-            writer.write("span.End()");
             writer.write("return out, metadata, err");
         });
         writer.write("");
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
@@ -344,7 +344,8 @@ public abstract class HttpRpcProtocolGenerator implements ProtocolGenerator {
 
             writer.write(goTemplate("""
                     _, span := $T(ctx, "OperationDeserializer")
-                    defer startMetricTimer(ctx, "client.call.deserialization_duration")()
+                    endTimer := startMetricTimer(ctx, "client.call.deserialization_duration")
+                    defer endTimer()
                     defer span.End()
                     """, SMITHY_TRACING.func("StartSpan")));
 
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationMetricsStruct.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationMetricsStruct.java
@@ -1,0 +1,187 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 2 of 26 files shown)


wty-Bryant (approved):

metrics/metrics.go
```diff
@@ -0,0 +1,136 @@
...
+// Callers use the Observe() API of this interface to report metrics to the
+// underlying collector.
+type Float64Observer interface {
+	Observe(context.Context, float64, ...RecordMetricOption)
+}
+
+// RecordMetricOption applies configuration to a recorded metric.
+type RecordMetricOption func(o *RecordMetricOptions)
+
+// RecordOption represents configuration for a recorded metric.
```
wty-Bryant, line 133: `// RecordMetricOptions `

Madrigal: I'm surprised we don't have something that catches this. Not a blocker for this PR, but we should add some linting to Go files. I definitely don't want to be squinting to see if the comment matches the actual type

lucix-aws: Yeah, basic `golint` should suffice here and would be a trivial add. The SDK already does that, I believe.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationMetricsStruct.java
```diff
@@ -0,0 +1,187 @@
...
+                ) (T, error) {
+                    instr := getOperationMetrics(ctx).histogramFor(metric)
+                    opts = append([]metrics.RecordMetricOption{withOperationMetadata(ctx)}, opts...)
+
+                    start := time.Now()
+                    v, err := fn()
+                    end := time.Now()
+
+                    elapsed := end.Sub(start)
+                    instr.Record(ctx, float64(elapsed)/1e9, opts...)
```
Madrigal, line 156: does the SRA mention what unit should the duration have? Seconds seems a bit too coarse, and ideally I'd like us to just pass a `time.Duration` object without doing math on top of it

metrics/smithy-otel-metrics/async.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (o *int64Observer) Observe(_ context.Context, v int64, opts ...metrics.RecordMetricOption) {
+	o.observer.ObserveInt64(o.instrument, v, withMetricProps(opts...))
+}
+
+// adaptInt64CB wraps an OTEL async instrument callback, binding it to a single
+// int64 instrument.
+func adaptInt64CB(io otelmetric.Int64Observable, cb metrics.Int64Callback) otelmetric.Callback {
+	return func(ctx context.Context, o otelmetric.Observer) error {
+		cb(ctx, &int64Observer{o, io})
```
Madrigal, line 37: I may be blind, but I don't see where `cb` is defined

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationMetricsStruct.java
```diff
@@ -0,0 +1,187 @@
...
+    private GoWriter.Writable generateHelpers() {
+        return goTemplate("""
+                func (m *operationMetrics) histogramFor(name string) metrics.Float64Histogram {
+                    switch name {
+                    case "client.call.duration":
+                        return m.Duration
+                    case "client.call.serialization_duration":
+                        return m.SerializeDuration
+                    case "client.call.resolve_identity_duration":
+                        return m.GetIdentityDuration
```
Madrigal, line 132: nit: shouldn't this also be `ResolveIdentity` ?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
```diff
@@ -378,7 +382,6 @@ private void generateOperationDeserializer(GenerationContext context, OperationS
             }
             writer.write("");
 
-            writer.write("span.End()");
```
Madrigal, line 381: is this expected?

metrics/smithy-otel-metrics/async.go
lucix-aws: image

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
lucix-aws: Yeah - for the deserialize middlewares, the initial defer is enough, because deserialize calls the rest of the middleware chain first.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationMetricsStruct.java
lucix-aws: Discussed offline, this is an OTEL spec mandate.

lucix-aws: But for the record I completely agree. It should just be nanoseconds (most atomic unit) and observability backends should figure it out(tm)

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationMetricsStruct.java
```diff
@@ -0,0 +1,187 @@
...
+                }
+
+                func startMetricTimer(ctx context.Context, metric string, opts ...metrics.RecordMetricOption) func() {
+                    instr := getOperationMetrics(ctx).histogramFor(metric)
+                    opts = append([]metrics.RecordMetricOption{withOperationMetadata(ctx)}, opts...)
+
+                    var ended bool
+                    start := time.Now()
+                    return func() {
+                        if ended {
```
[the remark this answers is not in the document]
lucix-aws: No race conditions, since it's just in the context of a single operation.

The reason this is here is just to make the "stop" call idempotent -- that way you can just have a defer() on it which will catch all of the error cases in the middleware, and a direct call to it when the timed operation actually ends.

metrics/smithy-otel-metrics/async.go
Madrigal: Thanks for showing me my blindness :P

Madrigal (approved):

PR #537, opened 2024-09-18
observability naming adjustments

lucix-aws:
* Remove dashes from single-package modules `smithyoteltracing` and `smithyotelmetrics`. Conventionally, when a module is also meant to be used as a package, it should have a valid package name.
* align span attribute names with internal recommendations
  * I also removed the `api.error_fault` attribute since I don't think it really added any value


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
```diff
@@ -398,14 +398,13 @@ defer endTimer()
...
-                        span.SetProperty("error.go.type", fmt.Sprintf("%T", err))
-                        span.SetProperty("error.go.error", err.Error())
+                        span.SetProperty("exception.type", fmt.Sprintf("%T", err))
+                        span.SetProperty("exception.message", err.Error())
 
                         var aerr smithy.APIError
                         if $errors.As:T(err, &aerr) {
-                            span.SetProperty("error.api.code", aerr.ErrorCode())
-                            span.SetProperty("error.api.message", aerr.ErrorMessage())
-                            span.SetProperty("error.api.fault", aerr.ErrorFault().String())
```
Madrigal, line 408: What's fault in the context of Smithy?

wty-Bryant (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
lucix-aws: "client", "server" or "unknown"

Madrigal: oh, like "who's fault is this". 

I may be biased towards the errors I happen to be looking at right now, but this does seem useful for communicating with customers and saying "this is a client error, not an issue with the server", although I'd defer to your experience if you think this is not important

Madrigal (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ServiceGenerator.java
lucix-aws: Honestly, I'm fine leaving it

PR #539, opened 2024-09-18
Run `golint` and `go vet`

Madrigal:
*Issue #, if available:*

*Description of changes:*
Ran `golint` and `make vet` on the repo.

We had an offline discussion about running this always. `golint` is unfortunately [deprecated](https://github.com/golang/go/issues/38968), and there's no easy way to tell the tool "ignore this module/file"

* Using `find -exec golint` doesn't work as expected since golint expects to be run in the context of a package
* running `golint ./...` marks singleflight as failing check, so we can't use the exit code to accept/reject a change

Looked into incorporating `golangci-lint`, but the scope of changes it flags doesn't seem too valuable, so I'm a bit hesitant to add it to the default run if we're just going to ignore the result.

`golint` flagged these changes, which I

```
golint   ./...
encoding/doc.go:3:1: package comment is detached; there should be no blank lines between it and the package statement
encoding/cbor/encode.go:192:1: comment on exported type EncodeFixedUint should be of the form "EncodeFixedUint ..." (with optional leading article)
encoding/cbor/encode.go:206:1: comment on exported type EncodeFixedNegInt should be of the form "EncodeFixedNegInt ..." (with optional leading article)
encoding/httpbinding/encode.go:36:1: comment on exported function NewEncoderWithRawPath should be of the form "NewEncoderWithRawPath ..."
endpoints/private/rulesfn/doc.go:3:1: package comment is detached; there should be no blank lines between it and the package statement
endpoints/private/rulesfn/strings.go:3:1: comment on exported function SubString should be of the form "SubString ..."
endpoints/private/rulesfn/uri.go:85:2: struct field IsIp should be IsIP
internal/sync/singleflight/docs.go:7:1: package comment is detached; there should be no blank lines between it and the package statement
internal/sync/singleflight/singleflight.go:84:1: error should be the last type when returning multiple items (NOTE this is the one we don't want to fix, since it will break compatibility with the published package)
middleware/stack_values.go:24:1: comment on exported function GetStackValue should be of the form "GetStackValue ..."
middleware/stack_values.go:65:1: receiver name c should be consistent with previous receiver name m for stackValues
private/requestcompression/request_compression.go:18:7: exported const MaxRequestMinCompressSizeBytes should have comment or be unexported
testing/gzip.go:10:1: exported function GzipCompareCompressBytes should have comment or be unexported
testing/xml/doc.go:1:1: package comment should be of the form "Package xml ..."
testing/xml/xmlToStruct.go:12:6: type name will be used as xml.XMLNode by other packages, and that stutters; consider calling this Node
testing/xml/xmlToStruct.go:42:6: func name will be used as xml.XMLToStruct by other packages, and that stutters; consider calling this ToStruct
transport/http/headerlist.go:58:8: should replace j += 1 with j++
transport/http/properties.go:49:1: comment on exported function GetSigV4ASigningRegions should be of the form "GetSigV4ASigningRegions ..."
```

### DIFF
--- a/document/cbor/decode.go
+++ b/document/cbor/decode.go
@@ -48,7 +48,7 @@ func (d *decoder) Decode(v cbor.Value, to interface{}) error {
 
 	rv := reflect.ValueOf(to)
 	if rv.Kind() != reflect.Ptr || rv.IsNil() || !rv.IsValid() {
-		return &document.InvalidUnmarshalError{reflect.TypeOf(to)}
+		return &document.InvalidUnmarshalError{Type: reflect.TypeOf(to)}
 	}
 
 	return d.decode(v, rv, serde.Tag{})
--- a/document/cbor/encode.go
+++ b/document/cbor/encode.go
@@ -145,7 +145,7 @@ func (e *encoder) encodeMap(rv reflect.Value) (cbor.Map, error) {
 	for _, key := range rv.MapKeys() {
 		keyName := fmt.Sprint(key.Interface())
 		if keyName == "" {
-			return nil, &document.InvalidMarshalError{"map key cannot be empty"}
+			return nil, &document.InvalidMarshalError{Message: "map key cannot be empty"}
 		}
 
 		cv, err := e.encode(rv.MapIndex(key), serde.Tag{})
--- a/encoding/cbor/encode.go
+++ b/encoding/cbor/encode.go
@@ -189,7 +189,7 @@ func (v EncodeRaw) encode(p []byte) int {
 	return len(v)
 }
 
-// FixedUint encodes fixed-width Uint values.
+// EncodeFixedUint encodes fixed-width Uint values.
 //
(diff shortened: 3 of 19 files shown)


Madrigal:
API diff failed with

```
# github.com/aws/smithy-go/endpoints/private/rulesfn
## incompatible changes
URL.IsIp: removed
## compatible changes
URL.IsIP: added

# github.com/aws/smithy-go/testing
## incompatible changes
GzipCompareCompressBytes: removed

# github.com/aws/smithy-go/testing/xml
## incompatible changes
XMLNode: removed
XMLToStruct: removed
## compatible changes
Node: added
ToStruct: added
```

I thought these changes were private to the package and not exposed outside, @lucix-aws are they breaking changes?

Madrigal:
Discussed offline, these changes are indeed backwards compatible, this is [an example](https://github.com/aws/aws-sdk-go-v2/issues/2529) where we were bitten by this in the past.

Will revert those changes

PR #543, opened 2024-10-02
add http client metrics

lucix-aws:
Final list of metrics:

 | Metric Name | Unit | Type | Description | Attributes (Dimensions) |
|-------------|------|------|-------------|-------------------------|
| client.http.connections.acquire_duration | s | Histogram | The time it takes a request to acquire a connection |
| client.http.connections.dns_lookup_duration | s | Histogram | The time it takes to do DNS lookup |
| client.http.connections.tls_handshake_duration | s | Histogram | The time it takes to perform a TLS handshake |
| client.http.connections.usage | {connection} | UpDownCounter | State of the connection pool | state: idle/acquired*** |
| client.http.do_request_duration | s | Histogram | Total time spent in Do() |
| client.http.time_to_first_byte | s | Histogram | Time from Do() to first response byte received |

*** We are unable to determine the state of connections right now (more specifically, exactly when a connection is idle isn't known to us). We're adding the dimension of state=acquired for now in case that ever changes.

### DIFF
--- a/transport/http/metrics.go
+++ b/transport/http/metrics.go
@@ -1,0 +1,184 @@
+package http
+
+import (
+	"context"
+	"crypto/tls"
+	"net/http"
+	"net/http/httptrace"
+	"time"
+
+	"github.com/aws/smithy-go/metrics"
+)
+
+var now = time.Now
+
+// withMetrics instruments an HTTP client and context to collect HTTP metrics.
+func withMetrics(parent context.Context, client ClientDo, meter metrics.Meter) (
+	context.Context, ClientDo, error,
+) {
+	hm, err := newHTTPMetrics(meter)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	ctx := httptrace.WithClientTrace(parent, &httptrace.ClientTrace{
+		DNSStart:          hm.DNSStart,
+		ConnectStart:      hm.ConnectStart,
+		TLSHandshakeStart: hm.TLSHandshakeStart,
+
+		GotConn:              hm.GotConn(parent),
+		PutIdleConn:          hm.PutIdleConn(parent),
+		ConnectDone:          hm.ConnectDone(parent),
+		DNSDone:              hm.DNSDone(parent),
+		TLSHandshakeDone:     hm.TLSHandshakeDone(parent),
+		GotFirstResponseByte: hm.GotFirstResponseByte(parent),
+	})
+	return ctx, &timedClientDo{client, hm}, nil
+}
+
+type timedClientDo struct {
+	ClientDo
+	hm *httpMetrics
+}
+
+func (c *timedClientDo) Do(r *http.Request) (*http.Response, error) {
+	c.hm.doStart = now()
+	resp, err := c.ClientDo.Do(r)
+
+	c.hm.DoRequestDuration.Record(r.Context(), elapsed(c.hm.doStart))
+	return resp, err
+}
+
+type httpMetrics struct {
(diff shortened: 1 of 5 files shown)


transport/http/metrics.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	hm.TLSHandshakeDuration, err = meter.Float64Histogram("client.http.connections.tls_handshake_duration", func(o *metrics.InstrumentOptions) {
+		o.UnitLabel = "s"
+		o.Description = "The time it takes an HTTP request to perform the TLS handshake."
+	})
+	if err != nil {
+		return nil, err
+	}
+	hm.ConnectionUsage, err = meter.Int64UpDownCounter("client.http.connections.usage", func(o *metrics.InstrumentOptions) {
+		o.UnitLabel = "{connection}"
```
Madrigal, line 93: nit: can we name this `connections`?

also, probably outside the scope of this PR but this would be nice if it had been named `client.http.connections.total` or something like that, usage doesn't make it clear what kind of usage we're talking about (bandwidth, idle or active connections)

transport/http/metrics.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (m *httpMetrics) GotConn(ctx context.Context) func(httptrace.GotConnInfo) {
+	return func(httptrace.GotConnInfo) {
+		m.addConnAcquired(ctx, 1)
+	}
+}
+
+func (m *httpMetrics) PutIdleConn(ctx context.Context) func(error) {
+	return func(error) {
+		m.addConnAcquired(ctx, -1)
```
Madrigal, line 137: Can we count a connection set as idle as "not acquired"? I'm curious how this plays with `keep-alive` connections

transport/http/metrics.go
```diff
@@ -0,0 +1,184 @@
...
+type timedClientDo struct {
+	ClientDo
+	hm *httpMetrics
+}
+
+func (c *timedClientDo) Do(r *http.Request) (*http.Response, error) {
+	c.hm.doStart = now()
+	resp, err := c.ClientDo.Do(r)
+
+	c.hm.DoRequestDuration.Record(r.Context(), elapsed(c.hm.doStart))
```
Madrigal, line 48: How do we keep track of which request is the one that just ended?

transport/http/metrics.go
lucix-aws: This comes from the SRA list of suggested metrics (which I copied in [here](https://github.com/aws/aws-sdk-go-v2/issues/2809)) - I would rather not deviate where naming has been established.

transport/http/metrics.go
lucix-aws: Not sure I follow - this is all happening inside Do(), which is one request?

transport/http/metrics.go
lucix-aws: The SRA defined this as having one dimension: state = idle | acquired - acquired meaning "in use right now" and idle meaning "not in use" - might have been better to say be active | acquired but again this comes from external recommended naming.

transport/http/metrics.go
Madrigal: Yeah, that's what I was hinting with "outside the scope of this PR". I didn't know the unit was also described there, my bad.

transport/http/metrics.go
Madrigal: Got it, then disregard the comment

Madrigal (approved):

PR #546, opened 2024-10-07
add annotation to package docs when the service shape is deprecated

lucix-aws:


wty-Bryant (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoWriter.java
```diff
@@ -756,6 +756,13 @@ public boolean writePackageShapeDocs(Shape shape) {
                 .map(DocumentationTrait::getValue)
                 .map(docs -> {
                     writePackageDocs(docs);
+                    if (shape.hasTrait(DeprecatedTrait.class)) {
```
Madrigal, line 759: rant: I see what they were trying to do with the shape API returning optional, but this whole setup is just awkward to do on map-streams. I'd rather have this as just imperative code at this point, but not blocking the review on this since the rest of the methods are written on this style

Madrigal (approved):

PR #553, opened 2024-11-13
Fix: URI path element replace issue due to elements name overlap

### LINKED ISSUE #552: Error replacing URI path elements when element names overlap
When replacing URI path elements using the function [`encoding/httpbinding.replacePathElement`](https://github.com/aws/smithy-go/blob/253cd261adbd51728215b44a4ce34583c663d6d7/encoding/httpbinding/path_replace.go#L24), if a later path element's name is a substring of a previous element, e.g. `/{namespace}/{name}`, it will fail with the following error:

> serialization failed: invalid path element, does not contain token stop, /{namespace}/{name}

This does not happen if the substring occurs earlier in the URI, e.g. `/{name}/{namespace}`.

To reproduce, add the following test cases to [`TestPathReplace`](https://github.com/aws/smithy-go/blob/253cd261adbd51728215b44a4ce34583c663d6d7/encoding/httpbinding/path_replace_test.go#L8):

```go
{
	Orig:       []byte("/{namespace}/{name}"),
	ExpPath:    []byte("/{namespace}/value"),
	ExpRawPath: []byte("/{namespace}/value"),
	Key:        "name", Val: "value",
},
{
	Orig:       []byte("/{name}/{namespace}"),
	ExpPath:    []byte("/value/{namespace}"),
	ExpRawPath: []byte("/value/{namespace}"),
	Key:        "name", Val: "value",
},
```

The first case will fail with

> expected no error, got invalid path element, does not contain token stop, /{namespace}/{name}

wty-Bryant:
*Issue #, if available:*
fixes #552 

*Description of changes:*
Old serialization code only indexes first occurrence of target element with left bracket (e.g. `{key`) without considering common prefix cases like `{key1111}` and `{key}`. this pr fixes that by directly searching index of the whole element with the format of `{element}` or `{element+}`

### DIFF
--- a/encoding/httpbinding/path_replace.go
+++ b/encoding/httpbinding/path_replace.go
@@ -39,7 +39,6 @@ func replacePathElement(path, fieldBuf []byte, key, val string, escape bool) ([]
 
 		start = bytes.Index(path, fieldBuf)
 		if start < 0 {
-			// TODO what to do about error?
 			return path, fieldBuf, fmt.Errorf("invalid path index, start=%d. %s", start, path)
 		}
 		encodeSep = false
--- a/encoding/httpbinding/path_replace.go
+++ b/encoding/httpbinding/path_replace.go
@@ -22,6 +22,7 @@ func bufCap(b []byte, n int) []byte {
 // replacePathElement replaces a single element in the path []byte.
 // Escape is used to control whether the value will be escaped using Amazon path escape style.
 func replacePathElement(path, fieldBuf []byte, key, val string, escape bool) ([]byte, []byte, error) {
+	// search for "{<key>}". If not found, search for the greedy version "{<key>+}". If none are found, return error
 	fieldBuf = bufCap(fieldBuf, len(key)+2) // { <key> }
 	fieldBuf = append(fieldBuf, uriTokenStart)
 	fieldBuf = append(fieldBuf, key...)
--- a/encoding/httpbinding/path_replace.go
+++ b/encoding/httpbinding/path_replace.go
@@ -22,32 +22,32 @@ func bufCap(b []byte, n int) []byte {
 // replacePathElement replaces a single element in the path []byte.
 // Escape is used to control whether the value will be escaped using Amazon path escape style.
 func replacePathElement(path, fieldBuf []byte, key, val string, escape bool) ([]byte, []byte, error) {
-	fieldBuf = bufCap(fieldBuf, len(key)+3) // { <key> [+] }
+	fieldBuf = bufCap(fieldBuf, len(key)+2) // { <key> }
 	fieldBuf = append(fieldBuf, uriTokenStart)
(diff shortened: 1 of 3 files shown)


encoding/httpbinding/path_replace.go
```diff
@@ -22,33 +22,33 @@ func bufCap(b []byte, n int) []byte {
 // replacePathElement replaces a single element in the path []byte.
 // Escape is used to control whether the value will be escaped using Amazon path escape style.
 func replacePathElement(path, fieldBuf []byte, key, val string, escape bool) ([]byte, []byte, error) {
-	fieldBuf = bufCap(fieldBuf, len(key)+3) // { <key> [+] }
```
Madrigal, line 25: I think some comments would make this clearer

// search for "{\<key\>}". If not found, search for the greedy version "{\<key\>+}". If none are found, return error

I'm also curious if this is a hot function or why we need to do this whole `bufCap` byte slice checking. This would be easier if we didn't do this weird thing

Madrigal:
> Old serialization code only indexes first occurrence of target element via prefix without considering duplication cases,

true, but to expand on it. The old code only looked for `{key` instead of `{key}` (unsure if optimization or off-by-one error on original implementation). Because of this, `{name` and `{namespace}` both matched `{name}`

Madrigal (approved):

encoding/httpbinding/path_replace.go
wty-Bryant: I'm also confused why they repeatedly use the `bufCap` to do bytes modification, maybe just to save memory in some long path scenario

Madrigal: just to expand, this could actually be hot function if you're calling it every time that you route a request that has an httpbinding. Unfortunately without some benchmark or profiling changing this may cause a regression, so I'm fine just leaving it as-is for now

encoding/httpbinding/path_replace.go
```diff
@@ -22,33 +22,34 @@ func bufCap(b []byte, n int) []byte {
...
+	if start < 0 {
+		fieldBuf = bufCap(fieldBuf, len(key)+3) // { <key> [+] }
+		fieldBuf = append(fieldBuf, uriTokenStart)
+		fieldBuf = append(fieldBuf, key...)
+		fieldBuf = append(fieldBuf, uriTokenSkip)
+		fieldBuf = append(fieldBuf, uriTokenStop)
+
+		start = bytes.Index(path, fieldBuf)
+		if start < 0 {
+			// TODO what to do about error?
```
lucix-aws, line 42: Can we just get rid of this comment?

lucix-aws (commented):
Can we add an error test case? e.g. path is `/{bucket}/{key+}`, try to replace key `buck` with some value, expect error

lucix-aws (approved):

PR #561, opened 2025-01-09
add missing waiter retry breakout on non-nil non-matched error

lucix-aws:
Needed for https://github.com/aws/aws-sdk-go-v2/issues/2937

From [waiter workflow](https://smithy.io/2.0/additional-specs/waiters.html#waiter-workflow) step 4:

> If none of the acceptors are matched and an error was encountered while calling the operation, then transition to the failure state and stop waiting.

Our implementation, for reasons probably lost to time, omits this step, causing waiters to spin longer than they should. This change adds that missing behavior.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters.java
@@ -670,6 +670,7 @@ public class Waiters implements GoIntegration {
                     });
 
                     writer.write("");
+                    writer.write("if err != nil { return false, err }");
                     writer.write("return true, nil");
                 });
     }
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters2.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/Waiters2.java
@@ -649,6 +649,7 @@ public class Waiters2 implements GoIntegration {
                     });
 
                     writer.write("");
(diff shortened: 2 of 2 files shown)


wty-Bryant (approved):

Madrigal:
From https://github.com/aws/aws-sdk-go-v2/issues/2937

The code is currently
```

func objectExistsStateRetryable(ctx context.Context, input *HeadObjectInput, output *HeadObjectOutput, err error) (bool, error) {

	if err == nil {
		return false, nil
	}

	if err != nil {
		var errorType *types.NotFound
		if errors.As(err, &errorType) {
			return true, nil
		}
	}

	return true, nil
}
```

If we add `if err != nil { return false, err }` right before the last statement, wouldn't it contradict the first statement on the waiter?

```
func objectExistsStateRetryable(ctx context.Context, input *HeadObjectInput, output *HeadObjectOutput, err error) (bool, error) {

	if err == nil {
		return false, nil
	}

	if err != nil {
		var errorType *types.NotFound
		if errors.As(err, &errorType) {
			return true, nil
		}
	}
        if err != nil {
                return false, err
        }
        // this is equivalent to `if err == nil` and it was already covered 
        // at the beginning, where we returned "false, nil"
	return true, nil
}
```

Madrigal:
Discussed offline. The first `if err == nil { return false, nil}` is due to the acceptor on the S3 model, where we specify that if no error is found, we don't need to retry. The 2nd equivalent statement `if err == nil {return true, nil}` is due to waiters retrying if there is no error found. 

It still bothers me a bit to see two `ifs` with the same condition but different logic, but it's just an artifact of codegen and nothing to worry about

Madrigal (approved):

PR #565, opened 2025-01-22
coerce nil numbers to 0 in jmespath codegen

lucix-aws:
There's an undocumented quirk in jmespath (as far as i could [find](https://jmespath.org/specification.html)) where `nil` numerics are coerced to 0 for order comparison.

e.g.

image
image
image

This fixes that in codegen. We caught this while attempting to turn on generated jmespath waiters for `autoscaling` in the Go SDK. This will be further verified when those changes are pulled in there.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
@@ -408,25 +408,32 @@ public class GoJmespathExpressionGenerator {
             return goTemplate("$1L := $5L($2L) $4L $5L($3L)", ident, left.ident, right.ident, cmp, cast);
         }
 
-        // undocumented jmespath behavior: null in numeric comparisons coerces to 0
+        // undocumented jmespath behavior: null in numeric _ordering_ comparisons coerces to 0
         // this means the subsequent nil checks for numerics are moot, but it's either this or branch the codegen even
         // further for questionable benefit
         var nilCoerceLeft = emptyGoTemplate();
         var nilCoerceRight = emptyGoTemplate();
-        if (isLPtr && left.shape instanceof NumberShape) {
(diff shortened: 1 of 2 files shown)


codegen/smithy-go-codegen/src/test/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGeneratorTest.java
```diff
@@ -372,9 +376,10 @@ public void testComparatorStringBothNil() {
                 }
                 v4 := input.SimpleShape
                 var v5 bool
+
                 if v2 != nil && v4 != nil {
                     v5 = string(*v2) == string(*v4)
-                }
+                }else { v5 = v2 == nil && v4 == nil }
```
lucix-aws, line 382: Yes, `gofmt` will handle this.

wty-Bryant (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/GoJmespathExpressionGenerator.java
```diff
@@ -420,10 +458,20 @@ private GoWriter.Writable compareVariables(String ident, Variable left, Variable
...
+                        "cast", cast,
+                        "nilCoerceLeft", nilCoerceLeft,
+                        "nilCoerceRight", nilCoerceRight
+                ),
+                Map.of(
+                        "elseCheckPtrs", elseCheckPtrs
                 ));
     }
 
+    private static boolean isOrderComparator(ComparatorType cmp) {
```
Madrigal, line 470: rant (?), non-blocking: I wouldn't know how to name this function so I would get "comparisons excluding direct equality check" without reading inside it. TIL they are technically called "inequalities"[link](https://en.wikipedia.org/wiki/Inequality_(mathematics)) but I'd definitely be thinking `!=` with that name. 

(order comparator is fine BTW)

Madrigal (approved):

PR #566, opened 2025-01-28
Add metric and span support for CBOR serialize/deserialize

Madrigal:
# Description
When doing some testing for CBOR, I realized we were not publishing metrics for serialization and deserialization. This PR fixes that.

Operation metrics are already being generated for CBOR methods as part of [OperationMetricsStruct](https://github.com/aws/smithy-go/blob/d2ad136d4774f0bef7fa710ddc774063ada2aa68/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationMetricsStruct.java#L70), so this just starts using them.

Note that there's no service that handles CBOR currently on the SDK, so this change won't have a corresponding change on the SDK

# Testing
There are currently no methods on the SDK that directly use CBOR as their protocol, so testing was done as follows:
1. Modify a service that supports CBOR `JSON` description to generate CBOR methods
2. Run `make generate`
3. Run `make` to run all tests
4. Make calls. Ensure that both serialize and deserialize metrics are being published

Not sure if we have any unit or integration tests for this, but it would be nice to add them


lucix-aws:
We _do_ have one service - the downstream protocol test - did you check that this regenerates there and compiles / tests OK?

Madrigal:
> We do have one service - the downstream protocol test - did you check that this regenerates there and compiles / tests OK?

Yes, my bad, adding that to the list of things tested (running `make` to test everything under the sun)

What I meant is that I don't think we have a test for "does creating a service with X protocol generates metric Y?".

wty-Bryant (approved):

PR #573, opened 2025-03-05
drop old waiters impl

lucix-aws:
This is now handled by Waiters2, which does generated jmespath instead of runtime.


wty-Bryant (approved):

lucix-aws:
Yes good callout.

Madrigal (approved):

PR #587, opened 2025-06-26
Match CBOR errors via shape name instead of shapeId

Madrigal:
*Issue #, if available:*

*Description of changes:*
There's an ongoing discussion about the right way to match errors on CBOR. This would be one way of checking only against shape name instead of against shapeId. Don't merge this until the discussion is finalized

Diff would look like (`typ` is the full shape id)
```
-     switch string(typ) {
-     case "com.amazonaws.cloudwatch#ResourceNotFound": 
+     // namespace can be mangled by service, so matching by error shape name
+     errorParts := strings.Split(typ, "#")
+     errorName := errorParts[len(errorParts)-1]
+     switch string(errorName) {
+     case "ResourceNotFound": 
```

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/rpc2/cbor/Rpc2CborProtocolGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/rpc2/cbor/Rpc2CborProtocolGenerator.java
@@ -136,7 +136,10 @@ public class Rpc2CborProtocolGenerator extends Rpc2ProtocolGenerator {
                     }
 
                     _ = v
-                    switch string(typ) {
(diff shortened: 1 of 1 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/rpc2/cbor/Rpc2CborProtocolGenerator.java
```diff
@@ -193,6 +197,10 @@ private GoWriter.Writable deserializeErrorCase(GenerationContext ctx, StructureS
                 ));
     }
 
+    private String getShapeName(ShapeId id) {
```
wty-Bryant, line 200: nit: is this new function used anywhere?

PR #590, opened 2025-07-18
add interceptors

lucix-aws:
Adds HTTP interceptors (runtime/codegen).

The behavior and lifecycle of interceptors aren't tested here, they're verified as part of the kitchen sink test service in the downstream SDK.

### DIFF
--- a/transport/http/interceptor_middleware.go
+++ b/transport/http/interceptor_middleware.go
@@ -290,33 +290,6 @@ func (m *InterceptAfterDeserialization) HandleDeserialize(
 	return out, md, err
 }
 
-// InterceptAfterExecution runs AfterExecutionInterceptors.
-type InterceptAfterExecution struct {
-	Interceptors []AfterExecutionInterceptor
-}
-
-// ID identifies the middleware.
-func (m *InterceptAfterExecution) ID() string {
-	return "InterceptAfterExecution"
-}
-
-// HandleInitialize runs the interceptors.
-func (m *InterceptAfterExecution) HandleInitialize(
-	ctx context.Context, in middleware.InitializeInput, next middleware.InitializeHandler,
-) (
-	out middleware.InitializeOutput, md middleware.Metadata, err error,
-) {
-	out, md, err = next.HandleInitialize(ctx, in)
-
-	for _, i := range m.Interceptors {
-		if err := i.AfterExecution(ctx, getIctx(ctx)); err != nil {
-			return out, md, err
-		}
-	}
-
-	return out, md, err
-}
-
 // InterceptAttempt runs AfterAttemptInterceptors.
 type InterceptAttempt struct {
 	BeforeAttempt []BeforeAttemptInterceptor


transport/http/interceptor_middleware.go
```diff
@@ -0,0 +1,352 @@
...
+		if err := i.AfterSigning(ctx, getIctx(ctx)); err != nil {
+			return out, md, err
+		}
+	}
+
+	return next.HandleFinalize(ctx, in)
+}
+
+// InterceptTransmit runs BeforeTransmitInterceptors and AfterTransmitInterceptors.
+type InterceptTransmit struct {
```
Madrigal, line 185: Curious on why some interceptor have before/after while some have their dedicated before and after structs

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationInterceptors.java
```diff
@@ -0,0 +1,114 @@
...
+                                .resolvedFunction(buildPackageSymbol(name))
+                                .useClientOptions()
+                                .build()
+                )
+                .build();
+    }
+
+    @Override
+    public byte getOrder() {
+        return 127;
```
Madrigal, line 39: nit, but do we need this to be the absolute ultimate thing that runs in GoIntegration?

transport/http/interceptor.go
```diff
@@ -0,0 +1,321 @@
...
+// definition for more details.
+type InterceptorContext struct {
+	Input   any
+	Request *Request
+
+	Output   any
+	Response *Response
+}
+
+// InterceptorRegistry holds a list of operation interceptors.
```
Madrigal, line 26: Love these detailed docstrings <3

transport/http/interceptor.go
```diff
@@ -0,0 +1,321 @@
...
+// Available InterceptorContext fields:
+//   - Input
+//   - Request
+//   - Response
+type BeforeDeserializationInterceptor interface {
+	BeforeDeserialization(ctx context.Context, in *InterceptorContext) error
+}
+
+// AfterDeserializationInterceptor runs right after the incoming HTTP response
+// is deserialized. This hook is invoked regardless of whether the deserialized
```
Madrigal, line 282: not against it, but curious what's the use case of calling the hook even when this failed deserialization. This means that Response needs to be checked as well since it can be nil

transport/http/interceptor_middleware.go
lucix-aws: It really just comes down how the stack works. For something like before and after execution, it's at the top of the stack, so you can just do it on both sides of next.Whatever, but for something like serialize, if you did it in one middleware, next.HandleSerialize is technically _everything_ else (like, the entire rest of the operation).

lucix-aws: The attempt ones are another one where the insertion point is just conveniently surrounding the rest of the `next` call chain.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationInterceptors.java
lucix-aws: Preferrably yes so the sandwiching ones are as close to their actual focal points as we can get them.

transport/http/interceptor.go
lucix-aws: That decision was made according to the SRA spec but it's maybe ambiguous:

```
When : This will ALWAYS be called once per attempt , except when a
355 * failure occurs earlier in the request pipeline . The duration
356 * between invocation of this hook and beforeDeserialization is
357 * very close to the amount of time spent unmarshalling the
358 * service response . Depending on the protocol and operation ,
359 * the duration may include the time spent downloading
360 * the response data .
```

I guess in my mind if we returned a service error, deserialization still _completed_, if that makes sense. Unfortunately our middleware structure doesn't distinguish between deserializing an error and an error from a failure to deserialize. If it did I would probably want to call this conditionally based on that difference.

Madrigal: that's a great point

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/OperationInterceptors.java
Madrigal: my gripe is that this means that you can never put anything after this since we used the absolute last value, which is most of the times fine but if we ever want to do something like say a middleware to print which interceptors were registered, we now can't

I know the number is arbitrary, but unless this really needs to be the absolutely last thing I'd prefer this to be something like `120`

transport/http/interceptor_middleware.go
```diff
@@ -0,0 +1,352 @@
...
+		if err := i.AfterDeserialization(ctx, getIctx(ctx)); err != nil {
+			return out, md, err
+		}
+	}
+
+	return out, md, err
+}
+
+// InterceptAfterExecution runs AfterExecutionInterceptors.
+type InterceptAfterExecution struct {
```
wty-Bryant: Do we need that middleware if we already have `InterceptExecution`

transport/http/interceptor.go
```diff
@@ -0,0 +1,321 @@
...
+	copy(s, v)
+	return s
+}
+
+// InterceptorContext is all the information available in different
+// interceptors.
+//
+// Not all information is available in each interceptor, see each interface
+// definition for more details.
+type InterceptorContext struct {
```
wty-Bryant, line 18: Curious about a general question: how do we make sure user only uses available fields in each phase's interceptor? Or we don't control that at all?

lucix-aws: We don't -- they have to refer to the API doc comments on availability.

transport/http/interceptor_middleware.go
lucix-aws: We do not, I'll remove it.

wty-Bryant (approved):

Issue #571, opened 2025-02-27
Add shapes-only codegen


tlindsay42:
Would like to have the option to generate only data shapes similar to the feature in smithy-java described in https://github.com/smithy-lang/smithy/issues/2514 ([example](https://github.com/smithy-lang/smithy-java/tree/0.0.1/examples/standalone-types)) and in the works for smithy-python per https://github.com/smithy-lang/smithy-python/pull/373.

PR #592, opened 2025-08-01
add go-shape-codegen plugin

lucix-aws:
Base plugin for shape codegen, which is basically just client/server codegen without the service/operations. Doesn't affect the generated output of the downstream SDK at all.

Related to #571 but doesn't really close it.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/AbstractDirectedCodegen.java
```diff
@@ -35,6 +37,31 @@
 
 @SmithyInternalApi
 public abstract class AbstractDirectedCodegen implements DirectedCodegen<GoCodegenContext, GoSettings, GoIntegration> {
+    public static void run(PluginContext context, DirectedCodegen<GoCodegenContext, GoSettings, GoIntegration> directedCodegen) {
```
Madrigal, line 40: I was curious why we had this class just laying around. The only implementation of it was the `ServerCodegenPlugin` to generate the server code of a model, which is still a work in progress from what I know

lucix-aws: Correct, server codegen was bootstrapped as a hackathon project in I think 2024.

PR #599, opened 2025-08-26
sort map keys in JSON Document shape encodes

lucix-aws:
We've gotten correspondence about this a handful of times over the last quarter and it appears to be a behavioral requirement for some downstream AWS services in caching scenarios, so I'm just doing it. It also makes unit testing easier.

This DOES NOT cover general map serialization, that will have to be addressed downstream.


Madrigal:
Discussed offline, and to double clarify, this impacts only the encoding of `document` types, not of general serialization of maps, so for example this wouldn't solve aws/aws-sdk-go-v2#3131

Madrigal (approved):
This will have a small performance hit for now having to store the keys on a separate slice and sorting, but since we have to do it anyway there's not much we can improve upon

PR #635, opened 2026-03-06
serde2: use our own json

lucix-aws:
I had benchmarked this after merging it and the old (new old) deserializer performance was pretty bad, basically because of json.Token being an interface. It was some odd ~15% allocation increase across the board.

Thus we're just going to toss that and roll our own.

Dave Cheney has a pretty in-depth article about fast json parsing https://dave.cheney.net/paste/gophercon-sg-2023.html which I've largely followed for this. I didn't make every single optimization that he did though.

Running this against the deserialize stuff in jsonrpc10 we're now at a slight performance boost over the old static version across the board, I've made results available at https://gist.github.com/lucix-aws/ba3ca20113b23752144ad25e0f237a58.

### DIFF
--- a/aws-protocols/awsjson10/awsjson10.go
+++ b/aws-protocols/awsjson10/awsjson10.go
@@ -1,0 +1,191 @@
+package awsjson10
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/aws/smithy-go"
+	awsjson "github.com/aws/smithy-go/aws-protocols/internal/json"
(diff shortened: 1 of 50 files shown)


lucix-aws:
we talked about this offline this morning, we're okay with shipping this if we need to behind the appropriate tests, but we don't want to write off the possibility of using json/v2 (now or in the future) since it seems there are a lot of improvements there.

mostly it is important that we're aware that the current stdlib json solution is a performance regression when applied to our use case and it's something we need to address before shipping

since there's plenty of other work to be done for schema-serde (serde2) I'm going to leave this up for the moment but carry on with implementing the other protocols and event streams, we don't have to make a decision right now

lucix-aws:
closing in favor of #660

Issue #458, opened 2023-10-03
Move protocol generators out of aws-sdk-go-v2 into smithy-go


lucix-aws:

lucix-aws:
Requires completion of #505 as well.

marsheth:
Hey, any update?

lucix-aws:
I'm investigating work on this as of this week. The current intent is to entirely remove the concept of "protocol generation" and instead push that code mostly into the runtime. We hope to gain significant artifact size reduction from doing so with little to no performance cost, which is what I'm prototyping right now. I have no ETA.

PR #646, opened 2026-04-08
implement rpcv2cbor

lucix-aws:
Port `smithy.protocols#rpcv2Cbor` to schema-serde. I elected to rewrite the codec in this instance to get away from the old DOM method, and because reading from bytes better fits the new ShapeDeserializer API.

I realized in doing this that we needed to actually call ReadUnion iteratively to "close" the union structure in a deserialize, so I added a helper like the existing smithy.ReadStruct that does this. I backported that change to the json codec. I also realized that we missed query-compat support for json so I added a TODO. I did implement it for cbor and checked those protocol tests as well.

### DIFF
--- a/aws-protocols/internal/json/shape_deserializer.go
+++ b/aws-protocols/internal/json/shape_deserializer.go
@@ -453,22 +453,48 @@ func (d *ShapeDeserializer) ReadStructMember() (*smithy.Schema, error) {
 	return member, nil
 }
 
+type unionCtx struct {
+	schema *smithy.Schema
+}
+
 // ReadUnion implements [smithy.ShapeDeserializer].
 func (d *ShapeDeserializer) ReadUnion(s *smithy.Schema) (*smithy.Schema, error) {
-	// Decode the entire union object to find the non-null member.
-	var raw map[string]json.RawMessage
-	if err := d.dec.Decode(&raw); err != nil {
-		return nil, fmt.Errorf("decode union: %w", err)
+	if _, ok := d.head.Top().(*unionCtx); !ok {
+		if isNil, err := d.ReadNil(s); isNil || err != nil {
+			return nil, err
+		}
+
+		tok, err := d.token()
+		if err != nil {
+			return nil, err
+		}
+		delim, ok := tok.(json.Delim)
+		if !ok || delim != '{' {
+			return nil, fmt.Errorf("expected '{', got %v", tok)
+		}
+		d.head.Push(&unionCtx{schema: s})
 	}
 
-	for key, val := range raw {
-		if string(val) == "null" {
+	for d.more() {
+		tok, err := d.token()
+		if err != nil {
+			return nil, err
+		}
+		key, ok := tok.(string)
+		if !ok {
+			return nil, fmt.Errorf("expected string key, got %T", tok)
+		}
+
+		// skip null values
+		if isNil, err := d.ReadNil(nil); isNil || err != nil {
+			if err != nil {
+				return nil, err
+			}
 			continue
 		}
 
 		member := s.Member(key)
 		if member == nil && d.opts.UseJSONName {
-			// Try matching by jsonName trait
 			for _, m := range s.Members() {
 				if jn, ok := smithy.SchemaTrait[*traits.JSONName](m); ok && jn.Name == key {
 					member = m
@@ -477,17 +503,17 @@ func (d *ShapeDeserializer) ReadUnion(s *smithy.Schema) (*smithy.Schema, error) 
 			}
 		}
 		if member == nil {
+			if err := d.skip(); err != nil {
+				return nil, err
+			}
 			continue
 		}
 
-		// Replace the decoder with one that reads just this value
-		dec := json.NewDecoder(bytes.NewReader(val))
-		dec.UseNumber()
-		d.dec = dec
 		return member, nil
 	}
 
-	return nil, nil
+	d.head.Pop()
+	return nil, d.expectDelim('}')
 }
 
 // used to skip over a struct member that we didn't have a schema for, though
--- a/aws-protocols/internal/json/shape_deserializer.go
+++ b/aws-protocols/internal/json/shape_deserializer.go
@@ -486,7 +486,12 @@ func (d *ShapeDeserializer) ReadUnion(s *smithy.Schema) (*smithy.Schema, error) 
 		}
 
 		// skip null values
-		if isNil, err := d.ReadNil(nil); isNil || err != nil {
+		isNil, err := d.ReadNil(nil)
+		if err != nil {
+		    return nil, err
+		} else {
+		    continue
+		}
 			if err != nil {
 				return nil, err
 			}
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -301,13 +302,13 @@ final class CodegenVisitor extends ShapeVisitor.Default<Void> {
                     // members initialized before they are used as targets in
                     // AddMember calls. This ensures MapKey(), MapValue(), and
                     // ListMember() are non-nil when the member schema is created.
-                    var shapeIds = new java.util.HashSet<software.amazon.smithy.model.shapes.ShapeId>();
-                   shapes.stream().map(s -> s.getId()).collect(collectors.toSet())
+                    var shapeIds = new HashSet<ShapeId>();
+                    for (Shape s : shapes) {
                         shapeIds.add(s.getId());
                     }
 
-                    var sorted = new java.util.ArrayList<Shape>();
-                    var visited = new java.util.HashSet<software.amazon.smithy.model.shapes.ShapeId>();
+                    var sorted = new ArrayList<Shape>();
+                    var visited = new HashSet<ShapeId>();
                     for (Shape s : shapes) {
                         topoVisit(s, model, shapeIds, visited, sorted);
                     }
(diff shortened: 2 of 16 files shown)


lucix-aws:
relates #458

Madrigal:
> I also realized that we missed query-compat support for json so I added a TODO

do we not have protocol tests for this? Curious if it's something that you just happened to catch

aws-protocols/internal/json/shape_deserializer.go
```diff
@@ -453,22 +453,48 @@ func (d *ShapeDeserializer) ReadStructMember() (*smithy.Schema, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		key, ok := tok.(string)
+		if !ok {
+			return nil, fmt.Errorf("expected string key, got %T", tok)
+		}
+
+		// skip null values
+		if isNil, err := d.ReadNil(nil); isNil || err != nil {
```
Madrigal: nit, but I'd rather just do 
```suggestion
		isNil, err := d.ReadNil(nil)
		if err != nil {
		    return nil, err
		} else {
		    continue
		}
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -297,7 +297,22 @@ void execute() {
                 writer.write("");
                 writer.writeDocs("Initialize schema members after all schemas are declared to avoid initialization cycles");
                 writer.openBlock("func init() {", "}", () -> {
-                    for (Shape shape : shapes) {
+                    // Topologically sort shapes so that target schemas have their
+                    // members initialized before they are used as targets in
+                    // AddMember calls. This ensures MapKey(), MapValue(), and
+                    // ListMember() are non-nil when the member schema is created.
+                    var shapeIds = new java.util.HashSet<software.amazon.smithy.model.shapes.ShapeId>();
```
Madrigal: nit: do we need to call shapes by their full government name here? we already have `import software.amazon.smithy.model.shapes.ShapeId` at the top, same with `HashSet`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -297,7 +297,22 @@ void execute() {
...
+                    for (Shape s : shapes) {
```
Madrigal, line 306: nit: on simple transforms I prefer the stream version

```suggestion
                   shapes.stream().map(s -> s.getId()).collect(collectors.toSet())
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -497,4 +512,24 @@ public Void intEnumShape(IntEnumShape shape) {
...
+            Model model,
+            java.util.Set<ShapeId> inSet,
+            java.util.Set<ShapeId> visited,
+            java.util.List<Shape> result
+    ) {
+        if (visited.contains(shape.getId())) {
+            return;
+        }
+        visited.add(shape.getId());
+        for (var member : shape.members()) {
```
Madrigal, line 528: does the iteration of members have a predetermined order? curious if this can change from execution to execution. And if it doesn't, does it matter?

internal/errors/errors.go
```diff
@@ -0,0 +1,56 @@
+package errors
+
+import (
+	"reflect"
+	"strings"
+
+	"github.com/aws/smithy-go"
+)
+
+// SetErrorCodeOverride sets the ErrorCodeOverride field on err if present.
```
Madrigal, line 10: Java called, it wants its comments back :P

internal/errors/errors.go
```diff
@@ -0,0 +1,56 @@
...
+//
+// This is used by protocols in awsquery-compatible mode that need to override
+// the error code on a deserialized error.
+func SetErrorCodeOverride(err error, code string) {
+	// yes it's reflection but i don't really view errors as a hot path so i
```
Madrigal, line 15: agree

internal/errors/errors.go
```diff
@@ -0,0 +1,56 @@
...
+	// think it's fine
+	v := reflect.ValueOf(err)
+	if v.Kind() == reflect.Ptr {
+		v = v.Elem()
+	}
+	if v.Kind() != reflect.Struct {
+		return
```
Madrigal, line 22: is it valid if this happens? unsure if we should silently fail here (now that we know `panic` will get us a call from the big guys)

smithy-http-protocols/internal/cbor/float16.go
```diff
@@ -0,0 +1,45 @@
+package cbor
+
+func float16to32(f uint16) uint32 {
```
Madrigal, line 3: this is just copied/moved from `smithy-go/encoding/cbor/float16.go` (which is fine, more of a note for reviewers)

smithy-http-protocols/internal/cbor/shape_deserializer.go
```diff
@@ -0,0 +1,785 @@
...
+		if v > math.MaxInt64 {
+			return 0, fmt.Errorf("cbor uint %d exceeds max int64", v)
+		}
+		return int64(v), nil
+	case majorTypeNegInt:
+		v, err := d.readArg()
+		if err != nil {
+			return 0, err
+		}
+		// CBOR negint: actual value is -1 - v
```
Madrigal, line 142: nit: I'd rephrase this to say that the argument is read as positive int and it's us that return it as negative in this case

smithy-http-protocols/internal/cbor/shape_deserializer.go
```diff
@@ -0,0 +1,785 @@
...
+		if v > math.MaxInt64 {
+			return 0, fmt.Errorf("cbor negint exceeds min int64")
+		}
+		return -1 - int64(v), nil
```
Madrigal, line 146: comment: it's kind of weird to me that you have to offset the value by `-1` manually when converting, but that's how the protocol works

smithy-http-protocols/internal/cbor/shape_deserializer.go
```diff
@@ -0,0 +1,785 @@
...
+			return float64(math.Float32frombits(bits)), nil
+		case major7Float64:
+			if d.off+9 > len(d.p) {
+				return 0, errUnexpectedEOF
+			}
+			bits := binary.BigEndian.Uint64(d.p[d.off+1 : d.off+9])
+			d.off += 9
+			return math.Float64frombits(bits), nil
+		default:
+			return 0, fmt.Errorf("expected float, got minor %d", minor)
```
Madrigal: nit: `given majorType7, expected a minor of float type, instead got %d`. Adding as nit since I doubt we'll ever see this in prod

smithy-http-protocols/internal/cbor/shape_deserializer.go
```diff
@@ -0,0 +1,785 @@
...
+}
+
+// ReadStructMember implements [smithy.ShapeDeserializer].
+func (d *ShapeDeserializer) ReadStructMember() (*smithy.Schema, error) {
+	sc := d.head.top()
+	if sc == nil || sc.kind != deserCtxStruct {
+		return nil, fmt.Errorf("ReadStructMember called without ReadStruct")
+	}
+
+	if sc.remaining == -1 {
```
Madrigal, line 549: what does `sc.remaining == -1` means here? more so because I see we check for `<=0` below

smithy-http-protocols/internal/cbor/shape_deserializer.go
```diff
@@ -0,0 +1,785 @@
...
+		if d.off < len(d.p) && d.p[d.off] == 0xff {
+			d.off++
+			d.head.pop()
+			return nil, nil
+		}
+	} else {
+		if sc.remaining <= 0 {
+			d.head.pop()
+			return nil, nil
```
Madrigal, line 558: both of these cases return `nil,nil`, is the only side effect to pop from the head and advance the offset?

smithy-http-protocols/internal/cbor/shape_deserializer.go
```diff
@@ -0,0 +1,785 @@
...
+	}
+	d.head.push(deserCtx{kind: deserCtxMap, remaining: int(count)})
+	return nil
+}
+
+// ReadMapKey implements [smithy.ShapeDeserializer].
+func (d *ShapeDeserializer) ReadMapKey(s *smithy.Schema) (string, bool, error) {
+	mc := d.head.top()
+	if mc == nil || mc.kind != deserCtxMap {
+		return "", false, fmt.Errorf("ReadMapKey called without ReadMap")
```
Madrigal: isn't this what causes linter errors because we're calling `errorf` without a format string?

smithy-http-protocols/rpcv2/rpcv2.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/aws/smithy-go"
+	"github.com/aws/smithy-go/eventstream"
+	internalerrors "github.com/aws/smithy-go/internal/errors"
+	smithyio "github.com/aws/smithy-go/io"
+	"github.com/aws/smithy-go/middleware"
+	internalcbor "github.com/aws/smithy-go/smithy-http-protocols/internal/cbor"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+)
+
+// Protocol implements an RPC v2 protocol.
```
Madrigal, line 19: nit: can we add a link to smithy docs?

internal/errors/errors.go
lucix-aws: So we technically control the call site, and in that context we passed this an instance of smithy.DeserializableError which is going to be a struct. But for safeguard purposes if we are reflectionificating something I would say just liberally check "is this safe to do, if not don't do it" to avoid a panic.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
lucix-aws: I am not sure the order is guaranteed but in this context it doesn't matter because we're just trying to sort shapes by dependency graph, so member order doesn't really matter.

smithy-http-protocols/internal/cbor/shape_deserializer.go
lucix-aws: nil, nil is expected when the struct runs out of members, and then the pop just throws the list context away

smithy-http-protocols/internal/cbor/shape_deserializer.go
lucix-aws: remaining -1 is used as a sentinel to say "this was an indefinite cbor list" which is why in the followup we're looking for the terminator byte 0xff

Madrigal (approved):

PR #649, opened 2026-04-17
check @enum on strings for cbor

lucix-aws:
we need to handle both EnumShape and StringShape + @enum (latter is deprecated).

the old json code already does this, https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/DocumentMemberSerVisitor.java#L177-L184, cbor never got it.

note to self to check this for the schema-serde stuff as well

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde/cbor/CborSerializerGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde/cbor/CborSerializerGenerator.java
@@ -42,6 +42,7 @@ import software.amazon.smithy.model.shapes.ShapeType;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.shapes.TimestampShape;
 import software.amazon.smithy.model.shapes.UnionShape;
+import software.amazon.smithy.model.traits.EnumTrait;
 import software.amazon.smithy.model.traits.StreamingTrait;
 import software.amazon.smithy.utils.MapUtils;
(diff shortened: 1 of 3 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde/cbor/CborSerializerGenerator.java
```diff
@@ -223,9 +226,11 @@ private Writable generateSerializeField(MemberShape member) {
             case BLOB, LIST, SET, MAP, STRUCTURE, UNION ->
                     serializeNilableMember(member, target, false);
             case STRING ->
-                    isPointable(symbol)
-                            ? serializeNilableMember(member, target, true)
-                            : serializeString(member, target);
+                    target.hasTrait(EnumTrait.class)
```
wty-Bryant, line 229: nit: all enum fields I have ever seen is just string value type, considering enum type also falls into `serializeString` below, do we need to check enum trait here?

wty-Bryant (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde/cbor/CborSerializerGenerator.java
lucix-aws: it's either EnumShape _or_ StringShape + @enum. so we should be checking both now unless I'm missing something

Issue #651, opened 2026-04-23
cache/lru: Putting entries with the same key leaks memory


ella-chao:
[Putting](https://github.com/aws/smithy-go/blob/main/container/private/cache/lru/lru.go#L47-L58) an entry with an existing key into the LRU cache overwrites the `entries` map with the new entry but does not handle the old entry with the same key in the `mru` list. If the number of unique keys is small, `evict` won't run because the number of entries in the map can stay under the capacity when the list accumulates.

The following is a heap profile shown in terms of inuse objects, and we can see millions of objects allocated by the LRU.

<img width="1914" height="948" alt="Image" src="https://github.com/user-attachments/assets/9b3a765b-dfe3-476b-87d1-669ca32f133d" />

lucix-aws:
excellent find, see linked pr

ella-chao:
Thank you for the quick fix!

PR #652, opened 2026-04-23
fix lrucache memory leak on existing item put

lucix-aws:
The cache needs to check for existing entries and just update that value if so (and move it to the front).


lucix-aws:
closes #651

wty-Bryant (approved):

Madrigal (approved):

PR #653, opened 2026-04-23
implement awsquery/ec2query

lucix-aws:
awsquery and ec2query protocols

The combination of query serializer and xml deserializer are probably the most complicated these will get (due to query's prefixing and xml traits/flattening).

`@xmlAttribute` isn't implemented yet, that will be added in the next PR for restxml.

### DIFF
--- a/aws-protocols/internal/query/shape_deserializer.go
+++ b/aws-protocols/internal/query/shape_deserializer.go
@@ -1,0 +1,4 @@
+package query
+
+// ShapeDeserializer doesn't exist for query. The awsquery and ec2query
+// protocols use XML responses.
--- a/aws-protocols/internal/query/shape_serializer.go
+++ b/aws-protocols/internal/query/shape_serializer.go
@@ -23,6 +23,8 @@ const (
 	ctxKindMap
 	ctxKindMapValue
 	ctxKindStruct
+
+	ctxKindNone ctxKind = 0xff
 )
 
 type serCtx struct {
@@ -109,7 +111,7 @@ func (s *ShapeSerializer) Bytes() []byte {
 
 func (s *ShapeSerializer) top() ctxKind {
 	if len(s.stack) == 0 {
-		return 0xff
+		return ctxKindNone
 	}
 	return s.stack[len(s.stack)-1].kind
 }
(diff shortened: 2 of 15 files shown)


lucix-aws:
relates #458

aws-protocols/internal/query/shape_deserializer.go
```diff
@@ -0,0 +1,4 @@
+package query
+
+// ShapeDeserializer doesn't exist for query. The awsquery and ec2query
```
Madrigal, line 3: praise: Love this empty places with comments that just explain why something that you'd expect to find here is not here

aws-protocols/internal/query/shape_serializer.go
```diff
@@ -0,0 +1,641 @@
...
+	keys := make([]string, 0, len(s.values))
+	for k := range s.values {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var buf strings.Builder
+	for _, k := range keys {
+		for _, v := range s.values[k] {
+			if buf.Len() > 0 {
```
Madrigal, line 101: praise: nice solution for adding `&` only after the first occurrence

aws-protocols/internal/query/shape_serializer.go
```diff
@@ -0,0 +1,641 @@
...
+			buf.WriteByte('=')
+			buf.WriteString(url.QueryEscape(v))
+		}
+	}
+	return []byte(buf.String())
+}
+
+func (s *ShapeSerializer) top() ctxKind {
+	if len(s.stack) == 0 {
+		return 0xff
```
Madrigal: if we're already using types, can this be defined as a constant?

aws-protocols/internal/query/shape_serializer.go
```diff
@@ -0,0 +1,641 @@
...
+		entry := &ctx.mapBuf[len(ctx.mapBuf)-1]
+		suffix := strings.TrimPrefix(key, entry.prefix)
+		entry.values = append(entry.values, kv{k: suffix, v: value})
+		return true
+	}
+
+	return false
+}
+
+// WriteInt8Ptr implements [smithy.ShapeSerializer].
```
Madrigal, line 229: Don't we have this clear as part of the interface check on line 74?

aws-protocols/internal/query/shape_serializer.go
```diff
@@ -0,0 +1,641 @@
...
+
+	if v {
+		s.writeValue(s.resolveKey(schema), "true")
+	} else {
+		s.writeValue(s.resolveKey(schema), "false")
+	}
+}
+
+// WriteInt8 implements [smithy.ShapeSerializer].
+func (s *ShapeSerializer) WriteInt8(schema *smithy.Schema, v int8) {
```
Madrigal: I remember on previous implementations we were able to have these `intX` be generic, can we do the same here?

aws-protocols/internal/query/shape_serializer.go
```diff
@@ -0,0 +1,641 @@
...
+	}
+	if s.top() != ctxKindMap {
+		return
+	}
+
+	ctx := s.pop()
+
+	// Sort entries by map key for deterministic output.
+	sort.Slice(ctx.mapBuf, func(i, j int) bool {
+		return ctx.mapBuf[i].key < ctx.mapBuf[j].key
```
Madrigal, line 563: I know sorting of strings can be tricky when doing plan comparisons, I assume this is what we had before and the spec doesn't mention any special treatment we need to do to strings and it's fine to treat them by their underlying bytes?

aws-protocols/internal/query/shape_serializer.go
```diff
@@ -0,0 +1,641 @@
...
+	if t, ok := smithy.SchemaTrait[*traits.EC2QueryName](schema); ok {
+		return t.Name
+	}
+	if xn, ok := smithy.SchemaTrait[*traits.XMLName](schema); ok {
+		return capitalize(xn.Name)
+	}
+	return capitalize(schema.MemberName())
+}
+
+func capitalize(s string) string {
```
Madrigal, line 609: is it fine to handle this just with ASCII input?

aws-protocols/internal/xml/error.go
```diff
@@ -0,0 +1,40 @@
...
+	c, err := ExtractElement(errorBody, "Code")
+	if err != nil && err != io.EOF {
+		return
+	}
+	m, err := ExtractElement(errorBody, "Message")
+	if err != nil && err != io.EOF {
+		return
+	}
+
+	err = nil
```
Madrigal, line 33: nit, but I had to read this twice
```suggestion
    // clear err if it was io.EOF
	err = nil
```

aws-protocols/internal/xml/extract.go
```diff
@@ -0,0 +1,72 @@
...
+//	<XmlListsResponse>
+//	 	<XmlListsResult>
+//	 		<member1>foo</member1>
+//	 		...
+//	 	</XmlListsResult>
+//	</XmlListsResponse>
+//
+// ExtractElement gives you "<member1>foo</member1>...".
+//
+// ExtractElement will forward the io.EOF returned by the underlying decoder if
```
Madrigal, line 26: why forward the `io.EOF` instead of having the caller just check `len(res)`?

aws-protocols/internal/xml/shape_deserializer.go
```diff
@@ -0,0 +1,686 @@
...
+	flattened bool
+
+	// for flattened list/map, ReadStructMember consumes the start element
+	// when it discovers the member, subsequent calls to ReadMapKey or
+	// ReadListItem will do that instead
+	first bool
+
+	// for map, ReadX after ReadMapKey will leave the stream at </value>
+	// which the next call to ReadMapKey must consume
+	inEntry bool
```
Madrigal: nit: 

```suggestion
	inMapEntry bool
```

Madrigal (approved):
LGTM, minor comments

aws-protocols/internal/query/shape_serializer.go
lucix-aws: you have to have // [StructName] <...> on everything exported otherwise the linter gets upset

aws-protocols/internal/query/shape_serializer.go
lucix-aws: it would appear so per https://smithy.io/2.0/spec/model.html#shape-id-abnf

aws-protocols/internal/query/shape_serializer.go
lucix-aws: yeah, query did this although directly with sort.Strings https://github.com/aws/aws-sdk-go-v2/blob/main/service/sns/serializers.go#L2985-L3001 but the comparison is the same

aws-protocols/internal/xml/extract.go
lucix-aws: no particular reason, i just picked that as a sentinel, tbh the true idiom would probably be to return []byte, bool, error but this is internal code so no huge pref

PR #655, opened 2026-04-30
switch to explicit WriteStruct

lucix-aws:
Prior to this change we were overloading WriteMap/CloseMap to open and close structs. That assumes that map and struct are serialized the exact same way in every protocol/codec which isn't the case. Evidently I ran into this and noted it in the past, based on the TODO comment in the source that is removed, but I ran into it again with XML.

Thus we add WriteStruct/CloseStruct to remove the overload.

This also means we need to add a "serialize my fields" onto structs, which is what SerializeMembers is for.

Adjustments to existing codecs were necessary to keep things working, protocol tests for all migrated codecs are still passing w/ this change.

### aws-protocols / smithy-http-protocols

I have elected to un-modulize these for now. I question the real value of separating them (versus our other existing separated modules - signing and the observability ones - which have actual reasons to be that way). It also adds complexity to the codegen process and in general has made developing this more of a pain than it could've been.

### DIFF
--- a/eventstream/serializer.go
+++ b/eventstream/serializer.go
@@ -191,8 +191,11 @@ func (s *ShapeSerializer) WriteTimePtr(schema *smithy.Schema, v *time.Time) {
 }
 
 // WriteStruct implements [smithy.ShapeSerializer].
-func (s *ShapeSerializer) WriteStruct(schema *smithy.Schema, v smithy.Serializable) {
-	v.Serialize(s)
+func (s *ShapeSerializer) WriteStruct(schema *smithy.Schema) {
+}
+
+// CloseStruct implements [smithy.ShapeSerializer].
+func (s *ShapeSerializer) CloseStruct() {
 }
 
 // WriteUnion implements [smithy.ShapeSerializer].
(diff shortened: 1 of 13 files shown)


Madrigal (approved):

eventstream/serializer.go
```diff
@@ -191,8 +191,11 @@ func (s *ShapeSerializer) WriteTimePtr(schema *smithy.Schema, v *time.Time) {
 }
 
 // WriteStruct implements [smithy.ShapeSerializer].
-func (s *ShapeSerializer) WriteStruct(schema *smithy.Schema, v smithy.Serializable) {
-	v.Serialize(s)
```
wty-Bryant, line 195: so previous `Serializable.Serialize(s)`  workflow now happens within `Serializable.SerializeMembers(serializer)` correct?

wty-Bryant (approved):

eventstream/serializer.go
lucix-aws: yes, generated Serialize()s now basically look like

```
WriteStruct()
myself.SerializeMembers()
CloseStruct()
```

PR #656, opened 2026-05-01
implement restxml

lucix-aws:
This supports both protocol tests, restxml + restxmlwithnamespace, because it's just one protocol trait underneath.

I had to rework the decoder a bit to operate on full xml instead of inner, since we need to grab any xmlAttributes.

This PR also adds the concept of "direct traits" (for members, traits that were applied specifically to a member) and a corresponding SchemaDirectTrait, because that matters for xmlName and xmlNamespace.

### DIFF
--- a/aws-protocols/internal/xml/extract.go
+++ b/aws-protocols/internal/xml/extract.go
@@ -7,12 +7,15 @@ import (
 )
 
 // ExtractElement finds the first occurrence of the named element in the XML
-// and returns its INNER content as raw bytes.
+// and returns its content as raw bytes. If preserveEnclosing is true, the
+// matched element's opening and closing tags (with attributes) are included
+// in the result; otherwise only the inner content is returned.
 //
 // This is used to extract both success and error responses from the protocol
 // XML that wraps them.
 //
-// e.g. for awsquery, with an operation output shape named XmlListsResult:
+// For example for awsquery, with an operation output shape named
+// XmlListsResult:
 //
 //	<XmlListsResponse>
 //	 	<XmlListsResult>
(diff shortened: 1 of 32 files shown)


aws-protocols/internal/xml/extract.go
```diff
@@ -21,24 +24,64 @@ import (
...
+//	<member1>foo</member1>...
+//
+// ExtractElement(payload, "XmlListsResult", true) yields:
+//
+//	<XmlListsResult><member1>foo</member1>...</XmlListsResult>
+//
+// ExtractElement will forward the io.EOF returned by the underlying decoder
+// if the target element is not found, which the caller can index on if
+// they're looking for an optional element.
+func ExtractElement(payload []byte, name string, preserveEnclosing bool) ([]byte, error) {
```
Madrigal: I really hate `DoStuff(arg1, true)` functions, since you always need to look into the actual implementation to figure out what that boolean flag does. While a bit more verbose, could we instead do

```go
func ExtractElementInner(payload []byte, name string) ([]byte, error)
func ExtractElementIOuter(payload []byte, name string) ([]byte, error)

func extractElement(payload []byte, name string, preserveEnclosing bool) ([]byte, error) 
```

this matches the inner/outer operations on JS https://developer.mozilla.org/en-US/docs/Web/API/Element/outerHTML or .NET https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlnode.innerxml?view=net-10.0

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SchemaGenerator.java
```diff
@@ -43,9 +45,22 @@ public static String getMemberSchemaName(Shape shape, MemberShape member, Servic
         return String.format("%s_%s", getSchemaName(shape, service), member.getMemberName());
     }
 
+    // modelShapeID returns the original model shape ID, preferring the
+    // Synthetic trait's archetype (if any) so that synthetic clones of Input /
+    // Output inline shapes carry the Smithy model's shape ID rather than the
+    // Go-codegen synthetic ID.
```
Madrigal, line 55: where we having any protocol tests failures because of this?

lucix-aws:
relates #458

aws-protocols/internal/xml/extract.go
lucix-aws: i ended up just dropping the boolean entirely, the error handler does the stdlib Unmarshal now like json

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/SchemaGenerator.java
lucix-aws: yes, for an input we generate smithy.go.synthetic#FooInput (always ending in Input), if the model had FooRequest or something like that, we put the wrong thing in the opening tag.

aws-protocols/internal/xml/extract.go
Madrigal: 🙌

Madrigal (approved):

PR #657, opened 2026-05-01
serde2: misc fixes

lucix-aws:
* members in services targeting prelude shapes were causing duped primtiive schemas to get generated - added support for those in the runtime which the generator now references
* we still need to handle the old idl1 convention of string shape + `@enum` trait (see #649) everywhere

every downstream service now compiles w/ schema-serde _except_
* s3 (shocker) because of some custom deserialization middleware
* sagemakerruntimehttp2 because it's in a weird state from not generating any operations

i don't expect schema-serde to come online for either of those any time soon, they will likely be the last, so i've just left those alone for now


lucix-aws:
relates #458

wty-Bryant (approved):

PR #658, opened 2026-05-11
make traits indexable

lucix-aws:
Representing traits internally w/ a slice instead of a map gives constant-time lookup.

The artifact size increase is not all that much, either. Across all services:
* the map approach at runtime was ~5.2mB (assuming 48 bytes for the internal hmap map header and 40 per entry)
* the slice approach is ~6.5 MiB (the bitmask for tracking directs is a non-trivial save, without it and instead having a sort of []traitEntry thing it was at like 9MiB and then with two slices it was probably worse).

I selected the indices based on frequency across our models. Without the frequency sort we occupy ~800k slots (16 bytes per slot) and with it we occupy ~310k.


lucix-aws:
relates #458

schema.go
```diff
@@ -64,29 +63,48 @@ func stoid(s string) ShapeID {
...
-	id           ShapeID
-	typ          ShapeType
-	members      map[string]*Schema // member name -> schema
-	traits       map[string]Trait   // trait ID -> trait (effective view; for members, target's traits merged with member's overrides)
-	directTraits map[string]Trait   // trait ID -> trait (member schemas only: only traits declared directly on the member)
-	targetID     ShapeID            // for member schemas, the target's shape ID
+	id         ShapeID
+	typ        ShapeType
+	members    map[string]*Schema // member name -> schema
+	traits     map[string]Trait   // trait ID -> non-indexed traits only
```
wty-Bryant: I see `AddMember` and `addTrait` now add (indexed) traits from target schema and traits slice at the same time, does that mean `traits` and `indexed` now refer to the same set of traits? (ignoring zero slot in indexed)

wty-Bryant (approved):

schema.go
lucix-aws: `traits` is _only_ traits that aren't defined by us (that don't implement IndexableTrait). in practice this will always be empty unless someone is generating a client using smithy-go _and_ registering their own traits for whatever reason.

PR #659, opened 2026-05-11
serde2 lightning round

lucix-aws:
* normalize protocol interface
  * generate ServiceSchema like we do for operations, this allows us to get rid of a context lookup as well
  * all protocol constructors now take options, though they are unused atm
  * i've removed the "Codec" concept for now, I'm not sure what the story is on customers actually providing their own Codecs, since ours should be more or less the most performant possible in the context of generated code. They can still override the protocol if they wanted to. We can very easily expose Codec on protocol options in the future if we chose to.
* do querycompatible in jsonrpc
* trait and protocol IDs are now ShapeID
* awsjson10 -> awsjson (since it exposes both 10 and 11)
* dedupe the serde stack code that everything uses
* some latent protocol test failures around httpbinding:
  * deserializers we re-implementing the read in ReadXPtr outside of bindings, it needs to delegate to the body codec
  * missed a payload case in reading strings


lucix-aws:
relates #458

Madrigal:
>  I'm not sure what the story is on customers actually providing their own Codecs, since ours should be more or less the most performant possible in the context of generated code. 

General comment, but I'm not sure if this is true. We use standard library decoding which has shown to not be the most performant. We now have a different story for it which may make it faster, but we're still using stdlib components. I agree that we may not want to expose this until we have a better story for it, but I'd phrase it more for "let's wait to see what people want" rather than saying "we will always have the most performant thing".

aws-protocols/internal/query/shape_serializer.go
```diff
@@ -46,7 +47,7 @@ type ShapeSerializer struct {
 	opts ShapeSerializerOptions
 
 	values     url.Values
-	stack      []serCtx
+	stack      serde.Stack[serCtx]
```
Madrigal, line 50: praise: I prefer this style

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde2/StructureDeserializer.java
```diff
@@ -30,6 +31,9 @@ public StructureDeserializer(GoCodegenContext ctx, StructureShape shape) {
     @Override
     public void accept(GoWriter writer) {
         writer.addImport(ctx.settings().getModuleName() + "/schemas", "schemas");
+        if (Prelude.isPublicPreludeShape(shape)) {
+            writer.addUseImports(SmithyGoDependency.SMITHY_PRELUDE);
```
Madrigal, line 35: Why do we need to add an import for prelude? I thought these shapes were available always without imports

internal/serde/stack.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Push adds a value to the top of the stack.
+func (s *Stack[T]) Push(v T) {
+	s.values = append(s.values, v)
+}
+
+// Pop removes and returns the top value.
+func (s *Stack[T]) Pop() T {
+	v := s.values[len(s.values)-1]
+	var zero T
```
Madrigal, line 22: is this to prevent memory leaks?

lucix-aws:
I meant in the post-serde2 world, where our json and cbor are purpose-built for our implementations and should basically be as good as it gets. For XML technically we are handwaving that and just using the "less-performant" stdlib impl since AWS services are getting away from it. But someone might care.

I mostly am interested in not exposing codec config on the initial release just because I like being minimal with API surface changes, and someone technically can still override the serde if they just roll their own protocol impl but ofc they'd have to dupe some stuff.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde2/StructureDeserializer.java
Madrigal: Discussed offline. We added prelude shapes in #657 to prevent us from generating duplicate shape when the service was declaring their own shapes with a conflicting name, like

```
var String = schema(com.service#String)
var String = schema(smithy.api#String)
```

We define the prelude shapes in the new `prelude` package in Go, which mirrors the [prelude shapes in Smithy](https://smithy.io/2.0/spec/model.html#prelude)

These are not generated from Smithy, but Smithy adding new shapes to the prelude is likely going to be a very rare event, if ever

Madrigal (approved):

internal/serde/stack.go
lucix-aws: yes, i only caught onto this recently, towards the top of https://go.dev/wiki/SliceTricks it talks about this. for scalars it doesn't matter but with generics that could be pointers it would.

PR #661, opened 2026-05-14
add event stream test generator

lucix-aws:
Adds support for generating `@eventStreamTest`.

We are skipping some of these tests because we apparently do not behave exactly on spec, see the SKIP_TESTS list.

Note that I'm doing this against main and not feat-serde2. The intent was to verify against main and then we can pull this into serde2 etc.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
@@ -111,6 +111,17 @@ public final class ShapeValueGenerator {
     }
 
     /**
+     * Writes a non-pointable structure shape value inline (no leading &amp;).
+     */
+    public void writeStructureShapeValueInline(GoWriter writer, StructureShape shape, Node params) {
+        Symbol symbol = symbolProvider.toSymbol(shape);
+        writer.write("$T{", symbol);
+        params.accept(new ShapeValueNodeVisitor(writer, this, shape, ListUtils.copyOf(shape.getAllTraits().values()),
+                config));
+        writer.writeInline("}");
+    }
+
+    /**
      * Writes generation of a member shape value type declaration for the given the parameters.
      *
      * @param writer writer to write generated code with.
@@ -709,12 +720,21 @@ public final class ShapeValueGenerator {
         public Void stringNode(StringNode node) {
             switch (currentShape.getType()) {
                 case BLOB:
+                    writer.writeInline("$S", node.getValue());
+                    break;
                 case STRING:
                     writer.writeInline("$S", node.getValue());
                     break;
 
                 case ENUM:
                     writer.writeInline("$S", node.getValue());
+                    break;
+
+                case TIMESTAMP:
+                    writer.addUseImports(SmithyGoDependency.TIME);
+                    writer.writeInline(
+                            "func() time.Time { t, _ := time.Parse(time.RFC3339, $S); return t }()",
+                            node.getValue());
                     break;
 
                 case BIG_INTEGER:
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
@@ -46,12 +46,9 @@ public class EventStreamProtocolTestGenerator {
             "MissingRequiredInitialResponseOutput",
             "DuplexMissingRequiredInitialResponseOutput",
 
-            // SDK silently drops unrecognized/malformed event stream messages
-            // instead of surfacing an error.
-            "ClientUnexpectedErrorOutput",
-            "DuplexClientUnexpectedErrorOutput",
-            "MissingEventTypeOutput",
-            "DuplexMissingEventTypeOutput",
+            // SDK does not validate the type of the :event-type header. A blob
+            // value is treated as an unknown event (UnknownUnionMember) rather
+            // than surfacing an error.
             "MalformedEventTypeOutput",
             "DuplexMalformedEventTypeOutput",
 
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
@@ -1,0 +1,658 @@
+package software.amazon.smithy.go.codegen.integration;
+
+import java.util.Base64;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.codegen.core.SymbolProvider;
+import software.amazon.smithy.go.codegen.GoCodegenContext;
+import software.amazon.smithy.go.codegen.GoDelegator;
+import software.amazon.smithy.go.codegen.GoSettings;
+import software.amazon.smithy.go.codegen.GoWriter;
+import software.amazon.smithy.go.codegen.ShapeValueGenerator;
+import software.amazon.smithy.go.codegen.SmithyGoDependency;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.EventStreamIndex;
+import software.amazon.smithy.model.knowledge.EventStreamInfo;
+import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.node.ObjectNode;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.shapes.UnionShape;
+import software.amazon.smithy.protocoltests.traits.AppliesTo;
+import software.amazon.smithy.protocoltests.traits.eventstream.Event;
+import software.amazon.smithy.protocoltests.traits.eventstream.EventHeaderValue;
+import software.amazon.smithy.protocoltests.traits.eventstream.EventStreamTestCase;
+import software.amazon.smithy.protocoltests.traits.eventstream.EventStreamTestsTrait;
+import software.amazon.smithy.protocoltests.traits.eventstream.EventType;
+
+/**
+ * Generates protocol unit tests for event stream operations from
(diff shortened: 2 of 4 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
```diff
@@ -0,0 +1,658 @@
...
+ */
+public class EventStreamProtocolTestGenerator {
+    private static final Logger LOGGER = Logger.getLogger(EventStreamProtocolTestGenerator.class.getName());
+
+    private static final Set<String> SKIP_TESTS = Set.of(
+            // SDK does not validate required fields on responses client-side.
+            "MissingRequiredInitialResponseOutput",
+            "DuplexMissingRequiredInitialResponseOutput",
+
+            // SDK silently drops unrecognized/malformed event stream messages
```
Madrigal: So if we have an issue on one event stream message we just skip it and continue with the rest?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
```diff
@@ -0,0 +1,658 @@
...
+        this.symbolProvider = ctx.symbolProvider();
+        this.delegator = (GoDelegator) ctx.writerDelegator();
+        this.eventStreamIndex = EventStreamIndex.of(model);
+    }
+
+    public void generateProtocolTests() {
+        TopDownIndex topDownIndex = model.getKnowledge(TopDownIndex.class);
+
+        boolean hasAnyTests = false;
+        for (OperationShape operation : new TreeSet<>(topDownIndex.getContainedOperations(service))) {
```
Madrigal, line 84: nit, but could we `streamify` this whole operation? Something like

```java
boolean hasAnyTest = topDownIndex.getContainedOperations(service).stream()
.filter(operation -> operation.hasTrait(EventStreamTestsTrait.class))
.map(operation -> operation.expectTrait(EventStreamTestsTrait.class))
.map(trait -> trait.getTestCasesFor(AppliesTo.CLIENT))
.filter(testCase -> testCase.getProtocol().equals(settings.getProtocol()))
.findAny()
```

I just find it easier to skim streams and say "we're doing some map filter stuff without side effects"

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
```diff
@@ -0,0 +1,658 @@
...
+                }
+            }
+        }
+
+        if (!hasAnyTests) {
+            return;
+        }
+
+        // Generate the harness file
+        delegator.useFileWriter("eventstream_protocol_test_harness_test.go", settings.getModuleName(), writer -> {
```
Madrigal: nit: could we leave this just as `eventstream_protocol_test.go`?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
```diff
@@ -0,0 +1,658 @@
...
+                        if r.Body != nil {
+                            io.Copy(m.requestBuf, r.Body)
+                        }
+                    }()
+
+                    var body io.ReadCloser
+                    if m.StatusCode >= 400 {
+                        body = io.NopCloser(strings.NewReader(m.ResponseBody))
+                    } else {
+                        pr, pw := io.Pipe()
```
Madrigal, line 178: I get PTSD from reading `io.Pipe()`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
```diff
@@ -0,0 +1,658 @@
...
+                return;
+            }
+
+            writer.openBlock("if err != nil {", "}", () ->
+                    writer.write("t.Fatalf(\"expect no error, got %v\", err)"));
+
+            // For error HTTP responses: the SDK surfaces these asynchronously
+            // through the stream error, not as an operation-level error.
+            if (isErrorResponse) {
+                writer.write("<-resp.GetInitialReply()");
```
Madrigal, line 374: Ah, I suspect this works because we're testing this against the newer version of streams where we introduced the concept of InitialReply. We could run this against the v1 generator, but that ship already sailed, so even if we found any test failures there wouldn't be much we could do about it

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
```diff
@@ -0,0 +1,658 @@
...
+                // Collect events with params for assertion
+                List<Event> assertableEvents = responseEvents.stream()
+                        .filter(e -> e.getParams().isPresent())
+                        .collect(Collectors.toList());
+
+                writer.write("var receivedEvents int");
+                if (!expectFailure && !assertableEvents.isEmpty()) {
+                    generateEventAssertions(writer, operation, responseEvents);
+                } else {
+                    writer.openBlock("for event := range resp.GetStream().Events() {", "}", () -> {
```
Madrigal, line 395: Couldn't we do just `len(resp.GetStream().Events())`?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
```diff
@@ -0,0 +1,658 @@
...
+
+        // Close the stream to flush
+        writer.write("mock.CloseResponse()");
+        writer.write("resp.GetStream().Close()");
+        writer.write("");
+
+        // Filter to only sendable (non-error) events for assertion
+        List<Event> sendableEvents = requestEvents.stream()
+                .filter(e -> {
+                    if (e.getParams().isEmpty()) return false;
```
Madrigal: move this check to be part of the stream pipeline instead of having it be part of the same filter clause

```java
.filter(e -> e.getParams().isPresent())
.filter(e -> { String mn = e.getParams() //....})
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
```diff
@@ -0,0 +1,658 @@
...
+        int idx = 0;
+        for (Event event : sendableEvents) {
+            final int eventIdx = idx;
+            if (event.getBytes().isPresent()) {
+                // Compare full message bytes
+                byte[] raw = event.getBytes().get();
+                writer.openBlock("{", "}", () -> {
+                    writer.write("expectRaw, _ := base64.StdEncoding.DecodeString($S)",
+                            Base64.getEncoder().encodeToString(raw));
+                    writer.write("expectMsg, _ := eventstream.NewDecoder().Decode(bytes.NewReader(expectRaw), make([]byte, 1024))");
```
Madrigal, line 550: is it reasonable to expect all messages will fit on the 1024 byte?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
```diff
@@ -0,0 +1,658 @@
...
+                    writer.write("actualMsg := requestMsgs[$L]", eventIdx);
+                    // Compare headers
+                    writer.openBlock("for _, eh := range expectMsg.Headers {", "}", () -> {
```
Madrigal, line 553: I chuckled a bit with `eh` and `ah`. I thought about a different name, but `expectHead` may not be an amazing replacement, and `expectedHeaders` seems to verbose for something that will live for 5 lines

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
```diff
@@ -709,6 +720,8 @@ public Void numberNode(NumberNode node) {
         public Void stringNode(StringNode node) {
             switch (currentShape.getType()) {
                 case BLOB:
+                    writer.writeInline("$S", node.getValue());
```
Madrigal, line 723: nit: is this because we prefer to have the case being explicit instead of relying on case fall-through?

Madrigal (approved):
Glad to see we didn't completely explode with this, minor suggestions but overall LGTM

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
lucix-aws: (yes)

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
lucix-aws: not on a channel, no

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ShapeValueGenerator.java
lucix-aws: oh this is actually an artifact from earlier when the test model had base64'd blobs (had to handle them differently). i didn't realize that was wrong at that point. i upstreamed that fix to smithy, but also, yes i'd prefer not fallthrough i guess

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
lucix-aws: that is just seed space for a bytes.Buffer https://github.com/aws/aws-sdk-go-v2/blob/main/aws/protocol/eventstream/decode.go#L155, it'll grow on the Copy() call there if it has to

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/EventStreamProtocolTestGenerator.java
lucix-aws: minor update here, i didn't actually realize that there was a difference between "unknown message type" and "unknown event type". the former we do error on, the latter just becomes the "unknown" variant in the union. so some of these got unskipped.

PR #662, opened 2026-05-15
serde2: get the endpoint/auth generators out of the protocol generator

lucix-aws:
Protocol generator will end up going away so these need to generate on their own.

The SDK overrode the protocol generators to add aws endpoint rule functions, so I just replaced that with a new integration hook that lets you add a custom one.

The codegen output is identical from before.

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
@@ -33,6 +33,10 @@ import software.amazon.smithy.build.PluginContext;
 import software.amazon.smithy.codegen.core.SymbolDependency;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.go.codegen.GoSettings.ArtifactType;
+import software.amazon.smithy.go.codegen.auth.AuthGenerator;
+import software.amazon.smithy.go.codegen.endpoints.EndpointResolutionGenerator;
+import software.amazon.smithy.go.codegen.endpoints.FnGenerator;
+import software.amazon.smithy.go.codegen.endpoints.FnProvider;
 import software.amazon.smithy.go.codegen.integration.GoIntegration;
 import software.amazon.smithy.go.codegen.integration.ProtocolGenerator;
 import software.amazon.smithy.go.codegen.integration.RuntimeClientPlugin;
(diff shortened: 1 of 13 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/auth/AuthGenerator.java
```diff
@@ -15,35 +15,25 @@
...
-        context.getWriter().get()
-                .write("$W\n", new AuthParametersGenerator(context).generate())
-                .write("$W\n", new AuthParametersResolverGenerator(context).generate())
-                .write("$W\n", getResolverGenerator().generate())
-                .write("$W\n", new ResolveAuthSchemeMiddlewareGenerator(context).generate())
-                .write("$W\n", new GetIdentityMiddlewareGenerator(context).generate())
-                .write("$W\n", new SignRequestMiddlewareGenerator(context).generate());
-    }
-
-    // TODO(i&a): allow consuming generators to overwrite
```
Madrigal, line 45: I know we only inlined this, but we got rid of this TODO, is this what we're now accomplishing with 

>  I just replaced that with a new integration hook that lets you add a custom one.

?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/GoIntegration.java
```diff
@@ -173,6 +174,14 @@ default List<ProtocolGenerator> getProtocolGenerators() {
         return Collections.emptyList();
     }
 
+    /**
+     * Gets a custom endpoint rule function provider. Returning null (the default) indicates no custom functions.
+     * The returned provider is consulted before the built-in default provider.
+     */
+    default FnProvider getEndpointFnProvider() {
```
Madrigal, line 181: I'd rather name this `getOptionalEndpointFnProvider` to make it clearer that you don't have to override it and make it more clear it can be `null`

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
```diff
@@ -381,33 +385,21 @@ void execute() {
...
-                    .protocolName(protocolGenerator.getProtocolName())
-                    .integrations(integrations)
-                    .model(model)
-                    .service(service)
-                    .settings(settings)
-                    .symbolProvider(symbolProvider)
-                    .delegator(writers);
+        {
+            FnProvider fnProvider = integrations.stream()
+                    .map(GoIntegration::getEndpointFnProvider)
```
Madrigal, line 390: on these cases streams are overkill, since we're just transforming 1 single thing and checking for not null. This is probably the only use case where I like the ternary operator 

```java
FnProvider fnProvider = GoIntegration::getEndpointFnProvider() != null? 
GoIntegration::getEndpointFnProvider(): 
new FnGenerator.DefaultFnProvider()
```

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/GoIntegration.java
lucix-aws: i mean it does say that above and also the method is defaulted

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
lucix-aws: how do we actually use the ternary there though? we have to loop over them and find one. we could

```
var provider = defaultprovider;
for (integrations) {
  if (integration.provider) {
    provider = integration.provider;
    break;
  }
}
```

but that seems equivalent or worse

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/auth/AuthGenerator.java
lucix-aws: I genuinely have no idea what this was about, we don't need this to be pluggable at all so this is probably just 2022/2023 Luc saying things he didn't understand.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/CodegenVisitor.java
Madrigal: sorry, for some reason I didn't realized that you ARE looping through a collection, in my mind this was just a stream of 1. Ignore this

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/GoIntegration.java
Madrigal: I don't feel strongly to block on this, but I find it way easier to reason on why we are filtering nonNull if I see it explicitly written as optional when we are using it

```java
 FnProvider fnProvider = integrations.stream()
                    .map(GoIntegration::getOptionalEndpointFnProvider)
                    .filter(java.util.Objects::nonNull)
                    .findFirst()
                    .orElse(new FnGenerator.DefaultFnProvider());
```

Madrigal (approved):

PR #663, opened 2026-05-18
serde2: drop writeptr shenanigans and optimize allocs on loop deserialize

lucix-aws:
relates #458 

Turns out the writeptr split is not necessary at all if we just generate the "don't serialize value types if zero" checks in SerializeMembers. In dynamodb the compiled output actually shrunk somewhat, I think because all the runtime stuff for pointers going away, but in practice it probably is like a 0.1% artifact size increase or something to generate the branches inline.

Also fixed an issue where we were reallocating every loop iteration for deserializing list/map values (i just moved the temporary to outside the loop, so it only allocates once).

Double-checked all protocol tests on this one, all still good.


Madrigal (approved):
🚀

PR #664, opened 2026-05-18
serde2: drop readptr shenanigans

lucix-aws:
same deal as #663 

relates #458

only relevant change here is that the two formats w/ nulls (json and cbor, cbor may not even need it but i did it anyway) will now just check for nil themselves in ReadStructMember. so basically the "read member" hook is _only_ ever called if there's a value

also, the deserializer was missing bignums, like the serializer they are "unimplemented" for now

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde2/StructureDeserializer.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde2/StructureDeserializer.java
@@ -57,21 +57,21 @@ public class StructureDeserializer implements Writable {
 
     private void renderMember(GoWriter writer, MemberShape member, Shape target, String ident) {
         var schemaName = SchemaGenerator.getMemberSchemaRef(shape, member, ctx.service());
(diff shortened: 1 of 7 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde2/StructureDeserializer.java
```diff
@@ -134,4 +134,13 @@ private void renderMember(GoWriter writer, MemberShape member, Shape target, Str
             case MEMBER, SERVICE, RESOURCE, OPERATION -> throw new UnsupportedShapeException(target.getType());
         }
     }
+
+    private void writeReadScalar(GoWriter writer, boolean isNillable, String ident, String schemaName,
+                                 String readMethod, String goType) {
+        if (isNillable) {
+            writer.write("$1L = new($4L)\nreturn d.$3L($2L, $1L)", ident, schemaName, readMethod, goType);
```
Madrigal, line 141: nit, but why do we Yoda arguments here?

Madrigal (approved):

PR #667, opened 2026-05-26
handle vars in auth scheme id tmpl

lucix-aws:
* every property now uses the template generator to handle vars etc.
* the "do we need string slice" logic was going to get more complicated so I just converted it to something in the runtime. the codegen that uses it didn't have a good way to inject a symbol so i just added a pinned usage
* previously we prepended the smithy shape id namespace to auth scheme ids in endpoint rules. since those values can now be dynamic that is no longer possible to do, thus the names are now returned as is and the auth scheme resolution logic matches without namespace

### DIFF
--- a/endpoints/private/rulesfn/string_slice.go
+++ b/endpoints/private/rulesfn/string_slice.go
@@ -1,0 +1,18 @@
+package rulesfn
+
+// StringSlice is a string slice with a negative-index-aware Get method for use
+// in endpoint rule evaluation.
+type StringSlice []string
+
+// Get returns a pointer to the string at index i, or nil if the index is out
+// of bounds. Negative indices count from the end of the slice.
+func (s StringSlice) Get(i int) *string {
+	if i < 0 {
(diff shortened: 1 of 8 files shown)


endpoints/private/rulesfn/string_slice.go
```diff
@@ -0,0 +1,18 @@
+package rulesfn
+
+// StringSlice is a string slice with a negative-index-aware Get method for use
+// in endpoint rule evaluation.
+type StringSlice []string
+
+// Get returns a pointer to the string at index i, or nil if the index is out
+// of bounds. Negative indices count from the end of the slice.
```
Madrigal, line 8: was the negative index something we had to add due to new requirements?

lucix-aws: it looks like it was added in "rules language 1.1" in smithy and just nothing has used it until now

Madrigal (approved):

PR #682, opened 2026-06-30
generate serde snapshot tests

lucix-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
@@ -1,0 +1,340 @@
+package software.amazon.smithy.go.codegen.integration;
+
+import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
+
+import java.util.ArrayList;
+import java.util.Set;
+import software.amazon.smithy.codegen.core.SymbolProvider;
+import software.amazon.smithy.go.codegen.ChainWritable;
+import software.amazon.smithy.go.codegen.GoDelegator;
+import software.amazon.smithy.go.codegen.GoSettings;
+import software.amazon.smithy.go.codegen.GoStdlibTypes;
+import software.amazon.smithy.go.codegen.SmithyGoDependency;
+import software.amazon.smithy.go.codegen.Writable;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.EventStreamIndex;
+import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.utils.MapUtils;
+
+public class SerdeSnapshotTests implements GoIntegration {
+    private static final Set<String> SKIP_OPERATIONS = Set.of(
+            "com.amazonaws.machinelearning#Predict"
+    );
+
+    @Override
+    public void writeAdditionalFiles(
+            GoSettings settings, Model model, SymbolProvider symbolProvider, GoDelegator goDelegator
+    ) {
+        goDelegator.useFileWriter("serde_snapshot_test.go", settings.getModuleName(), writer -> {
+            writer.addBuildTag("serde_snapshot");
+            writer.write(commonSource());
+
+            var service = settings.getService(model);
+            var generator = new SnapshotInputGenerator(model, symbolProvider);
+            writer.write(snapshotTests(model, service, symbolProvider, generator));
+            writer.write(snapshotUpdaters(model, service, symbolProvider, generator));
+        });
+    }
+
+    private Writable commonSource() {
+        return writer -> {
+            writer.addUseImports(SmithyGoDependency.OS);
+            writer.addUseImports(SmithyGoDependency.FS);
+            writer.addUseImports(SmithyGoDependency.IO);
+            writer.addUseImports(SmithyGoDependency.ERRORS);
+            writer.addUseImports(SmithyGoDependency.FMT);
+            writer.addUseImports(SmithyGoDependency.BYTES);
+            writer.addUseImports(SmithyGoDependency.CONTEXT);
+            writer.addUseImports(SmithyGoDependency.SMITHY_MIDDLEWARE);
+            writer.addUseImports(SmithyGoDependency.SMITHY_HTTP_TRANSPORT);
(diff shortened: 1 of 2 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
```diff
@@ -0,0 +1,386 @@
...
+import software.amazon.smithy.model.knowledge.EventStreamIndex;
+import software.amazon.smithy.model.knowledge.TopDownIndex;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.protocol.traits.Rpcv2CborTrait;
+import software.amazon.smithy.utils.MapUtils;
+
+public class SerdeSnapshotTests implements GoIntegration {
+    private static final Set<String> SKIP_OPERATIONS = Set.of(
+            "com.amazonaws.machinelearning#Predict"
```
Madrigal, line 25: a comment explaining why would be great

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
```diff
@@ -0,0 +1,386 @@
...
+            writer.write(snapshotUpdaters(model, service, symbolProvider, generator));
+        });
+    }
+
+    private static boolean isCbor(GoSettings settings) {
+        return Rpcv2CborTrait.ID.equals(settings.getProtocol());
+    }
+
+    // Emits serdeBodyEqual, the request-body comparator. Most protocols serialize
+    // deterministically, so a raw byte compare is correct. rpcv2Cbor encodes struct
```
Madrigal, line 49: didn't we also had the same issue with JSON serde, and a discussion on whether we should or should not make this deterministic? Can we make that guarantee now?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SnapshotInputGenerator.java
```diff
@@ -0,0 +1,330 @@
...
+        writer.writeInline("}");
+    }
+
+    private void writeUnion(GoWriter writer, UnionShape shape, Set<software.amazon.smithy.model.shapes.ShapeId> visited, UnionChoice choice) {
+        var members = shape.getAllMembers().values().stream().toList();
+        boolean recursing = visited.contains(shape.getId());
+
+        MemberShape chosenMember;
+        if (recursing) {
+            // We've recursed back into this union (e.g. a variant that is a
```
Madrigal, line 244: this must have been fun to debug

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SnapshotInputGenerator.java
```diff
@@ -0,0 +1,330 @@
...
+        var inputShape = model.expectShape(operation.getInputShape(), StructureShape.class);
+        return List.of(new TestCase("", writer -> {
+            writeStructure(writer, inputShape, new HashSet<>(), UnionChoice.DEFAULT, true);
+        }));
+    }
+
+
+    private static final int MAX_DEPTH = 10;
+
+    private void writeStructure(GoWriter writer, StructureShape shape, Set<software.amazon.smithy.model.shapes.ShapeId> visited, UnionChoice choice,
```
Madrigal: do we have another class that conflicts with ShapeId? if not, we should just import that

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
```diff
@@ -0,0 +1,386 @@
...
+            writer.addUseImports(SmithyGoDependency.SMITHY_MIDDLEWARE);
+            writer.addUseImports(SmithyGoDependency.SMITHY_HTTP_TRANSPORT);
+            writer.addUseImports(SmithyGoDependency.STRINGS);
+            writer.addUseImports(SmithyGoDependency.SLICES);
+            writer.write("""
+                    const serdeSSPrefix = "serde_snapshot"
+
+                    type serdeSnapshotOK struct{}
+
+                    func (serdeSnapshotOK) Error() string { return "error: success" }
```
Madrigal: 1. Why do we check if the run was successful by returning an error? can we return `nil` instead?
2. If we need to, can we check for the type of error with `errors.Is/As` instead of checking for a literal `error: success`?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
```diff
@@ -0,0 +1,386 @@
...
+                            return stack.Finalize.Add(&captureSerdeRequestMiddleware{
+                                body: body, method: &method, rawPath: &rawPath, rawQuery: &rawQuery, header: &header,
+                            }, $middlewareBefore:T)
+                        })
+                    })
+                    if err != nil && !strings.Contains(err.Error(), "error: success") {
+                        t.Fatal(err)
+                    }
+                    if err := serdeTestSnapshot(method, rawPath, rawQuery, header, body.Bytes(), $name:S); err != nil {
+                        if err != nil && !strings.Contains(err.Error(), "error: success") {
```
Madrigal: nit: we already checked that `err != nil`, although as mentioned on another comment I'd like to see if we can do something else instead

Madrigal (commented):
To confirm, we are only generating snapshot of the `input`, correct?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
lucix-aws: no, schema-serde could be adjusted to make that possible but right now json maps are unordered so we're just serializing one map entry. we had to do this for cbor because cbor serializes structs using a map in the old pre-schema-serde code

lucix-aws:
> To confirm, we are only generating snapshot of the input, correct?

yes - response snapshotting is a followup and requires more schema-serde work since we need it to actually generate mock responses

Madrigal (approved):

PR #683, opened 2026-07-06
feat: move close-body, logger, and service-metadata work out of the middleware stack

Tongs2000:
*Issue #, if available:*

N/A

*Description of changes:*

Reduce the number of per-request middlewares on every operation's stack by moving work that does not need to be a standalone middleware into codegen: into the generated operation deserializer or `invokeOperation`. Each removed middleware saves at least 2 allocations per request (the struct plus its handler wrapper) on a known hot path.

Three independent changes, all in codegen plus shared runtime helpers:

**1. Close response body in the operation deserializer (Deserialize -1)**
- Adds `CloseResponseBody(ctx, resp, isStreaming, opErr)`, a shared function that closes the response body. It leaves the body open only for a successful response whose payload is a caller-owned stream (`isStreaming` true with a nil `opErr`); on error, or for a non-streaming response, it always closes the body (an error response body is diagnostic, not a caller-owned stream).
- Codegen emits a deferred `CloseResponseBody` call at the top of each generated operation deserializer, across the `HttpRpc`, `HttpBinding`, and `protocol/serde2` deserializer generators. The call is deferred in a closure (`defer func() { CloseResponseBody(..., err) }()`) so it observes the final operation error.
- Streaming detection differs by path. For the legacy `HttpRpc` / `HttpBinding` generators, `isStreaming` is decided at codegen time via `ProtocolUtils.isCallerOwnedResponseStream` (event stream, or a `@streaming` payload). For `serde2`, the per-service middleware cannot see the operation at codegen time, so it is decided at runtime: `m.operationSchema.IsOutputEventStream()` for event streams, or the `smithy.StreamingOutput` interface for a `@streaming` payload output.
- The standalone close-response-body middleware is no longer registered; the now-unused `HttpProtocolUtils.getCloseResponseClientPlugins` helper is removed.
- `AddCloseResponseBodyMiddleware` and `AddErrorCloseResponseBodyMiddleware` are kept and marked `// Deprecated` (they now delegate to `CloseResponseBody`) so external callers are not broken.

**2. Operation context resolver codegen hook**
- Adds a `RuntimeClientPlugin` extension point, `operationContextResolvers`: a set of symbols pointing to functions with the signature `func(context.Context, Options, string) context.Context`. `invokeOperation` emits `ctx = resolver(ctx, options, opID)` for each, after options are finalized and before the middleware stack is built.
- This lets a plugin set request-scoped context values directly in `invokeOperation` instead of via a per-request Initialize-step middleware, for values that live in a keyspace the smithy-go base template cannot reference.

**3. Set the logger via the context resolver hook (Initialize -1)**
- The `setLogger` middleware only stashed `options.Logger` into the request context at the start of Initialize. `ClientLogger` now generates a `setLoggerContext` resolver (`return middleware.SetLogger(ctx, options.Logger)`) and registers it through the operation context resolver hook above, so the logger is set in `invokeOperation` instead of by a standalone middleware.
- `ClientLogger` no longer registers a middleware; it still contributes the `Logger` config field and the default-logger resolver. The public `AddSetLoggerMiddleware` function is unchanged.


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ClientLogger.java
```diff
@@ -55,21 +54,14 @@ public void writeAdditionalFiles(
...
-            writer.openBlock("func $L(stack $P, o Options) error {", "}", REGISTER_MIDDLEWARE, stackSymbol, () -> {
-                writer.write("return $T(stack, o.$L)", helperSymbol, LOGGER_CONFIG_NAME);
-            });
-            writer.write("");
         });
     }
 
     @Override
     public List<RuntimeClientPlugin> getClientPlugins() {
+        // The logger is set on the context directly in invokeOperation
```
lucix-aws: shouldn't this now have an OperationContextResolver instead? if someone overrides this integration the generated code that references options.Logger elsewhere will be broken

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/DeserializeResponseMiddleware.java
```diff
@@ -62,9 +64,15 @@ public void accept(GoWriter writer) {
     public abstract Writable generateDeserialize();
 
     private Writable generateHandleDeserialize() {
+        // Close body on success unless the output is a caller-owned event stream.
```
lucix-aws: dont both of these serialize/deserialize middleware changes also need replicated in the schema-serde path? this is the legacy serde path

transport/http/middleware_close_response_body.go
```diff
@@ -8,9 +8,40 @@ import (
 	"github.com/aws/smithy-go/middleware"
 )
 
+// DrainAndCloseResponseBody drains and closes the HTTP response body. It always
```
lucix-aws: for something we're adding to the runtime like this we should really just explicitly parameterize it rather than have behavior flags that are documented as "pass x when y".

e.g. `DrainAndClose(ctx, resp, err, isStreaming)`

transport/http/middleware_close_response_body.go
```diff
@@ -8,9 +8,40 @@ import (
...
+// closes on error; on success it closes only when closeOnSuccess is true (pass
+// false when the response is a streaming payload owned by the caller).
+func DrainAndCloseResponseBody(ctx context.Context, resp *Response, closeOnSuccess bool, opErr error) {
+	if resp == nil || resp.Body == nil {
+		return
+	}
+
+	if opErr != nil {
+		// Consume the full body to prevent TCP connection resets on some platforms.
```
lucix-aws: I know this is existing but I'm wondering if we really care to do this at all. In the success case we know we read the whole body to deserialize it so it's pointless. If at this point we _did not_ read the whole response body it's either because

1. it's a streaming payload and it's the caller's problem
2. we had an actual error reading the response body, which means any further attempt to read it would probably also error, and also if this claim is true about connection reuse (idk if it is tbh) we probably want to throw that connection away anyway

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/DeserializeResponseMiddleware.java
```diff
@@ -62,9 +64,15 @@ public void accept(GoWriter writer) {
...
+        boolean closeOnSuccess = EventStreamIndex.of(ctx.getModel()).getOutputInfo(operation).isEmpty();
         return goTemplate("""
                 out, metadata, err = next.HandleDeserialize(ctx, in)
 
+                // Close the body on every exit path in place of the standalone close middleware.
+                resp, _ := out.RawResponse.($response:P)
+                defer $drainClose:T(ctx, resp, $closeOnSuccess:L, err)
```
lucix-aws: arguments passed to defer are evaluated/captured _when defer is called_, so if err gets set to something else later you're dropping its value

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/DeserializeResponseMiddleware.java
```diff
@@ -62,9 +64,15 @@ public void accept(GoWriter writer) {
     public abstract Writable generateDeserialize();
 
     private Writable generateHandleDeserialize() {
+        // Close body on success unless the output is a caller-owned event stream.
+        boolean closeOnSuccess = EventStreamIndex.of(ctx.getModel()).getOutputInfo(operation).isEmpty();
```
lucix-aws: pretty sure this should also be the case for `@streaming` blob? e.g. GetObject

lucix-aws:
We are probably going to want some new kitchen sink tests for these things over in the SDK (`internal/kitchensinktest`, it's just a bogus generated client with an operation or two that you can write tests against). especially with the response body close stuff, it would be good to have tests actually go into main _right now_ (before this is merged) that verify the behavior so we can check it didn't change overall here.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/DeserializeResponseMiddleware.java
Tongs2000: Good call. This is the legacy path. serde2 relied on the same standalone close/content-length middlewares (registered for all services), which this change removes, so I've replicated the inline behavior in the serde2 path:

  - `Serde2DeserializeResponseMiddleware`: `defer CloseResponseBody` after deserialize, skipping caller-owned streams (`IsOutputEventStream()` or the `smithy.StreamingOutput` interface on the output).
  - `Serde2SerializeRequestMiddleware`: compute content length after serialize, skipping event-stream inputs.

  Streaming is detected at runtime here since the serde2 middlewares are per-service rather than per-operation, but the behavior matches the legacy paths.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ClientLogger.java
Tongs2000: Good catch, agreed. Done. `ClientLogger` now generates a `setLoggerContext` resolver and registers it via `addOperationContextResolver`, and I removed the hardcoded `SetLogger` line from the `invokeOperation` template. So overriding the integration removes the logger setup too, no dangling `options.Logger` reference.

transport/http/middleware_close_response_body.go
Tongs2000: Agreed, done. Switched to an explicit `isStreaming` param. It actually simplified further than your example: since we're dropping the success-path drain (the body's already consumed during deserialization), the success and error paths do the same thing, so the `err` param isn't needed either. It's now `CloseResponseBody(ctx, resp, isStreaming)`.

transport/http/middleware_close_response_body.go
Tongs2000: Agreed. Removed the drain entirely. Success path: the body's already fully read during deserialization, so there's nothing to drain. Error path: as you said, a failed read won't get better by re-reading, and we'd want to discard that connection anyway. So `CloseResponseBody` now just closes (unless it's a caller-owned stream) and never drains.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/DeserializeResponseMiddleware.java
Tongs2000: Good catch. Rather than wrap it in a closure to capture the final `err`, the reworked function doesn't take `err` at all anymore (dropping the drain made the success/error paths identical, so there's nothing to branch on). The generated call is now `defer CloseResponseBody(ctx, resp, isStreaming)` with no `err`, so the early-capture problem goes away entirely.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/DeserializeResponseMiddleware.java
Tongs2000: Good catch. And while doing it I noticed the three deserialize paths had drifted: `HttpBinding` checked both event streams and `@streaming` payloads, but `HttpRpc` and this serde path only checked event streams. I pulled the check into a shared `ProtocolUtils.isCallerOwnedResponseStream(model, operation)` (event stream OR `@streaming` payload) and pointed all three paths at it, so they can't drift again.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/DeserializeResponseMiddleware.java
Tongs2000: Following up on my previous reply: I walked this back. Dropping `err` was wrong for streaming outputs.

  The concern is the error path for a caller-owned stream. If an operation whose output is a streaming payload (or event stream) fails during deserialization, `isStreaming` is `true`, so a no-`err` `CloseResponseBody(ctx, resp, isStreaming)` would skip the close and leak the body, since there is no successful
  stream for the caller to consume and close. The standalone error-close middleware closed it in that case, so dropping `err` would have been a behavior change (the kitchensink test caught exactly this: streaming output + error path, body not closed).

  So the function keeps `err`: `CloseResponseBody(ctx, resp, isStreaming, opErr)`. It leaves the body open only when `isStreaming && opErr == nil` (a successful caller-owned stream); on error, or for a non-streaming response, it always closes. This preserves the original behavior on every path.

  To your original point about `defer` capturing `err` too early: the generated call is a closure, `defer func() { CloseResponseBody(ctx, resp, isStreaming, err) }()`, so it reads the final `err` at return time rather than capturing it at the `defer` line. `err` is a named return value in the generated
  deserializer, so the closure sees whatever the deserialize logic last set.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ClientLogger.java
```diff
@@ -56,14 +56,15 @@ public void writeAdditionalFiles(
...
                     .build();
-            Symbol helperSymbol = SymbolUtils.createValueSymbolBuilder("AddSetLoggerMiddleware",
-                    SmithyGoDependency.SMITHY_MIDDLEWARE).build();
-
-            writer.openBlock("func $L(stack $P, o Options) error {", "}", REGISTER_MIDDLEWARE, stackSymbol, () -> {
-                writer.write("return $T(stack, o.$L)", helperSymbol, LOGGER_CONFIG_NAME);
-            });
+            writer.openBlock("func $L(ctx $T, options Options, operation string) $T {", "}",
+                    SET_LOGGER_CONTEXT_RESOLVER, contextSymbol, contextSymbol, () -> {
+                        writer.write("_ = operation");
```
wty-Bryant, line 65: Is there any other resolver func implementing this func(ctx, options, string) required by operationContextResolvers? Asked since the opId here is just dropped
My understanding is this opId is kept for forward-compatibility once any other operation ctx resolvers are added

Tongs2000: Yes. The `operation` argument is consumed by the paired aws-sdk-go-v2 PR ([#3480](https://github.com/aws/aws-sdk-go-v2/pull/3480)), which contributes a `resolveServiceMetadata` resolver of this same signature:

  ```go
  func resolveServiceMetadata(ctx context.Context, options Options, operation string) context.Context {
        ctx = awsmiddleware.SetServiceID(ctx, ServiceID)
        if options.Region != "" {
                ctx = awsmiddleware.SetRegion(ctx, options.Region)
        }
        ctx = awsmiddleware.SetOperationName(ctx, operation) // <- uses opID
        if options.EndpointResolver != nil {
                ctx = awsmiddleware.SetRequiresLegacyEndpoints(ctx, true)
        }
        return ctx
  }
  ```

  This replaces the old per-operation `RegisterServiceMetadata` Initialize middleware. That middleware stamped the operation name into the request context, and downstream code reads it back via `GetOperationName(ctx)`:

  - `aws/retry/attempt_metrics.go` sets `rpc.method` on the metrics from it
  - `aws/retry/middleware.go` logs `service, operation` on retries
  - `feature/s3/manager` branches on `GetOperationName(ctx) == "CreateMultipartUpload"`
  
  So the `opID` has to reach the context somehow.

  It now needs to be passed in because the old middleware was generated per operation, so each instance carried its own name hard-coded in the struct (`OperationName: "ListQueues"`) and just read `s.OperationName` at runtime, no argument needed. Folding it into `invokeOperation` collapses that into a single shared
  `resolveServiceMetadata` function used by every operation, which therefore can't carry any one operation's name. The only way it still knows which operation is running is the opID that `invokeOperation` already has, passed in via the hook: `ctx = resolveServiceMetadata(ctx, options, opID)`.

  `setLoggerContext` is the one in-tree (smithy-go) resolver, and the logger genuinely doesn't depend on the operation, so it drops the arg (`_ = operation`). The third parameter is part of the shared contract for resolvers that do need it, `resolveServiceMetadata` being the motivating case.

transport/http/middleware_close_response_body.go
```diff
@@ -28,20 +47,19 @@ func (m *errorCloseResponseBodyMiddleware) HandleDeserialize(
 ) {
 	out, metadata, err := next.HandleDeserialize(ctx, input)
 	if err != nil {
-		if resp, ok := out.RawResponse.(*Response); ok && resp != nil && resp.Body != nil {
```
lucix-aws, line 31: Is there any reason you're changing these deprecated implementations? can't they just stay as-is?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -404,8 +427,7 @@ defer endTimer()
                     defer span.End()
                     """, SMITHY_TRACING.func("StartSpan")));
 
-            writer.write("response, ok := out.RawResponse.($P)", responseType);
-            writer.openBlock("if !ok {", "}", () -> {
+            writer.openBlock("if response == nil {", "}", () -> {
```
lucix-aws: ~~why change this?~~

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
```diff
@@ -390,10 +390,33 @@ private void generateOperationDeserializerMiddleware(GenerationContext context,
...
         middleware.writeMiddleware(goWriter, (generator, writer) -> {
             writer.addUseImports(SmithyGoDependency.FMT);
+            writer.addUseImports(SmithyGoDependency.SMITHY_HTTP_TRANSPORT);
 
             writer.write("out, metadata, err = next.$L(ctx, in)", generator.getHandleMethodName());
+            writer.write("");
+
+            // Close the response body after deserialization (a streaming payload is
+            // kept open on success). Event streams close their own body.
+            writer.write("response, _ := out.RawResponse.($P)", responseType);
```
lucix-aws: you really need to keep the explicit type check here, I know that you have the nil guard in CloseResponseBody but you don't have to if you're doing the proper (response, ok -> check ok) like we do everywhere else

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
lucix-aws: i think it relates to the above

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpRpcProtocolGenerator.java
```diff
@@ -334,11 +335,34 @@ private void generateOperationDeserializer(GenerationContext context, OperationS
...
             writer.addUseImports(SmithyGoDependency.FMT);
             writer.addUseImports(SmithyGoDependency.SMITHY);
+            writer.addUseImports(SmithyGoDependency.SMITHY_HTTP_TRANSPORT);
 
             writer.write("out, metadata, err = next.$L(ctx, in)", generator.getHandleMethodName());
+            writer.write("");
+
+            // Close the response body after deserialization (a streaming payload is
+            // kept open on success). Event streams close their own body.
+            writer.write("response, _ := out.RawResponse.($P)", responseType);
```
lucix-aws: same deal as above

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/protocol/DeserializeResponseMiddleware.java
```diff
@@ -62,9 +64,31 @@ public void accept(GoWriter writer) {
...
+                        MapUtils.of(
+                                "closeBody", SmithyGoDependency.SMITHY_HTTP_TRANSPORT.func("CloseResponseBody"),
+                                "isStreaming", isStreaming ? "true" : "false"
+                        ));
+
         return goTemplate("""
                 out, metadata, err = next.HandleDeserialize(ctx, in)
 
+                // Close the response body once deserialization is done.
+                resp, _ := out.RawResponse.($response:P)
```
lucix-aws: same as above

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
lucix-aws: basically the pattern should always be, for deserialize

```
out,md,err = next.HandleDeserialize
if err != nil {
  return out,md,err
}

resp, ok := out.whatever.(smithyhttp.Response)
if !ok {
  return out, md, error about wrong transport
}

proceed
```

transport/http/middleware_close_response_body.go
Tongs2000: You're right, will revert to the original implementation. This was from an earlier iteration where I was consolidating both paths before deciding to deprecate. Since they're deprecated now it doesn't matter either way, but no reason to touch them.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/HttpBindingProtocolGenerator.java
Tongs2000: Acknowledged on all of these. Will update all three generators (HttpBinding, HttpRpc, serde2)

lucix-aws (approved):

lucix-aws:
approve pending CI pass AND downstream SDK CI pass

PR #687, opened 2026-07-14
support sigv4[a] and non-rule endpoint resolution in generic clients

lucix-aws:
- new module aws-http-auth-schemes bridges the standalone aws-http-auth signers into smithy-go's client auth interfaces, so generic (non-SDK) clients can sign with SigV4/SigV4A
without depending on aws-sdk-go-v2
- `@Replaces` annotation lets one GoIntegration suppress another, so the SDK can swap in its own auth scheme/HTTP client/endpoint resolver over the generic defaults
- DefaultEndpointResolverV2 and DefaultHTTPClient integrations give generic clients working defaults (base-endpoint-only resolution, plain http.Client) that were previously never
wired up

there wil be a downstream PR that adds `@Replaces` in SDK codegen where necessary, verified that this doesn't affect the output of the generated SDK at all with those.

### DIFF
--- a/aws-http-auth-schemes/identity/identity.go
+++ b/aws-http-auth-schemes/identity/identity.go
@@ -1,0 +1,81 @@
+// Package identity provides the AWS credential identity types shared by the
+// generic smithy-go AWS auth schemes (SigV4, SigV4A).
+package identity
+
+import (
+	"context"
+	"time"
+
+	smithy "github.com/aws/smithy-go"
+	"github.com/aws/smithy-go/auth"
+	"github.com/aws/smithy-go/aws-http-auth/credentials"
+)
+
+// AWSCredentialIdentity is the [auth.Identity] carrying AWS credentials
+// through the auth pipeline.
+type AWSCredentialIdentity struct {
+	credentials.Credentials
+}
+
+var _ auth.Identity = (*AWSCredentialIdentity)(nil)
+
+// Expiration returns when the underlying credentials expire.
+func (i *AWSCredentialIdentity) Expiration() time.Time {
+	return i.Expires
+}
+
+// AWSCredentialIdentityResolver resolves an [AWSCredentialIdentity].
+//
+// This is the concrete, credential-flavored counterpart to [auth.IdentityResolver]
+// for the SigV4/SigV4A auth schemes. Implementations resolve AWS credentials
+// (static, environment, assumed-role, etc.) and return them boxed as an
+// [AWSCredentialIdentity].
+type AWSCredentialIdentityResolver interface {
+	GetIdentity(context.Context, smithy.Properties) (*AWSCredentialIdentity, error)
+}
+
+// staticAWSCredentialIdentityResolver resolves a fixed, unchanging
+// [AWSCredentialIdentity].
+type staticAWSCredentialIdentityResolver struct {
+	identity AWSCredentialIdentity
+}
+
+var _ AWSCredentialIdentityResolver = (*staticAWSCredentialIdentityResolver)(nil)
+
+// GetIdentity implements [AWSCredentialIdentityResolver].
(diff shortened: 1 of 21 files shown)


aws-http-auth-schemes/identity/identity.go
```diff
@@ -0,0 +1,81 @@
...
+	"time"
+
+	smithy "github.com/aws/smithy-go"
+	"github.com/aws/smithy-go/auth"
+	"github.com/aws/smithy-go/aws-http-auth/credentials"
+)
+
+// AWSCredentialIdentity is the [auth.Identity] carrying AWS credentials
+// through the auth pipeline.
+type AWSCredentialIdentity struct {
```
wty-Bryant, line 16: not sure where to put this nit question: with support of sigv4(a) and non-rule ep resolution for user's own client, should we concern min go version in smithy go like go v2 now?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/GoIntegrationResolver.java
```diff
@@ -0,0 +1,74 @@
...
+        for (GoIntegration integration : integrations) {
+            for (Replaces replaces : integration.getClass().getAnnotationsByType(Replaces.class)) {
+                var target = replaces.value();
+
+                if (target.equals(integration.getClass())) {
+                    throw new CodegenException("GoIntegration " + integration.getClass().getName()
+                            + " declares itself as a replacement target.");
+                }
+
+                var existing = replacedBy.get(target);
```
wty-Bryant, line 45: Is it possible integration A and B replaces each other mutually? The answer might be no if replaces type is only used from sdk side

aws-http-auth-schemes/internal/payloadhash/payloadhash.go
```diff
@@ -0,0 +1,45 @@
...
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+)
+
+// Hash returns the SHA256 hash of the request payload, or the SigV4/SigV4A
+// unsigned-payload sentinel if the payload is not signed.
+//
+// The aws-http-auth signers support implicit payload hashing, but in the
+// client pipeline the body is carried on the separate Stream field, so it has
+// to be hashed here instead.
+func Hash(r *smithyhttp.Request, props *smithy.Properties) ([]byte, error) {
```
wty-Bryant, line 21: So my understanding is this func collapsed/ported all payload hash workflow originated from go v2's sigv4 and runs unsigned payload only when it presents in prop. Can it cover all unsigned cases? For example [dynamicPayloadSigningMiddleware](https://github.com/aws/aws-sdk-go-v2/blob/00daf3640234e0941ba63b2a12cfd21d4e6b2d15/aws/signer/v4/middleware.go#L77) applies unsigned payload only for TLS schemed operations like s3.PutObject

aws-http-auth-schemes/identity/identity.go
lucix-aws: these two things wouldn't matter for that, but also i thought we've been enforcing the same minimum version in smithy-go that we have been in the sdk

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/GoIntegrationResolver.java
lucix-aws: if used in the _intended_ way, no. you could have two integrations in a single codebase replacing each other but that wouldn't make sense

aws-http-auth-schemes/internal/payloadhash/payloadhash.go
lucix-aws: no this is just signing the payload all the time no matter what, which should work for sigv4. the dynamic middleware you're referencing is basically for weird unmodeled AWS behavior

wty-Bryant (approved):

PR #689, opened 2026-07-30
handle null cbor struct val

lucix-aws:
relates https://github.com/aws/aws-sdk-go-v2/issues/3501

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde/cbor/CborDeserializerGenerator.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde/cbor/CborDeserializerGenerator.java
@@ -317,10 +317,9 @@ public final class CborDeserializerGenerator {
             return emptyGoTemplate(); // event stream, not an actual field
         }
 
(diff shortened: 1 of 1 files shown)


kstich (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/serde/cbor/CborDeserializerGenerator.java
```diff
@@ -333,11 +332,11 @@ private Writable deserializeField(MemberShape member) {
                         "fieldName", symbolProvider.toMemberName(member),
                         "deserialize", getDeserializerName(normalize(target)),
                         "deref", generateStructFieldDeref(member, "dv"),
-                        "nilable", isNilable(memberSymbol) ? handleSparseField() : emptyGoTemplate()
+                        "skipNull", handleNullField()
```
Madrigal, line 335: so `smithycbor.Nil` doesn't pass `isNilable(memberSymbol)` because it's an actual value that represents a nil value?

Madrigal (approved):

PR #692, opened 2026-08-06
fixup json snaps and deser

lucix-aws:
fixup all awsjson/restjson response snapshot issues

serde bugs
* dynamically cache collection members instead of trying to precache them on AddMember, this doesn't work when recursive structures call AddMember on some collection before its own members are set
* don't serialize unset json document as `null`

snapshot bugs
* response snapshots have to actually have input because of httpbinding path validation
* actually test `@streaming` blobs in payloads
* compare snapshots by structure rather than bytes, because legacy and schema-serde json order struct members differently
* errors
  * the error discriminator is spliced into the body instead of decode/mutate/re-encode, to prevent more payload shuffling
  * json errors carry `__type` in the body instead of `X-Amzn-ErrorType`. the header is the highest-precedence source, so it was the only one we ever covered, and the body is what services actually send.
  * legacy resolves body `code` before `__type` and schema-serde does the opposite, so a placeholder there resolves to a different type under each, to make them consistent, I just made any code field in an error member also carry the error shape name
* **skip operations that cannot be serialized symmetrically**. see source comment. This only amounts to ~20 operations over the 10k+ we have so I think it's fine
* `CompareValues` had no `reflect.Interface` case, so union and document fields fell through to `DeepEqual` and skipped the reader/document comparators

### DIFF
--- a/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeResponseSnapshotTests.java
+++ b/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeResponseSnapshotTests.java
@@ -15,14 +15,11 @@
 
 package software.amazon.smithy.go.codegen.integration;
 
-import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
-
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import software.amazon.smithy.aws.traits.protocols.AwsJson1_0Trait;
 import software.amazon.smithy.aws.traits.protocols.AwsJson1_1Trait;
 import software.amazon.smithy.aws.traits.protocols.AwsQueryTrait;
@@ -47,15 +44,25 @@ import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.traits.ErrorTrait;
 import software.amazon.smithy.model.traits.HttpErrorTrait;
+import software.amazon.smithy.model.traits.HttpHeaderTrait;
+import software.amazon.smithy.model.traits.HttpLabelTrait;
+import software.amazon.smithy.model.traits.HttpPrefixHeadersTrait;
+import software.amazon.smithy.model.traits.HttpQueryParamsTrait;
+import software.amazon.smithy.model.traits.HttpQueryTrait;
 import software.amazon.smithy.protocol.traits.Rpcv2CborTrait;
-import software.amazon.smithy.utils.MapUtils;
 import software.amazon.smithy.utils.StringUtils;
+import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
 
 public class SerdeResponseSnapshotTests implements GoIntegration {
-    private static final Set<String> SKIP_OPERATIONS = Set.of(
-            // Predict has a customization where an input member goes into the URL. The check path still builds a
-            // request to drive the deserialize, so that customization complicates things here too -- just skip it.
-            "com.amazonaws.machinelearning#Predict"
+    // SDK customizations that aren't visible in the model, so they can't be
+    // detected by inspecting traits
+    //
+    // shape id -> reason
+    private static final Map<String, String> SKIP_OPERATIONS = Map.of(
+            "com.amazonaws.machinelearning#Predict", "endpoint customization",
+            "com.amazonaws.sqs#ReceiveMessage", "md5 customization",
+            "com.amazonaws.sqs#SendMessage", "md5 customization",
+            "com.amazonaws.sqs#SendMessageBatch", "md5 customization"
     );
 
     @Override
@@ -73,7 +80,7 @@ public class SerdeResponseSnapshotTests implements GoIntegration {
         }
 
         var serviceSchemaRef = "schemas." + StringUtils.capitalize(service.getId().getName(service));
-        var generator = new SnapshotOutputGenerator(model, symbolProvider);
+        var generator = new SnapshotOutputGenerator(model, symbolProvider, settings);
 
         // The Check tests are ALWAYS generated: they only read a fixture, inject it, run the real deserialize path,
         // and compare the output. They have no dependency on output Serialize/schemas, so they compile and run whether
@@ -82,7 +89,7 @@ public class SerdeResponseSnapshotTests implements GoIntegration {
         goDelegator.useFileWriter("response_snapshot_test.go", settings.getModuleName(), writer -> {
             writer.addBuildTag("response_snapshot");
             writer.write(checkCommonSource());
-            writer.write(checks(model, service, symbolProvider, generator));
+            writer.write(checks(settings, model, service, symbolProvider, generator));
         });
 
         // The Update tests (fixture generation) serialize the output value via a throwaway protocol, which requires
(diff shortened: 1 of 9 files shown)


codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeResponseSnapshotTests.java
```diff
@@ -339,15 +378,23 @@ func serdeRespXMLErrorEnvelope(body []byte, code string) []byte {
...
+            SnapshotOutputGenerator generator
     ) {
         var eventStreamIndex = EventStreamIndex.of(model);
+        var inputs = new SnapshotInputGenerator(model, symbolProvider, settings, false);
         var writables = new ArrayList<Writable>();
         var operations = sortedOperations(model, service, eventStreamIndex);
-        for (var operation : operations) {
-            var inputSymbol = symbolProvider.toSymbol(model.expectShape(operation.getInputShape()));
+        for (var operation : allOperations(model, service)) {
+            var reason = skipReason(model, operation, eventStreamIndex);
```
Madrigal, line 389: praise: I like this reason mechanism

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
```diff
@@ -32,32 +35,37 @@ public void writeAdditionalFiles(
...
+            var generator = new SnapshotInputGenerator(model, symbolProvider, settings, false);
             writer.write(snapshotTests(model, service, symbolProvider, generator));
             writer.write(snapshotUpdaters(model, service, symbolProvider, generator));
         });
     }
 
-    private static boolean isCbor(GoSettings settings) {
-        return Rpcv2CborTrait.ID.equals(settings.getProtocol());
+    // Emits serdeBodyEqual, the request-body comparator, per protocol. Every variant falls back to a byte compare
+    // when either side won't decode, so a malformed golden fails loudly rather than silently passing.
```
Madrigal, line 48: what's a malformed golden?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
```diff
@@ -32,32 +35,37 @@ public void writeAdditionalFiles(
...
-    private Writable bodyEqual(boolean isCbor) {
-        if (!isCbor) {
-            return goTemplate("""
-                    func serdeBodyEqual(got, expected []byte) bool {
-                        return bytes.Equal(got, expected)
-                    }
-                    """);
-        }
+    // rpcv2Cbor encodes struct fields as a CBOR map and the encoder emits map entries in Go map iteration order, so
+    // the same input produces different byte orderings across runs. Comparing decoded values is order-independent.
```
Madrigal, line 67: we don't have to resolve it here, but this has also been a point of friction on JSON encoders. Since we are now taking more control of the encoders, would it be worth it to sort map entries?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SnapshotInputGenerator.java
```diff
@@ -86,6 +108,67 @@ public List<TestCase> generateCasesForShape(StructureShape shape) {
 
     private static final int MAX_DEPTH = 10;
 
+    private static final String ERROR_TYPE_HEADER = "x-amzn-errortype";
+    private static final String ERROR_CODE_MEMBER = "code";
+
+    // legacy json resolves errortype header -> code -> __type
+    // schema-serde resolves errortype header -> __type -> code
+    //
+    // the latter is objectively more correct because some services e.g. chime actually have a separate modeled code
```
Madrigal, line 117: RIP Chime, but more than that, is this an ambiguous spec?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SnapshotInputGenerator.java
```diff
@@ -86,6 +108,67 @@ public List<TestCase> generateCasesForShape(StructureShape shape) {
...
+    }
+
+    private boolean isJsonProtocol() {
+        var protocol = settings.getProtocol();
+        return AwsJson1_0Trait.ID.equals(protocol)
+                || AwsJson1_1Trait.ID.equals(protocol)
+                || RestJson1Trait.ID.equals(protocol);
+    }
+
+    // Success fixtures are written with HTTP 200, so an @httpResponseCode member deserializes to exactly that.
```
Madrigal, line 154: Do we have any fixtures that have some success non-200 like 201 created?

schema.go
```diff
@@ -176,17 +166,31 @@ func cloneTraits(src map[ShapeID]Trait) map[ShapeID]Trait {
...
+	return s.lookup(&s.mapKey, "key")
 }
 
 // MapValue returns the "value" schema for map types.
 func (s *Schema) MapValue() *Schema {
-	return s.mapValue
+	return s.lookup(&s.mapValue, "value")
+}
+
+func (s *Schema) lookup(cached *atomic.Pointer[Schema], name string) *Schema {
```
Madrigal, line 182: if they are `atomic.Pointer` does this need to be synchronized?

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
lucix-aws: golden = snapshot file

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
lucix-aws: by default, a million percent no, it's bad for performance,
as an option, such that we could turn it on for snapshots - 100% and i intend to do so. we have an open issue for this over in the sdk.

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SnapshotInputGenerator.java
lucix-aws: the spec doesn't specify an order, yeah. but i would say schema-serde is more correct because code is basically a nonstandard legacy thing

schema.go
lucix-aws: it shouldn't? what would the issue be? the concurrent map read should be fine if that does happen at the start

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SnapshotInputGenerator.java
lucix-aws: i don't think so

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeSnapshotTests.java
Madrigal: ok, as an option it sounds good. 

> by default, a million percent no, it's bad for performance

Can't argue with this, but we've had multiple issues with serialization not being deterministic being an issue even if it's not a guarantee for JSON, so it's definitely a tredoff.

Interestingly, JSON v2 switched this behavior and now map keys are not deterministic by default and you have to pass a deterministic flag https://pkg.go.dev/encoding/json/v2#Deterministic

schema.go
Madrigal: yes basically concurrent map read, but if it's fine then it's fine

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SnapshotInputGenerator.java
Madrigal: not a blocker, but is that more correct because we saw it on a newer service? maybe worth seeing if we can spec-ify this

Madrigal (approved):

codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/SerdeResponseSnapshotTests.java
```diff
@@ -358,21 +405,47 @@ private Writable checks(
...
+    private List<OperationShape> allOperations(Model model, ServiceShape service) {
+        var operations = new ArrayList<>(TopDownIndex.of(model).getContainedOperations(service));
+        operations.sort(Comparator.comparing(o -> o.getId().toString()));
+        return operations;
+    }
+
+    // A test that exists only to record, in the generated SDK, that an operation or error has no snapshot coverage.
+    private Writable writeSkip(String testName, String reason) {
+        return goTemplate("""
+                func $name:L(t *$testingT:T) {
```
wty-Bryant, line 432: same, like this placeholder test with operation in signature and reason in log

wty-Bryant (approved):

PR #693, opened 2026-08-06
pool byte buffers on deserialize + string arena

lucix-aws:
* io.ReadAll is inefficient for response buffering because it starts at like 512 bytes and repeatedly doubles until it consumes the reader, which means we waste a bunch of intermediate allocations. pool byte buffers on deserialization for all protocols, up to a limit
* add a string arena, the goal here is to reduce allocation pressure for repeated small strings. It sizes to 1/4th of the response payload but I've also capped it at 4k for now, so big responses etc. won't cause a lot of memory to be pinned
* attempt to presize buffers for `@httpPayload + blob` rather than io.ReadAll

### DIFF
--- a/internal/serde/read_payload_blob.go
+++ b/internal/serde/read_payload_blob.go
@@ -1,0 +1,34 @@
+package serde
+
+import (
+	"bytes"
+	"io"
+)
+
+// maxPresize bounds how large a buffer we will allocate up front from a
+// Content-Length header. Past this point we allocate maxPresize and let the read
+// grow the buffer to fit whatever actually arrives.
+//
+// We arbitrarily choose 512K.
+const maxPresize = 512 * 1024
+
+// ReadPayloadBlob consumes the given reader into a buffer that does not come
+// from (and will never be returned to) a pool, sizing it from contentLength
+// when that is known and within bounds.
+func ReadPayloadBlob(r io.Reader, contentLength int64) (*bytes.Buffer, error) {
+	buf := &bytes.Buffer{}
+	if contentLength > 0 {
+		presize := contentLength
+		if presize > maxPresize {
+			presize = maxPresize
+		}
+
+		// + bytes.MinRead prevents pointless doubling on EOF
+		buf.Grow(int(presize) + bytes.MinRead)
+	}
+
+	if _, err := buf.ReadFrom(r); err != nil {
+		return nil, err
+	}
+	return buf, nil
+}
--- a/transport/http/protocol/awsjson/awsjson.go
+++ b/transport/http/protocol/awsjson/awsjson.go
@@ -7,15 +7,16 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/aws/smithy-go"
-	internaljson "github.com/aws/smithy-go/transport/http/protocol/internal/json"
 	internalerrors "github.com/aws/smithy-go/internal/errors"
 	internales "github.com/aws/smithy-go/internal/eventstream"
+	internalsync "github.com/aws/smithy-go/internal/sync"
 	smithyio "github.com/aws/smithy-go/io"
-	"github.com/aws/smithy-go/middleware"
 	"github.com/aws/smithy-go/traits"
 	smithyhttp "github.com/aws/smithy-go/transport/http"
+	internaljson "github.com/aws/smithy-go/transport/http/protocol/internal/json"
 )
 
 // ProtocolOptions configures aws.protocols#awsJson1_0.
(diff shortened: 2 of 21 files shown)


lucix-aws:
performance numbers when running this patch against our service/internal/serdebenchmark in the SDK

image

transport/http/protocol/awsjson/awsjson.go
```diff
@@ -89,7 +83,7 @@ func (p *Protocol) SerializeRequest(
 	if len(req.URL.Path) == 0 {
 		req.URL.Path = "/"
 	}
-	req.Header.Set("X-Amz-Target", fmt.Sprintf("%s.%s", p.serviceName, middleware.GetOperationName(ctx)))
+	req.Header.Set("X-Amz-Target", fmt.Sprintf("%s.%s", p.serviceName, schema.ID().Name))
```
Madrigal, line 86: is this available even on non-migrated services?

transport/http/protocol/internal/cbor/shape_deserializer.go
```diff
@@ -50,7 +52,10 @@ func NewShapeDeserializer(p []byte, opts ...func(*ShapeDeserializerOptions)) *Sh
 	for _, fn := range opts {
 		fn(&o)
 	}
-	return &ShapeDeserializer{p: p, head: serde.NewStack[deserCtx](), opts: o}
+
+	d := &ShapeDeserializer{p: p, head: serde.NewStack[deserCtx](), opts: o}
+	d.arena.Reset(len(p) / 4)
```
Madrigal: why `/4`?

transport/http/protocol/internal/cbor/zz_alias_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if key != "key" {
+		t.Errorf("map key ALIASES the payload: became %q", key)
+	}
+	if val != "value" {
+		t.Errorf("string value ALIASES the payload: became %q", val)
+	}
+}
+
+// An indefinite-length byte string is legal CBOR that ReadBlob does not
```
Madrigal, line 98: excellent corner case

transport/http/protocol/internal/httpbinding/ext_test.go
```diff
@@ -0,0 +1,101 @@
...
+// The walk this caches does not allocate, so allocation cannot tell you whether
+// the cache is live -- what it costs is time. Assert the mechanism instead: the
+// extension is built once and returned by pointer thereafter, and HasBlobPayload
+// reads it. BenchmarkHasBlobPayload measures what that saves.
+func TestBindingExtIsCached(t *testing.T) {
+	blob := smithy.NewSchema(smithy.ShapeID{Namespace: "test", Name: "Blob"}, smithy.ShapeTypeBlob, 0)
+	out := outputWith("body", blob, &traits.HTTPPayload{})
+
+	if !HasBlobPayload(out) {
+		t.Fatal("expected true")
```
Madrigal:
```suggestion
		t.Fatal("expected schema to have blob payload")
```

transport/http/protocol/internal/httpbinding/ext_test.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	if second := getExt(out); second != first {
+		t.Error("extension rebuilt on second access; the slot is not caching")
+	}
+	if !first.hasBlobPayload {
+		t.Error("cached extension holds the wrong answer")
+	}
+}
+
+func BenchmarkHasBlobPayload(b *testing.B) {
```
Madrigal: do we run `testing.B` tests?

internal/serde/read_payload_blob.go
```diff
@@ -0,0 +1,34 @@
...
+// when that is known and within bounds.
+func ReadPayloadBlob(r io.Reader, contentLength int64) (*bytes.Buffer, error) {
+	buf := &bytes.Buffer{}
+	if contentLength > 0 {
+		presize := contentLength
+		if presize > maxPresize {
+			presize = maxPresize
+		}
+
+		// + bytes.MinRead prevents pointless doubling on EOF
```
Madrigal, line 26: nice

Madrigal (approved):
Minor comments, super happy to see this!

transport/http/protocol/awsjson/awsjson.go
lucix-aws: as in non-schema serde services? they're not calling this

transport/http/protocol/internal/cbor/shape_deserializer.go
lucix-aws: a fourth of the payload up to 4k, i'll const this

transport/http/protocol/internal/httpbinding/ext_test.go
lucix-aws: this was mostly intermediate to check the schema extension perf improved it, i'm probably jst gonna delete this

PR #694, opened 2026-08-11
Set Content-Length when the request body is set instead of via middleware

Tongs2000:
*Issue #, if available:*

*Description of changes:*

  `Request.SetStream` now sets `ContentLength` when the stream length can be determined, so the SDK no longer needs the standalone `ComputeContentLength` middleware. That middleware is kept but deprecated.

  The length is set when the body is a known-length reader such as a bytes or strings reader or a seekable stream. When the length cannot be determined, for example a non-seekable stream, `ContentLength` is left unchanged so a user-provided value is preserved. Request compression updates the length after rewriting the body.

  This is paired with the aws-sdk-go-v2 change (aws/aws-sdk-go-v2#3513) that stops registering `ComputeContentLength` and regenerates clients.

### DIFF
--- a/transport/http/request.go
+++ b/transport/http/request.go
@@ -154,6 +154,25 @@ func (r *Request) SetStream(reader io.Reader) (rc *Request, err error) {
 	return rc, err
 }
 
+// SetStreamWithLength sets the request stream like SetStream, and additionally
+// sets ContentLength when the stream's length can be determined. If it cannot
+// (e.g. a non-seekable stream), ContentLength is left unchanged so an
+// explicitly provided value is preserved.
+func (r *Request) SetStreamWithLength(reader io.Reader) (rc *Request, err error) {
+	rc, err = r.SetStream(reader)
+	if err != nil {
+		return rc, err
+	}
+
+	if n, ok, err := rc.StreamLength(); err != nil {
+		return rc, err
+	} else if ok {
+		rc.ContentLength = n
+	}
+
+	return rc, nil
+}
+
 // Build returns a build standard HTTP request value from the Smithy request.
 // The request's stream is wrapped in a safe container that allows it to be
 // reused for subsequent attempts.
--- a/transport/http/request.go
+++ b/transport/http/request.go
@@ -118,6 +118,9 @@ func (r *Request) IsStreamSeekable() bool {
 // SetStream returns a clone of the request with the stream set to the provided
 // reader. May return an error if the provided reader is seekable but returns
 // an error.
+//
+// ContentLength is set to the stream's length when it can be determined, and
+// left unchanged otherwise.
 func (r *Request) SetStream(reader io.Reader) (rc *Request, err error) {
 	rc = r.Clone()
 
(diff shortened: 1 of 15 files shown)


lucix-aws:
Why the new API, why not just augment SetStream to do this? There already seems to be some logic that checks the stream length for something else there, even.

Tongs2000:
> Why the new API, why not just augment SetStream to do this? There already seems to be some logic that checks the stream length for something else there, even.

Some middlewares in the `Deserialize` step also uses `SetStream` to restore the body after reading it. They don't compute/set ContentLength.

lucix-aws:
Do you have an example?

Tongs2000:
> There already seems to be some logic that checks the stream length for something else there, even.

Yes. Like the request/response logging middleware, in its `HandleDeserialize`: https://github.com/aws/smithy-go/blob/c589db5abf678c5ea02821245e7c0887a5a34af6/transport/http/middleware_http_logging.go#L50

lucix-aws:
I'm not _entirely_ sure why that extra SetStream call is there, but regardless, why would it be a problem to re-set content length there?

Tongs2000:
> I'm not _entirely_ sure why that extra SetStream call is there, but regardless, why would it be a problem to re-set content length there?

Double checked, you are right. 

Besides `RequestCompression`, only two middlewares call `SetStream` — `RequestResponseLogger` in the `Deserialize` step and `InitializeStreamWriter` (eventstream) in the `Finalize` step.

For `RequestResponseLogger` it's restoring the body after `DumpRequestOut` reads it. For `InitializeStreamWriter` (eventstream), it swaps the body to an in-memory `io.Pipe()` so event stream messages can be written after the request is sent. 

Both pass a body whose length can't be determined. So, moving the length logic into `SetStream` won't affect them as the length can't be determined we leave `ContentLength` unchanged.

transport/http/request.go
```diff
@@ -151,6 +154,12 @@ func (r *Request) SetStream(reader io.Reader) (rc *Request, err error) {
 	rc.isStreamSeekable = isStreamSeekable
 	rc.streamStartPos = streamStartPos
 
+	if n, ok, err := rc.StreamLength(); err != nil {
```
lucix-aws, line 157: just double-checking but there is no practical place that this returns an error, right? because basically all of our payloads are in-memory so seeking won't fail

lucix-aws:
please add a changelog

transport/http/request.go
Tongs2000: Small correction to what I said earlier: I was wrong that only those three middlewares call `SetStream`. The middleware `OperationSerializer` call it too which is actually the main caller, including for user-supplied bodies. So the error branch isn't dead. A user can pass a seekable body without `Len()`. e.g. `PutObject` with `Body` set to an `*os.File`. If that seek fails `StreamLength` returns the error. So I'd keep the error handling.

lucix-aws (approved):

PR #698, opened 2026-08-19
Keep trailing slash when JoinPath right operand is a slash

### LINKED ISSUE #631: http: JoinPath drops the trailing / if RHS is /
See https://github.com/aws/smithy-go/pull/630#discussion_r2872771218,

e.g.

```go
smithyhttp.JoinPath("/foo", "/") => /foo (should be /foo/)
```

This _might_ break a custom implementation of an awsJson service. No AWS services today have path signatures like this though.

Hashim1999164:
## Summary
JoinPath("/foo", "/") returned "/foo" because the right operand was stripped to empty and the trailing slash was lost.

The result now keeps that trailing slash. Existing path joins are unchanged.

Fixes #631

## Test plan
Run go test ./transport/http/


Tongs2000 (approved):
Thanks for the fix! Approved.

PR #699, opened 2026-08-20
fixup some regressions for schema-serde found in json wave 2

lucix-aws:
* explicit empty lists were deserialized as nothing rather than an empty slice
* the event stream internals were missing some interface impls that the closing mechanisms depend on
* the new protocol-agnostic event stream stuff wasn't wired up to forward generic API error details

### DIFF
--- a/internal/eventstream/codec.go
+++ b/internal/eventstream/codec.go
@@ -16,6 +16,9 @@ type Codec struct {
 	Deserializer func([]byte) smithy.ShapeDeserializer
 	ContentType  string
 
+	// Protocol-specific hook to retrieve error information from some payload.
+	ErrorInfo func(payload []byte) (code, message string, err error)
+
 	encoder    *eventstream.Encoder
 	decoder    *eventstream.Decoder
 	payloadBuf []byte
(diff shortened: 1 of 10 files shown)


internal/eventstream/codec.go
```diff
@@ -44,8 +47,10 @@ func (c *Codec) SerializeEventMessage(schema, variant *smithy.Schema, v smithy.S
 
 	v.Serialize(ss)
 
-	msg.Headers.Set(eventstream.MessageTypeHeader, eventstream.StringValue(eventstream.EventMessageType))
+	// header order matches the legacy per-operation serializers: the event type
```
Madrigal, line 50: non-blocker: is this actually required by the spec?

internal/eventstream/codec.go
```diff
@@ -146,9 +151,26 @@ func (c *Codec) deserializeException(schema *smithy.Schema, types *smithy.TypeRe
 
 	perr, ok := types.DeserializableError(id)
 	if !ok {
+		code := exType.String()
+		message := "UnknownError"
+		if c.ErrorInfo != nil {
+			bodyCode, bodyMessage, err := c.ErrorInfo(msg.Payload)
+			if err != nil {
+				return &smithy.DeserializationError{
+					Err:      fmt.Errorf("get error info: %w", err),
```
Madrigal, line 160: nit: should it be `got error info:`?

Madrigal (approved):

internal/eventstream/codec.go
lucix-aws: not at all, i just chose to make the behavior match the old rather than the test in the other thing

internal/eventstream/codec.go
lucix-aws: no since it's an error when _trying_ to deserialize, right?

lucix-aws: error wraps typically go like fmt.Errorf("the thing i just tried to do: %w", err)

wty-Bryant (approved):

PR #702, opened 2026-08-26
Add regression test for waiter.Logger.AddLogger

lucix-aws:


wty-Bryant (approved):
not sure if we need a changelog for this test if we already merge the change earlier

PR #706, opened 2026-09-07
encoding: json and xml Value write non-finite floats as strings instead of panicking

lenamonj:
*Issue #, if available:*

*Description of changes:*

`encoding/json` and `encoding/xml` `Value.Float` and `Value.Double` hand every value to `encoding.EncodeFloat`, which panics on NaN and the infinities, so serializing one of those float64 values kills the process. The protocol serializers under `transport/http/protocol/internal` already write them as the strings `NaN`, `Infinity` and `-Infinity`; this change makes the two `Value` encoders do the same, quoted in JSON and as element text in XML.

Three cases each in the existing `TestValue` tables panic on main and pass with the change. `make unit-race` passes across all modules, and gorelease reports the change as compatible.

### DIFF
--- a/.changelog/b47e0e8f-e94e-4339-8398-f20a1af31ff2.json
+++ b/.changelog/b47e0e8f-e94e-4339-8398-f20a1af31ff2.json
@@ -1,0 +1,9 @@
+{
+    "id": "b47e0e8f-e94e-4339-8398-f20a1af31ff2",
+    "type": "bugfix",
+    "description": "encoding/json and encoding/xml Value encode NaN and the infinities as the strings NaN, Infinity and -Infinity instead of panicking.",
+    "collapse": false,
+    "modules": [
+        "."
+    ]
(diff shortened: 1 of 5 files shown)


lucix-aws:
Did you encounter this in a real client call? e.g. w/ an AWS service that is still using the old legacy serde? If you generate a smithy client today it will use schema-based (de)serialization by default and won't use these APIs.

With the recent release of schema-serde, the short-term intent is that these old encoding/json and encoding/xml "Value" APIs become deprecated, and the new codecs that are currently hidden in internal/ also become the new public API

lenamonj:
No, from auditing the package's public API, not from a service call. The schema-serde JSON serializer already writes Infinity, -Infinity and NaN as those strings, which is what this change makes the legacy Value do, so the two agree. If the Value APIs are on their way out, fine to close this with them.

PR #713, opened 2026-09-16
fix decode panic on event stream header type and bad cbor string/blob len

lucix-aws:
* header fix is self-explanatory and was previously fixed in the sdk, same as https://github.com/aws/aws-sdk-go-v2/pull/3355
* the cbor decode for string/blob just blindly casted a 64-bit value to an int to get an offset, which could underflow and then panic on slice bounds, so check the remaining buffer size against the uncasted number first


transport/http/protocol/internal/cbor/shape_deserializer_test.go
```diff
@@ -0,0 +1,113 @@
...
+	blobSchema := smithy.NewSchema(smithy.ShapeID{Namespace: "test", Name: "B"}, smithy.ShapeTypeBlob, 0)
+
+	tests := []struct {
+		name    string
+		payload []byte
+		read    func(*ShapeDeserializer) error
+	}{
+		{
+			name:    "string",
+			payload: oversizedLengthToken(0x60),
```
Madrigal: do these hex values mean something special or are they just "any value"?

eventstream/debug.go
```diff
@@ -139,6 +139,6 @@ func valueFromType(typ valueType, val any) (Value, error) {
 		copy(tv[:], v)
 		return tv, err
 	default:
-		panic(fmt.Sprintf("unknown type, %s, %T", typ.String(), val))
+		return nil, fmt.Errorf("unable to decode unknown type, %s, %T", typ.String(), val)
```
Madrigal, line 142: TIL we have a `debug` version, a bit worried that we have to fix it in two places but if you think that's useful go for it

Madrigal (approved):
Couple of non-blocking comments

eventstream/debug.go
lucix-aws: it's not actually debug code, the file is just called debug.go for whatever reason, the code in there is basically json marshaler stuff used when you turn on event stream logging

transport/http/protocol/internal/cbor/shape_deserializer_test.go
lucix-aws: it was cbor major typed but packed and shifted into an 8bit, i made it way clearer

